PR #2, opened 2015-04-29
Review for DM-2191 (pipe_tasks): Tasks for measurement transforms

TallJimbo:


python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -119,32 +119,18 @@ def run(self, inputCat, wcs, calib):
...
-        default="SingleFrame",
-        allowed={
-            "SingleFrame": "Single frame measurement",
-            "Forced": "Forced measurement"
-        }
-    )
 
 
-class RunTransformTask(pipeBase.CmdLineTask):
+class RunTransformTaskBase(object):
```
TallJimbo: Why do you need to inherit from `object` here and multiply-inherit later in `makeTransformCmdLineTask`?  It seems like you could just inherit from `CmdLineTask`, even if this intermediate base class remains abstract.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +138,85 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
+    forcedField = pexConfig.Field(dtype=bool, default=wasForced, doc="Measurement operation was forced")
+
+    configClass = type("%sTransformConfig" % sourceType.capitalize(),
+                       (RunTransformConfigBase, pexConfig.Config),
+                       {"sourceType": sourceTypeField, "calexpType": calexpTypeField,
+                        "outputType": outputTypeField, "measConfig": measConfigField,
+                        "wasForced": forcedField})
+    taskClass = type("%sTransformTask" % (sourceType.capitalize(),),
+                     (RunTransformTaskBase, pipeBase.CmdLineTask),
+                     {"wasForced": wasForced, "ConfigClass": configClass,
```
jdswinbank: That ``wasForced`` is not needed; sorry.

python/lsst/pipe/tasks/transformMeasurement.py
jdswinbank: My worry was that it's abstract-ness is not immediately clear: since we're providing a ``run()`` method, it looks like something you could pick up and use straight away as a command line task, but it wouldn't work.

I'm not particularly attached to this way of doing things though; I'm happy to just use single inheritance if you prefer.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +138,85 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
+    """
+    sourceTypeField = pexConfig.Field(dtype=str, default=sourceType,
+                                      doc="Dataset type of source to transform")
+    calexpTypeField = pexConfig.Field(dtype=str, default=calexpType,
+                                      doc="Dataset type of calibration exposure")
+    measConfigField = pexConfig.Field(dtype=str, default="processCcd_config",
+                                      doc="Dataset type of measurement operation configuration")
+    outputTypeField = pexConfig.Field(dtype=str, default=outputType,
+                                      doc="Dataset type of output sources")
+    forcedField = pexConfig.Field(dtype=bool, default=wasForced, doc="Measurement operation was forced")
```
TallJimbo: I don't think most of these should be config fields; I think class attributes would be sufficient for most cases.

Unfortunately, that's not sufficient for all cases, because the coadd outputs are indeed dependent on their configuration - but they're dependent on configuration options that are different from the ones here (i.e. one configures only the prefix, not the whole dataset name).  And for those, I think it'd be nice to make it possible to read the config values from the original processing task, the way the measurement config is.  I'll discuss this more elsewhere, but I think this is pushing me towards preferring having a base class for transformations that is manually subclassed for each specialization, rather than having a generating function for the new tasks.

TallJimbo:
I'm afraid I think I am going to recommend a rather major change in approach here (and right before you go on vacation, no less).  The task-generating function is neat, but I think we'd get more flexibility (allowing for a better task-user interface) and code readability by going with an approach in which the base class task simply requires subclasses to define a class attribute or callback method to define the information it needs.

Since code will explain this better than text, here's a bit of what I have in mind, limited to just how to set the input dataset:
```
class RunTransformTaskBase(CmdLineTask):
    inputCatalogDataSet = None
    inputConfigDataSet = None

    def getInputCatalogDataSet(self):
        """Return the butler dataset name for the catalog to be transformed.

        Default implementation returns self.inputCatalogDataSet; subclasses whose
        datasets are fixed can simply set that attribute instead of overriding.
        """
        return self.inputCatalogDataSet

    def getInputCatalogDataSet(self):
        """Return the butler dataset name for the configuration of the original
        task that produced the measurement.

        Default implementation returns self.inputConfigDataSet; subclasses whose
        datasets are fixed can simply set that attribute instead of overriding.
        """
        return self.inputConfigDataSet

class SrcTransformTask(RunTransformTaskBase):
    inputCatalogDataSet = "src"
    inputConfigDataSet = "processCcd_config"

class CoaddSrcTransformTask(RunTransformTaskBase):
    inputConfigDataSet = "processCoadd_config"
    def getInputCatalogDataSet(self):
        config = self.getInputTaskConfig()
        return "%s_src" % config.coaddName
```

That's a little more verbose for the definers of new tasks, but I think that's a price worth paying in this case.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +138,85 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
-            pluginRegistry = measBase.sfm.SingleFramePlugin.registry
-        elif self.config.measType == "Forced":
+        if self.config.wasForced:
             pluginRegistry = measBase.forcedMeasurement.ForcedPlugin.registry
+        else:
+            pluginRegistry = measBase.sfm.SingleFramePlugin.registry
         self.makeSubtask('transform', pluginRegistry=pluginRegistry,
                          measConfig=kwargs['butler'].get(self.config.measConfig).measurement.value,
-                         inputSchema=kwargs['butler'].get("src_schema").schema)
+                         inputSchema=kwargs['butler'].get(self.config.sourceType + "_schema").schema)
```
TallJimbo: We can't assume that the measurement configuration will always be in `.measurement`; we'll also need to handle e.g. the two measurements run in the `.calibrate` subtask of `ProcessCcdTask` (i.e. `root.calibrate.initialMeasurement` and `root.calibrate.measurement`).  The `initialMeasurement` run is a little weird, though, as it's added into the calibrate catalog dataset ("icSrc") via a SchemaMapper with a prefix, so it may not work with your lower-level code at all.  I'd be content with ignoring it, but we do need to be able to handle `root.calibrate.measurement`.

Given that `ProcessCcdTask` generates two different catalogs, I'm torn as to whether we should have a single transform CmdLineTask to handle both of them or a separate one for each catalog.  I suspect the latter is probably easier.

By the way, `ProcessCoaddTask` may well have a `CalibrateTask` stage as well, but if so, that's a historical artifact that will go away with the HSC merges, and it's not worth worrying about here.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +138,85 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
+# transformation tasks can be defined following this pattern in the appropriate obs_ package.
+SrcTransformConfig, SrcTransformTask = makeTransformCmdLineTask(
+                                                "src", "calexp", "processCcd_config", "transformedSrc", False)
+ForcedSrcTransformConfig, ForcedSrcTransformTask = makeTransformCmdLineTask(
+                                       "forced_src", "calexp", "forced_config", "transformedForced_src", True)
+DeepCoaddSrcTransformConfig, DeepCoaddSrcTransformTask = makeTransformCmdLineTask(
+           "deepCoadd_src", "deepCoadd_calexp", "deep_processCoadd_config", "transformedDeepCoadd_src", False)
+DeepCoaddForcedSrcTransformConfig, DeepCoaddForcedSrcTransformTask = makeTransformCmdLineTask(
+                                        "deepCoadd_forced_src", "deepCoadd_calexp", "deepCoadd_forced_config",
+                                                                      "transformedDeepCoadd_Forced_src", True)
```
TallJimbo: I don't think we want separate tasks for different kinds of coadds (e.g. "deepCoadd" vs "goodSeeingCoadd").  That needs to be something set via config (see my proposal on the main conversation section of the PR).

python/lsst/pipe/tasks/transformMeasurement.py
TallJimbo: I do think single inheritance is cleaner here, and it's a pattern we already use regularly (see `ProcessImageTask`).  Declaring that the class is abstract and should not be used directly in the documentation (along with having "Base" in the name somewhere) should be sufficient to avoid confusion.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +138,85 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
+                     (RunTransformTaskBase, pipeBase.CmdLineTask),
+                     {"wasForced": wasForced, "ConfigClass": configClass,
+                      "_DefaultName": "transform%sMeasurement" % (sourceType.capitalize(),)})
+    return configClass, taskClass
+
+
+# Explicitly instantiate command line transformation types for common source types. Camera-specific
+# transformation tasks can be defined following this pattern in the appropriate obs_ package.
+SrcTransformConfig, SrcTransformTask = makeTransformCmdLineTask(
+                                                "src", "calexp", "processCcd_config", "transformedSrc", False)
```
TallJimbo: Instead of "transformedSrc", let's go with "transformed_src", so we're just adding a prefix and not changing any capitalization (we don't at present have any naming conventions for butler datasets, so either is legal).  @RobertLuptonTheGood has analysis code that can operate generically on different types of data by just using different prefixes, and while we may not use that in the long run I think not requiring extra camelcase smarts to machine-generate related names from each other is good practice in general.

In fact, I think it'd be entirely appropriate for all of these tasks to essentially hard code that their output dataset is "transformed_" prepended to the input name; allowing flexibility doesn't seem to have any advantages.

RobertLuptonTheGood: I'm hoping that my prefix-hack will be replaced by some sort of scoping in @ktlim's New Butler

PR #4, opened 2015-05-26
Move PhotoCalTask and Colorterm class here from meas_astrom

SimonKrughoff:
In addition to those classes I moved:
- The one photoCal example
- The unit tests

I also copied some test data. It's not a lot of data and I feel much more comfortable
with a local copy that we can rely on, instead of using data from a different package
that might change without warning.


tests/astrometry_net_data/photocal/andConfig.py
```diff
@@ -0,0 +1,7 @@
+root.defaultMagColumn = 'r'
+root.starGalaxyColumn = 'star'
+#root.variableColumn = 'variable'
```
SimonKrughoff: I'm not exactly sure why we need all these configs.  If we don't need them all, I'd like to remove the unused ones.  I would also like to clean up the commented out code.

PaulPrice:
Do you want to pull the changes to Colorterm from HSC before you move this?

SimonKrughoff:
@PaulPrice I wasn't aware of changes from HSC what are they?

PaulPrice:
Changed Colorterms so it's not a global, and it can now be configured using `Config`.  See https://github.com/HyperSuprime-Cam/meas_astrom/blob/master/python/lsst/meas/photocal/colorterms.py

r-owen:
I would rather make updating Colorterm a separate ticket. It does sound like a useful change.

tests/astrometry_net_data/photocal/andConfig.py
r-owen: I am happy to remove uncommented code, but I don't think it is safe to remove the configs without extensive testing, and the example you give above looks useful.

r-owen: Never mind. You were right about the andconfig files. Only one of them (the one with no digit in the name) was wanted. I'll remove the others and remove the one commented-out line from the remaining file.

PaulPrice:
Having it another ticket is fine, but you may want to do it before moving packages.

jdswinbank:

PR #5, opened 2015-05-27
pipe_tasks tickets/DM-2533: Remove temporary schema version workarounds.

TallJimbo:


laurenam:
There are references to old algorithm names in the setDefaults function in calibrate.py:
https://github.com/lsst/pipe_tasks/blob/master/python/lsst/pipe/tasks/calibrate.py#L118

Should these be updated in this ticket?

TallJimbo:
Yup, that line in {{setDefaults}} should be updated.  Apparently it's not an error to try to remove algorithm names that don't exist (at least if it's done this way), which explains why it wasn't causing problems before.

TallJimbo:
Merged to master on tickets/DM-1766.

PR #8, opened 2015-06-01
pipe_tasks tickets/DM-2191, review 2

TallJimbo:
A fresh PR for a new review after John's refactoring.


TallJimbo:
Looks like you need a bin/ script for coadd measurement as well.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -62,22 +73,44 @@ class TransformTask(pipeBase.Task):
...
+
+    \section pipe_tasks_transform_invoke Task invocation
+
+    \copydoc run
     """
     ConfigClass = TransformConfig
     _DefaultName = "transform"
 
-    def __init__(self, measConfig, pluginRegistry, inputSchema, *args, **kwargs):
+    def __init__(self, measConfig, pluginRegistry, inputSchema, outputType, *args, **kwargs):
```
TallJimbo: You can actually get the `PluginRegistry` from `measConfig`: `measConfig.plugins.registry`, so it doesn't need to be a separate argument.

Also, can't comment on a few lines below, since it's not a change, but "measurments" is a typo.

tests/testTransform.py
```diff
@@ -53,11 +53,13 @@
 import lsst.afw.coord as afwCoord
 import lsst.afw.table as afwTable
 import lsst.meas.base as measBase
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
```
jdswinbank: (with apologies for "self-reviewing"), both of those new imports are unnecessary.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -95,7 +127,7 @@ def __init__(self, measConfig, pluginRegistry, inputSchema, *args, **kwargs):
     def getSchemaCatalogs(self):
         """!Return a dict containing an empty catalog representative of this task's output."""
         transformedSrc = afwTable.BaseCatalog(self.mapper.getOutputSchema())
-        return {'transformedSrc': transformedSrc}
+        return {self.outputType: transformedSrc}
```
TallJimbo: The fact that you have to know `self.outputType` here bugs me, but there's nothing to be done about it now.  I have just created https://jira.lsstcorp.org/browse/DM-2850 to fix it in the future.

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +184,172 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
+
+    \copydoc run
     """
-    ConfigClass = RunTransformConfig
     RunnerClass = pipeBase.ButlerInitializedTaskRunner
-    _DefaultName = "transformMeasurement"
+    ConfigClass = RunTransformConfig
+
+    # Subclasses should provide appropriate definitions for the attributes named below.
+    # Properties can be used if apprpriate.
```
TallJimbo: typo: apprpriate

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +184,172 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
+        """
+        return self.sourceType + "_schema"
+
+    @property
+    def outputType(self):
+        """!
+        The Butler dataset type for the schema of the output catalog.
+
+        By default, we prepend `transformed_` to the input source type. Subclasses may
+        customise if required.
```
TallJimbo: Standards say I have to ask you to spell "customize" like a Yank.

python/lsst/pipe/tasks/transformMeasurement.py
TallJimbo: In the meantime, it might be best to rename this parameter `outputDataset`, as that's what we typically call the named things we can get from the `Butler`, with `type` being reserved for the actually class of the saved thing (e.g. `SourceCatalog` or `ExposureF`).

python/lsst/pipe/tasks/transformMeasurement.py
```diff
@@ -152,38 +184,172 @@ class RunTransformTask(pipeBase.CmdLineTask):
...
+                         outputType=self.outputType)
 
     @pipeBase.timeMethod
     def run(self, dataRef):
         """!Transform the source catalog referred to by dataRef.
 
-        The result is both returned and written as dataset type "transformedSrc"
-        to the provided dataRef.
+        The result is both returned and written as dataset type "transformed" + the input
+        source type to the provided dataRef.
```
TallJimbo: It's actually `"transformed_" + the input source type` (note underscore), and you probably want to say "source dataset" rather than "source type".

python/lsst/pipe/tasks/transformMeasurement.py
jdswinbank: I did not realise this -- thanks. It makes the code significantly simpler. Great!

PR #9, opened 2015-06-11
Fix how PhotoCalTask calls Colorterm methods DM-2919

SimonKrughoff:
PhotoCalTask was mis-calling the Colorterm methods (by providing a filter name,
which is not wanted as of DM-2797).
Also ColortermLibrary.getColorterm was mis-handling glob expressions for the
library name (the arguments to fnmatch.fnmatch were in the wrong order).

### DIFF
--- a/python/lsst/pipe/tasks/colorterms.py
+++ b/python/lsst/pipe/tasks/colorterms.py
@@ -109,13 +109,13 @@ class ColortermDict(Config):
 
 
 class ColortermLibrary(Config):
-    """!A mapping of catalog name to ColortermDict
+    """!A mapping of photometric reference catalog name or glob to ColortermDict
 
-    This is intended to support a particular camera with a variety of reference catalogs
+    This allows photometric calibration using a variety of reference catalogs.
 
     To construct a ColortermLibrary, use keyword arguments:
     ColortermLibrary(data=dataDict)
-    where dataDict is a Python dict of reference_catalog_name: ColortermDict.
+    where dataDict is a Python dict of catalog_name_or_glob: ColortermDict
 
     For example:
     ColortermLibrary(data = {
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/tasks/colorterms.py
```diff
@@ -166,7 +166,7 @@ def getColorterm(self, filterName, refCatName, doRaise=True):
             ctDictConfig = self.data.get(refCatName)
             if ctDictConfig is None:
                 # try glob expression
-                matchList = [glob for glob in self.data if fnmatch.fnmatch(refCatName, glob)]
+                matchList = [glob for glob in self.data if fnmatch.fnmatch(glob, refCatName)]
```
SimonKrughoff: I'm not sure this was wrong.  According to documentation here:
https://docs.python.org/2/library/fnmatch.html
the glob should be the second argument.  Maybe I don't understand what is in glob.

```
>>> import fnmatch
>>> print fnmatch.fnmatch('sdss-dr8', 'sdss-*')
True
>>> print fnmatch.fnmatch('sdss-*', 'sdss-dr8')
False
```

SimonKrughoff:
I don't think I understand the first change and the second change looks fine.  I'm a little concerned that the unit test is being put in a another ticket, but if it's that much work I can be convinced.

python/lsst/pipe/tasks/colorterms.py
ktlim: I believe the intent is for the reference catalog names to be able to include metacharacters like "*", so the `refCatName` is actually the glob, whereas the variable named `glob` is the iterator across the elements in `self.data`.  I think this is a case where the variable should be renamed; if it were something like `configName` or even just `x`, this confusion would not occur.

PaulPrice:
"Also" in a commit message is usually a good indication that there are multiple components to the commit that should be split out.

python/lsst/pipe/tasks/colorterms.py
r-owen: I agree the variable needs to be renamed and I think Simon is right that the argument order is wrong. The problem is that I thought I could use a glob for config.refCatName but I can't now and probably shouldn't be able to; the globs are the catalog names in ColortermLibary (the keys of that dict).

This whole technique for managing color term data is a dreadful hack, but until we have better management of calibration data I doubt it's worth trying to clean it up beyond fixing the bug Simon noted.

PR #10, opened 2015-06-22
pipe_tasks review for DM-1945: tasks for multi-band coadd processing

TallJimbo:


python/lsst/pipe/tasks/multiBand.py
```diff
@@ -0,0 +1,666 @@
...
+        if fpSets.background:
+            backgrounds.append(fpSets.background)
+        return Struct(sources=sources, backgrounds=backgrounds)
+
+    def write(self, exposure, results, patchRef):
+        """Write out results from runDetection
+
+        @param exposure: Exposure to write out
+        @param results: Struct returned from runDetection
+        @param patchRef: data reference for patch
```
TallJimbo: Should use `"""!` here if using `@param`.

PR #11, opened 2015-06-26
pipe_tasks review for DM-436: port aperture corrections from HSC

TallJimbo:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -73,6 +73,16 @@ class CalibrateConfig(pexConfig.Config):
...
+    doMeasureApCorr = pexConfig.Field(
+        dtype = bool,
+        doc = "Compute aperture corrections?",
+        default = True,
+    )
+    doApplyApCorr = pexConfig.Field(
+        dtype = bool,
+        doc = "Apply aperture corrections? Ignored if doMeasureApCorr is False",
+        default = True,
+    )
```
TallJimbo: You should add a `validate` method to disallow `doMeasureApCorr=False and doApplyApCorr=True`, since we can't actually do that combination.

jdswinbank:

PR #15, opened 2015-08-12
Review for pipe_tasks in DM-2939: modernize astrometry interfaces

TallJimbo:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -87,6 +87,10 @@ class ImageDifferenceConfig(pexConfig.Config):
         dtype=bool,
         default=True
     )
+    astrometer = pexConfig.ConfigurableField(
+        target = measAstrom.AstrometryTask,
+        doc = "astrometry task; used to match sources to reference objects, but not to fit a WCS",
+    )
```
TallJimbo: It looks like you're touching all the places where the astrometry subtask is used here; could we rename it at the same time to `astrometry` for consistency?

There are also other places we use `astrometer`, which I'd also like to change; I prefer just calling the subtask `astrometry`, since we tend to name other tasks after the "act of doing something" (e.g. `measurement`) not "a thing that does something" (`measurer`).

python/lsst/pipe/tasks/processImage.py
```diff
@@ -232,17 +235,19 @@ def process(self, dataRef, inputExposure, enableWriteSources=True):
...
-        except AttributeError:
-            self.log.warn("Failed to find a solver in calibrate's astronomy task")
+            astrometer = self.calibrate.astrometry
+        except Exception:
+            self.log.warn("Failed to find an astrometry solver in the calibrate task")
             return None, None
 
-        astromRet = solver.useKnownWcs(sources, exposure=exposure)
-        # N.b. yes, this is what useKnownWcs calls the returned values
+        astromRet = astrometer.loadAndMatch(exposure=exposure, sourceCat=sources)
```
TallJimbo: Another place where we could replace `astrometer` with `astrometry`.

python/lsst/pipe/tasks/imageDifference.py
r-owen: OK. Done in pipe_tasks, and that exposed one instance where I had failed to update an interface. obs_subaru has an instance of a class that has an astrometer subtask and an astrometry method; that will take some thought.

TallJimbo: Do recall what class that was?  I think now obs_subaru is now integrated into the stack enough we should at least just consider changing it here.

PaulPrice: It's for using hscAstrom instead of meas_astrom.  That shouldn't be necessary any more.  Don't worry about it.  If we need it for HSC, I'll work around you.

r-owen: To answer Jim's question: it is SubaruAstrometryTask. I confess I haven't pulled obs_subaru in awhile. Paul: thank you!

PR #17, opened 2015-09-08
Review for DM-3651, pipe_tasks - bookkeeping improvements to makeDiscreteSkyMap

TallJimbo:


python/lsst/pipe/tasks/makeDiscreteSkyMap.py
```diff
@@ -62,21 +62,30 @@ def setDefaults(self):
...
-        # and checking it means we can't overwrite or append to one once we've
-        # written it.
-        return True
-
     def __call__(self, args):
+        """
+        @param args     Arguments for Task.run()
+
+        @return:
+        - None if doReturnResults false
```
TallJimbo: Maybe say "`self.doReturnResults`"; took me a moment to remember where that setting comes from.

tests/testMakeDiscreteSkyMap.py
```diff
@@ -0,0 +1,97 @@
...
+            config = self.config,
+            doReturnResults = True,
+        )
+        self.assertEqual(len(retVal.resultList), 1)
+        skyMap = retVal.resultList[0].result.skyMap
+        self.assertEqual(type(skyMap), DiscreteSkyMap)
+        self.assertEqual(len(skyMap), 1)
+        tractInfo = skyMap[0]
+        self.assertEqual(tractInfo.getId(), 0)
+        self.assertEqual(tractInfo.getNumPatches(), Extent2I(3, 3))
```
TallJimbo: Could we also test that the bounding boxes for the images are completely contained by the new tract?

r-owen: Good idea. I'll keep a list of the sky coords for all calexp bbox corners, then make sure all those points are in the tract

python/lsst/pipe/tasks/makeDiscreteSkyMap.py
r-owen: OK. I don't think "self" helps as much as talking about the constructor argument, but self is easy, short and precise (whereas one could override the constructor...)

PR #19, opened 2015-09-28
Tickets/dm-3766: A new Task ingestCalibs

hchiang2:
ingest.py is modified slightly for more general use, and ingestCalibs.py is created for parsing through calibration data and creating a calibration registry.

### DIFF
--- a/python/lsst/pipe/tasks/ingestCalibs.py
+++ b/python/lsst/pipe/tasks/ingestCalibs.py
@@ -1,0 +1,169 @@
+import collections
+import datetime
+import sqlite3
+import lsst.afw.image as afwImage
+from lsst.pex.config import Config, ListField, ConfigurableField
+from lsst.pipe.base import ArgumentParser
+from lsst.pipe.tasks.ingest import RegisterTask, ParseTask, RegisterConfig, IngestTask
+
+
+def _convertToDate(dateString):
+    """Convert a string into a date object, or return None
+    when the date string cannot be converted with format %Y-%m-%d
+    """
+    try:
+        return datetime.datetime.strptime(dateString, "%Y-%m-%d").date()
+    except ValueError:
+        return None
+
+
+class CalibsParseTask(ParseTask):
+    """Task that will parse the filename and/or its contents to get the
+    required information to populate the calibration registry."""
+    def getCalibType(self, filename):
+        """Return a a known calibration dataset type using
+        the observation type in the header keyword OBSTYPE
+
+        @param filename: Input filename
+        """
+        md = afwImage.readMetadata(filename, self.config.hdu)
+        if not md.exists("OBSTYPE"):
+            raise RuntimeError("Unable to find the required header keyword OBSTYPE")
+        obstype = md.get("OBSTYPE").strip().lower()
+        if "flat" in obstype:
+            obstype = "flat"
+        elif "zero" in obstype or "bias" in obstype:
+            obstype = "bias"
+        elif "dark" in obstype:
+            obstype = "dark"
+        elif "fringe" in obstype:
+            obstype = "fringe"
+        return obstype
+
+
+class CalibsRegisterConfig(RegisterConfig):
+    """Configuration for the CalibsRegisterTask"""
+    tables = ListField(dtype=str, default=["bias", "dark", "flat", "fringe"],
+                       doc="Name of tables")
+
+
+class CalibsRegisterTask(RegisterTask):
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -0,0 +1,164 @@
+import collections
+import datetime
+import sqlite3
+from lsst.pex.config import Config, ListField, ConfigurableField
+from lsst.pipe.base import Task, ArgumentParser
+from lsst.pipe.tasks.ingest import RegisterTask, ParseTask, RegisterConfig
+
+def _convertToDate(dateString):
+    """Convert a string into a date"""
```
timj: It's converting a string into a date object. "date" is ambiguous as it could just mean that it is finding the date in the string and returning the date as a string.

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -0,0 +1,164 @@
...
+        return None
+
+class CalibsParseTask(ParseTask):
+    """Task that will parse the filename and/or its contents to get the
+    required information to populate the calibration registry."""
+    def translate_obstype(self, md):
+        """Convert the observation type in the header
+        to a known calibration dataset type.
+
+        @param md: FITS header metadata
```
timj: what type of thing is md? Is it a dict? dict-like thing? PropertyList?

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        obstype = md.get("OBSTYPE").strip().lower()
+        if "flat" in obstype:
+            obstype = "flat"
+        elif "zero" in obstype or "bias" in obstype:
+            obstype = "bias"
+        elif "dark" in obstype:
+            obstype = "dark"
+        elif "fringe" in obstype:
+            obstype = "fringe"
```
timj: do we care if none of these checks match?

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -0,0 +1,164 @@
...
+class CalibsParseTask(ParseTask):
+    """Task that will parse the filename and/or its contents to get the
+    required information to populate the calibration registry."""
+    def translate_obstype(self, md):
+        """Convert the observation type in the header
+        to a known calibration dataset type.
+
+        @param md: FITS header metadata
+        """
+        obstype = md.get("OBSTYPE").strip().lower()
```
timj: can OBSTYPE be missing?

python/lsst/pipe/tasks/ingestCalibs.py
hchiang2: I'd say no.  That file would just be ignored and not written into the calibration registry.

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -0,0 +1,164 @@
...
+        sql += " ORDER BY calibDate"
+        cursor = conn.cursor()
+        cursor.execute(sql)
+        rows = cursor.fetchall()
+        valids = collections.OrderedDict([(_convertToDate(row["calibDate"]), 
+                                           [_convertToDate(row["validStart"]), 
+                                            _convertToDate(row["validEnd"])]) for row in rows])
+        dates = valids.keys()
+        numDates = len(dates)
+        midpoints = [ t1 + (t2 - t1)//2 for t1, t2 in zip(dates[:numDates-1], dates[1:]) ]
```
timj: If `_convertToDate` can return `None` this will break.

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -0,0 +1,164 @@
...
+                                                      - datetime.timedelta(args.validity)).isoformat()
+                                info['validEnd'] = (_convertToDate(info['calibDate'])
+                                                    + datetime.timedelta(args.validity)).isoformat()
+                            except TypeError:
+                                self.log.warn("Skipped setting validity period of %s" % args.validity)
+                        self.register.addRow(registry, info, dryrun=args.dryrun, create=args.create, table=table)
+                    else:
+                        self.log.warn("Skipped adding %s of obstype %s to registry" % (infile, info["obstype"]))
+            if args.dryrun:
+                print "Would update validity ranges to resolve overlaps"
```
timj: This should be a print function (add `from __future__ import print_function` at top of file), or an `info` log message.

PR #20, opened 2015-09-28
Tickets/DM-2915: Port safe coadd clipping from HSC

rearmstr:


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -130,6 +133,7 @@ def __init__(self, *args, **kwargs):
         self.makeSubtask("interpImage")
         self.makeSubtask("matchBackgrounds")
         self.makeSubtask("scaleZeroPoint")
+	self.testvar = 10
```
rearmstr: Remove this

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
             )
             self.refList.append(dataRef)
 
+def countMaskFromFootprint(mask,footprint, bitmask, ignoreMask):
```
rearmstr: add comma after mask

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
...
+    def __init__(self, *args, **kwargs):
+        super(SafeClipAssembleCoaddTask, self).__init__(*args, **kwargs)
+        schema = afwTable.SourceTable.makeMinimalSchema()
+        self.makeSubtask("clipDetection", schema=schema)
+
+    def assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgModelList):
+        """Assemble the coadd for a region
+
+        Identify clipped regions by detecting objects on the difference between unclipped and clipped coadd
+        and then flag these regions on the individual vists so they are ignored in the coaddtion process
```
rearmstr: typo "coaddition"

python/lsst/pipe/tasks/assembleCoadd.py
timj: do you mean add a space?

bin/assembleCoadd.py
```diff
@@ -21,6 +21,28 @@
...
+import argparse
 
-AssembleCoaddTask.parseAndRun()
+parser = argparse.ArgumentParser(add_help=False)
+parser.add_argument('--legacyCoadd', action='store_true', default=False)
+parser.add_argument('-h','--help', action='store_true',default=False)
+result, extra = parser.parse_known_args(sys.argv)
+
+if result.help:
+    print('Meta Arguments:\n  --legacyCoadd \t\tAn option to run with the original assembleCoadd '+\
```
timj: does this not use triple-quote because of the tabs? Also, I don't think the `+` is needed because python concatenates quoted strings automatically:
```python
a = "abc" "def"
print(a)
```
will print "abcdef". I also think a space is missing between the final two strings.

python/lsst/pipe/tasks/assembleCoadd.py
rearmstr: Yes, add a space.

bin/assembleCoadd.py
```diff
@@ -21,6 +21,28 @@
...
+	  'config.measurement.plugins["base_PixelFlags"].any = [] must be set\n')
+    SafeClipAssembleCoaddTask.parseAndRun(args=['--help']+extra[1:])
+
+if result.legacyCoadd:
+    legacy_message = '''
+    Remember, If you use the legacy adder you must set
+    config.measurement.plugins['base_PixelFlags'].any = []
+    so that pixel flags will not look for a non existent mask plane
+    '''
+    print(legacy_message)
```
timj: I know it makes little difference for one argument, but can you add `from __future__ import print_function` at the top so that when people see the print here they will understand they can use modern syntax.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -414,7 +427,7 @@ def assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgInfoL
         for subBBox in _subBBoxIter(skyInfo.bbox, subregionSize):
             try:
                 self.assembleSubregion(coaddExposure, subBBox, tempExpRefList, imageScalerList,
-                                       weightList, bgInfoList, statsFlags, statsCtrl)
+                                       weightList, bgInfoList, altMaskList, statsFlags, statsCtrl)
             except Exception, e:
```
timj: This was not changed, but since we are editing this file can we please update the except here to be:
```python
except Exception as e:
```
A distinct commit is fine and there are 5 in this file.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
...
+            tmpExp = tmpExpRef.get(self.getTempExpDatasetName(), immediate=True)
+            tmpExpMask = tmpExp.getMaskedImage().getMask()
+            tempExpClipList.append(tmpExpMask)
+
+        for footprint in footprints.getFootprints():
+            nPixel = footprint.getArea()
+            overlap = [] # hold the overlap with each visit
+            maskList = [] # which visit mask match
+            indexList = []# index of visit in global list
+            for i,tmpExpMask in enumerate(tempExpClipList):
```
timj: add space between `i` and `tmpExpMask`.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
...
+
+                # If we have more bad pixels than detection skip
+                if ignore > overlapDet or totPixel <= 0.5*nPixel or overlapDet == 0:
+                    continue
+                overlap.append(overlapDet/float(totPixel))
+                maskList.append(tmpExpMask)
+                indexList.append(i)
+
+            overlap = numpy.array(overlap)
+            if len(overlap) == 0:
```
timj: consider writing `if not len(overlap):`

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
...
+    """
+    bbox = footprint.getBBox()
+    bbox.clip(mask.getBBox(afwImage.PARENT))
+    fp = afwImage.MaskU(bbox)
+    subMask = mask.Factory(mask, bbox, afwImage.PARENT)
+    afwDet.setMaskFromFootprint(fp, footprint, bitmask)
+    return numpy.logical_and((subMask.getArray() & fp.getArray()) > 0,
+                             (subMask.getArray() & ignoreMask) == 0).sum()
+
+
```
rearmstr: SafeClipAssembleCoaddConfig should have some documentation about what it is doing.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
...
+            keep = False   # Should this footprint be marked as clipped?
+            keepIndex = [] # Which tempExps does the clipped footprint belong to
+
+            # If footprint only has one overlap use a lower threshold
+            if len(overlap) == 1:
+                if overlap[0] > self.config.minClipFootOverlapSingle:
+                    keep = True
+                    keepIndex = [0]
+            else:
+                # This is the general case for most clipped objects where only visit is responsible
```
rearmstr: Clarify the comment with "where only one visit is should be clipped".

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
...
+
+        maskClipBig = maskClip.Factory(mask.getBBox(afwImage.PARENT))
+        afwDet.setMaskFromFootprintList(maskClipBig, bigFootprints, maskClipValue)
+        maskClip |= maskClipBig
+
+        # Now combine again using mean, ignoring CLIPPED pixels
+        badMaskPlanes = self.config.badMaskPlanes[:]
+        badMaskPlanes.append("CLIPPED")
+        badPixelMask = afwImage.MaskU.getPlaneBitMask(badMaskPlanes)
+        coaddExp = AssembleCoaddTask.assemble(self,skyInfo, tempExpRefList, imageScalerList, weightList, bgModelList,
```
rearmstr: This line goes beyond the 110 column limit.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +624,280 @@ def makeDataRefList(self, namespace):
...
+        @param weightList: List of weights
+        @param bgModelList: List of background models from background matching
+        @return Difference image of unclipped and clipped coadd wrapped in an Exposure
+        """
+        # Build the unclipped coadd
+        coaddMean = AssembleCoaddTask.assemble(self, skyInfo, tempExpRefList, imageScalerList,weightList, bgModelList, doClip=False)
+
+        # Build the clipped coadd
+        coaddClip = AssembleCoaddTask.assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgModelList, doClip=True)
+
```
rearmstr: Both calls to assemble have lines that are too long.

jdswinbank:
A couple of generic comments on the commit messages:
* Please write in the imperative ("add", not "added");
* Please reword the summary of 4b03a5. The commit itself is an implementation of safe coadd clipping; the fact that it happens to be ported from HSC is a detail which we need to keep track of for book-keeping now, but in a few years time will be irrelevant. Further, that commit is currently attributed to Bob, and, while he did the work on HSC, he didn't do the porting.

bin/assembleCoadd.py
```diff
@@ -20,7 +20,34 @@
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
+from __future__ import print_function
 import lsst.pex.logging as pexLog
```
jdswinbank: `pexLog` is unused. You're not changing that line here, but maybe take this opportunity to remove it?

bin/assembleCoadd.py
```diff
@@ -20,7 +20,34 @@
...
+
+if result.help:
+    legacy_message = '''
+Meta Arguments:
+  --legacyCoadd         An option to run with the original assembleCoadd task which will
+                        not attempt to do any safe clipping.  If run, the config option
+                        'config.measurement.plugins["base_PixelFlags"].any = [] must be set
+'''
+    print(legacy_message)
+    SafeClipAssembleCoaddTask.parseAndRun(args=['--help']+extra[1:])
```
jdswinbank: Are you guaranteed to always in perpetuity get the same help message from `SafeClipAssembleCoaddTask` as from `AssembleCoaddTask`? If not, don't you have to check for the `legacyCoadd` argument and despatch the `--help` argument to the appropriate task?

bin/assembleCoadd.py
```diff
@@ -20,7 +20,34 @@
...
+if result.legacyCoadd:
+    legacy_message = '''
+    Remember, If you use the legacy adder you must set
+    config.measurement.plugins['base_PixelFlags'].any = []
+    so that pixel flags will not look for a non existent mask plane
+    '''
+    print(legacy_message)
+    AssembleCoaddTask.parseAndRun(args=extra[1:])
+else:
+    SafeClipAssembleCoaddTask.parseAndRun(args=extra[1:])
```
jdswinbank: It's unfortunate that both `assembleCoadd.py --help` and `assembleCoadd.py` produce help messages, but they aren't the same. Is it possible to catch usage errors from these tasks and add your "meta arguments" help block to the output?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -379,16 +382,21 @@ def assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgInfoL
         @param tempExpRefList: List of data references to tempExp
         @param imageScalerList: List of image scalers
         @param weightList: List of weights
-        @param bgInfoList: List of background data from background matching
+        @param bgInfoList: List of background data from background matching, or None
+        @param altMaskList: List of alternate masks to use than those on disk, or None
```
jdswinbank: I'm having difficulty parsing that description of `altMaskList`. I think the English is wrong (I'm struggling here -- because you can't use "alternate" in this sense in British English -- but I think you want "alternate to" rather than "alternate than"). I guess that by "those on disk" you mean "those that were stored with the tempExp", right? The alternatives could be stored on disk elsewhere, one assumes. Could you clarify the wording please?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -396,7 +404,7 @@ def assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgInfoL
             bit = afwImage.MaskU.getMaskPlane(plane)
             statsCtrl.setMaskPropagationThreshold(bit, threshold)
 
-        if self.config.doSigmaClip:
+        if doClip:
```
jdswinbank: I see no other reference to the `doSigmaClip` configuration option -- it looks like this change hard-codes it to be ignored. I suspect that's a mistake; if so, please fix it. If not, please justify it and remove the config option.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -464,6 +475,7 @@ def assembleSubregion(self, coaddExposure, bbox, tempExpRefList, imageScalerList
         @param imageScalerList: List of image scalers
         @param weightList: List of weights
         @param bgInfoList: List of background data from background matching
+        @param altMaskList: List alternate masks to use than on disk
```
jdswinbank: See comment above.

python/lsst/pipe/tasks/assembleCoadd.py
jdswinbank: I agree with this in principle. In practice, I'm not sure it's any less documented than the old `AssembleCoaddTask` (neither of them appears in [the Doxygen task list](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/group___l_s_s_t__task__documentation.html) for example. I don't properly documenting everything is a prerequisite for merging given that this is an HSC port, but we do need a "technical debt" issue to record that it needs fixing in future.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+    clipDetection = pexConfig.ConfigurableField(target=SourceDetectionTask,
+                                      doc="Detect sources on difference between unclipped and clipped coadd")
+    minClipFootOverlap = pexConfig.Field(
+        doc = "Minimum fractional overlap of clipped footprint with visit DETECTED to be clipped",
+        dtype = float,
+        default = 0.65
+    )
+    minClipFootOverlapSingle = pexConfig.Field(
+        doc = "Minimum fractional overlap of clipped footprint with visit DETECTED to be " \
+            "clipped when only one visit overlaps",
```
jdswinbank: Add a couple of extra spaces to help the strings line up, here and in the other `Field`s defined below.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+        self.clipDetection.reEstimateBackground = False
+        self.clipDetection.returnOriginalFootprints = False
+        self.clipDetection.thresholdPolarity = "both"
+        self.clipDetection.thresholdValue = 2
+        self.clipDetection.nSigmaToGrow = 4
+        self.clipDetection.minPixels = 4
+        self.clipDetection.isotropicGrow = True
+        self.clipDetection.thresholdType = "pixel_stdev"
+        self.sigmaClip = 1.5
+        self.clipIter = 3
```
jdswinbank: It would be nice to know what the rationale behind these numbers is. I seem to remember Bob telling us that he arrived at them by experiment, which is fine, but we should write that down somewhere (if it's correct -- if it's not, we should record what really happened).

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+    identifying artifacts that are found in only one or two visits.  A different threshold is applied
+    depending on the number overlapping visits.  The current thresholds work well at removing a majority
+    of artifacts.
+
+    """
+    ConfigClass = SafeClipAssembleCoaddConfig
+    _DefaultName = "safeClipAssembleCoadd"
+
+    def __init__(self, *args, **kwargs):
+        super(SafeClipAssembleCoaddTask, self).__init__(*args, **kwargs)
```
jdswinbank: The [coding standards say](https://confluence.lsstcorp.org/display/LDMDG/Python+Coding+Standard#PythonCodingStandard-DesigningforInheritance):
> 'super' SHOULD NOT be used unless the author really understands the implications (e.g. in a well-understood  multiple inheritance hierarchy).
Do we really need it here?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+
+        # Create mask of the current clipped footprints
+        maskClip = mask.Factory(mask.getBBox(afwImage.PARENT))
+        afwDet.setMaskFromFootprintList(maskClip, result.clipFootprints, maskClipValue)
+
+        maskClipBig = maskClip.Factory(mask.getBBox(afwImage.PARENT))
+        afwDet.setMaskFromFootprintList(maskClipBig, bigFootprints, maskClipValue)
+        maskClip |= maskClipBig
+
+        # Now combine again using mean, ignoring CLIPPED pixels
```
jdswinbank: "Combine again using mean" could be more helpful. Specifically, I think you mean you are passing `doClip=False` to `AssembleCoaddTask.assemble()` a few lines down.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+        badMaskPlanes = self.config.badMaskPlanes[:]
+        badMaskPlanes.append("CLIPPED")
+        badPixelMask = afwImage.MaskU.getPlaneBitMask(badMaskPlanes)
+        coaddExp = AssembleCoaddTask.assemble(self,skyInfo, tempExpRefList, imageScalerList, weightList,
```
jdswinbank: Needs a space after `self`.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+
+        clipFootprints = []
+        clipIndices = []
+
+        # build a list with a mask for each visit which can be modified with clipping information
+        tempExpClipList = []
+        for tmpExpRef in tempExpRefList:
+            tmpExp = tmpExpRef.get(self.getTempExpDatasetName(), immediate=True)
+            tmpExpMask = tmpExp.getMaskedImage().getMask()
+            tempExpClipList.append(tmpExpMask)
```
jdswinbank: Might be easier to read with a list comprehension? Not a big deal, though.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+
+        for footprint in footprints.getFootprints():
+            nPixel = footprint.getArea()
+            overlap = [] # hold the overlap with each visit
+            maskList = [] # which visit mask match
+            indexList = []# index of visit in global list
+            for i, tmpExpMask in enumerate(tempExpClipList):
+                # what is the overlap with the footprint
```
jdswinbank: "What is the overlap with the footprint"**?**

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+                ignore = countMaskFromFootprint(tmpExpMask, footprint, ignoreMask, 0x0)
+                overlapDet = countMaskFromFootprint(tmpExpMask, footprint, maskDetValue,
+                                                    ignoreMask)
```
jdswinbank: You don't need the newline in the middle of the above.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
...
+            clipFootprints.append(footprint)
+
+        return pipeBase.Struct(clipFootprints=clipFootprints, clipIndices=clipIndices,
+                               tempExpClipList=tempExpClipList)
+
+
+    def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipValue, maskDetValue):
+        """Find footprints from individual tempExp footprints for large footprints
+
+        @param tempExpClipList: List of tempExp mask's with clipping information
```
jdswinbank: Typo: please remove the apostrophe.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -478,6 +478,9 @@ class MeasureMergedCoaddSourcesConfig(Config):
     def setDefaults(self):
         Config.setDefaults(self)
         self.deblend.propagateAllPeaks = True
+        # The following line must be set if clipped pixel flags are to be added to the output table
+        # The clipped mask plane is added by running the clipped coadd task
```
jdswinbank: Please give the specific name of "the clipped coadd task".

python/lsst/pipe/tasks/processCoadd.py
```diff
@@ -61,6 +61,10 @@ def setDefaults(self):
         self.measurement.doApplyApCorr = "noButWarn"
         self.doDeblend = True
         self.deblend.maxNumberOfPeaks = 20
+        # The following line must be set if clipped pixel flags are to be added to the output table
+        # The clipped mask plane is added by running the clipped coadd task
```
jdswinbank: Please give the specific name of "the clipped coadd task".

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
             )
             self.refList.append(dataRef)
 
+def countMaskFromFootprint(mask, footprint, bitmask, ignoreMask):
+    """Function to count the number of pixels with a specific mask in a footprint
+    """
+    bbox = footprint.getBBox()
+    bbox.clip(mask.getBBox(afwImage.PARENT))
+    fp = afwImage.MaskU(bbox)
+    subMask = mask.Factory(mask, bbox, afwImage.PARENT)
```
RobertLuptonTheGood: I think this is equivalent to `subMask = mask[bbox].clone()` as PARENT is the default (see also DM-4132) which is perhaps cleaner.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -605,3 +622,298 @@ def makeDataRefList(self, namespace):
             )
             self.refList.append(dataRef)
 
+def countMaskFromFootprint(mask, footprint, bitmask, ignoreMask):
```
RobertLuptonTheGood: You can also write this as

    def countMaskFromFootprint(mask, footprint, bitmask, ignoreMask):
        footprint = afwDetect.Footprint(footprint)   # makes a copy
        footprint.intersect(mask, bitmask)
        footprint.intersect(mask, ~ignoreMask)
        return footprint.getNpix()

which is C++ and doesn't require a temp Mask (it makes a temp Footprint instead -- swings and roundabouts)

python/lsst/pipe/tasks/assembleCoadd.py
natelust: I will have to talk to bob about this

rearmstr: It should be reworded as John indicates to mean other than those stored with the tempExp.

python/lsst/pipe/tasks/assembleCoadd.py
natelust: Again, let me talk with bob

PR #22, opened 2015-10-27
Review for DM-3693 (HSC backport: allow photometric and astrometric calibrations to be required)

jdswinbank:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -453,13 +469,22 @@ def run(self, exposure, defects=None, idFactory=None):
...
+                    astromRet = self.astrometry.run(exposure, sources1)
+                    matches = astromRet.matches
+                except RuntimeError as e:
+                    if self.config.requireAstrometry:
+                        raise
+                    self.log.warn("Unable to perform astrometry (%s): attempting to proceed" % e)
+                finally:
+                    # Restore original Wcs: we're going to repeat the astrometry later, and if it succeeded
+                    # this time, running it again with the same basic setup means it should succeed again.
+                    exposure.setWcs(origWcs)
```
jdswinbank: Is this effectively the same as calling astrometry with `forceKnownWcs` set to `True`? If so, would that be preferable? (It means you'd have two astrometry tasks with different configs, but it's a more explicit use of the API.)

PaulPrice: No, we want to solve the astrometry (so `forceKnownWcs=False`), but we want to undo it when we're done.

laurenam: No, it's not.  An attempt to compute astrometry is actually being made to get the matches.  This reverts to the original after that matching, but it does not simply use any WCS already attached to the image.

jdswinbank:
Ugh; I think I've been commenting in the wrong place (on the individual commits rather than the overall diff). Sorry for the GitHub fail!.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -344,19 +358,22 @@ def DebugInfo(name):
...
 
         # the calibrate Source Catalog is divided into two catalogs to allow measurement to be run twice
         # schema1 contains everything except what is added by the second measurement task.
         # Before the second measurement task is run, self.schemaMapper transforms the sources into
-        # the final output schema, at the same time renaming the measurement fields to "initial_" 
-        self.schema1 = afwTable.SourceTable.makeMinimalSchema()
+        # the final output schema, at the same time renaming the measurement fields to "initial_"
+        if schema is None:
+            schema = afwTable.SourceTable.makeMinimalSchema()
+        self.schema1 = schema
```
jdswinbank: @jdswinbank:

> Is this change necessary? It doesn't look like either ProcessCcd or ProcessImage before it tries to pass a a schema to CalibrateTask. Even if being able to do so is a good idea in principle, I don't understand what it's doing in this commit.

@laurenam 

> I agree. I will remove it unless @PaulPrice pipes up with his original intention/motivation.

python/lsst/pipe/tasks/processCcd.py
```diff
@@ -12,30 +12,56 @@
...
+    doWriteCalibrate = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc = "Write calibration results?"
+    )
+    doWriteCalibrateMatches = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc = "Write icSrc to reference matches?"
+    )
```
jdswinbank: I don't think we should have spaces in the kwargs, but I note the same pattern elsewhere in the file so I guess consistency is a virtue.

python/lsst/pipe/tasks/processCcd.py
```diff
@@ -48,20 +74,51 @@ class ProcessCcdTask(ProcessImageTask):
...
+            self.makeSubtask("isr")
+        self.makeSubtask("calibrate")
+        # Setup our schema by starting with fields we want to propagate from icSrc.
+        calibSchema = self.calibrate.schema
+        self.schemaMapper = afwTable.SchemaMapper(calibSchema, self.schema)
+
+        # Add fields needed to identify stars used in the calibration step
+        self.calibSourceKey = self.schemaMapper.addOutputField(
+            afwTable.Field["Flag"]("calib_detected", "Source was detected as an icSrc")
+            )
```
jdswinbank: Funky indentation.

python/lsst/pipe/tasks/processImage.py
```diff
@@ -240,68 +185,15 @@ def matchSources(self, exposure, sources):
...
         """
-        try:
-            astrometry = self.calibrate.astrometry
-        except Exception:
-            self.log.warn("Failed to find an astrometry solver in the calibrate task")
-            return None, None
-
-        astromRet = astrometry.loadAndMatch(exposure=exposure, sourceCat=sources)
+        astromRet = self.getAstrometer().loadAndMatch(exposure=exposure, sourceCat=sources)
+        # N.b. yes, this is what useKnownWcs calls the returned values
```
jdswinbank: But we're not calling anything called useKnownWcs here; it seems like that's an implementation detail of ANetBasicAstrometryTask, but we're not obviously using that either.

python/lsst/pipe/tasks/processImage.py
```diff
@@ -132,101 +111,67 @@ def makeIdFactory(self, sensorRef):
...
+        @param backgrounds: afwMath.BackgroundList to be appended to
         @param enableWriteSources: if True then writing sources is allowed.
             Set False if you need to postprocess sources before writing them.
 
         @return pipe_base Struct containing these fields:
-        - postIsrExposure: exposure after ISR performed if calib.doIsr or config.doCalibrate, else None
-        - exposure: calibrated exposure (calexp): as computed if config.doCalibrate,
-            else as upersisted and updated if config.doDetection, else None
-        - calib: object returned by calibration process if config.doCalibrate, else None
+        - exposure: input exposure (as modified in the course of processing)
```
jdswinbank: Not sure exactly what that means -- "_will be_ modified in the course of processing"?

python/lsst/pipe/tasks/processImage.py
```diff
@@ -100,25 +92,12 @@ class ProcessImageTask(pipeBase.CmdLineTask):
...
-            afwTable.Field["Flag"]("calib_detected", "Source was detected as an icSrc")
-            )
-
-        for key in self.calibrate.getCalibKeys():
-            self.schemaMapper.addMapping(key)
-        self.schema = self.schemaMapper.getOutputSchema()
         self.algMetadata = dafBase.PropertyList()
+        if schema is None:
+            schema = afwTable.SourceTable.makeMinimalSchema()
+        self.schema = schema
```
jdswinbank: Same question as earlier -- I don't object to the schema dance per se, but I don't see why it's folded into this commit.

python/lsst/pipe/tasks/processImage.py
```diff
@@ -132,101 +111,67 @@ def makeIdFactory(self, sensorRef):
...
-                self.log.warn("The calibration task did not return any backgrounds. " +
-                    "Any background subtracted in the calibration process cannot be persisted.")
-        else:
-            calib = None
+        if backgrounds is None:
+            backgrounds = afwMath.BackgroundList()
 
         if self.config.doDetection:
+            if idFactory is None:
+                idFactory = self.makeIdFactory(dataRef)
```
jdswinbank: Same pattern as we're seeing for schemas?

laurenam: No.  I added this as the introduction of idFactory in processCcd.py meant that it was being computed once (I discovered this when the demo failed due to mismatched object IDs).  Here, there actually is a case where the idFactory is being passed in (and I thought I should still allow for the case that it may not be from another caller).

jdswinbank: Just testing!

(Thanks)

python/lsst/pipe/tasks/processImage.py
laurenam: Hhhmmm...this was clear to me.  The inputExposure gets returned with the modifications that occur during this processing step (I believe it's mostly getting things like detection masks and metadata attached to it).  How about:
"with modifications made during this processing stage"?

jdswinbank: I already changed that one to "ignore" -- I misread.

python/lsst/pipe/tasks/processCcd.py
laurenam: I'm happy to maintain consistency by changing all occurrences to the desired format :)

python/lsst/pipe/tasks/calibrate.py
jdswinbank: Thanks both.

python/lsst/pipe/tasks/processImage.py
```diff
@@ -132,101 +111,67 @@ def makeIdFactory(self, sensorRef):
...
-            srcMatches = None; srcMatchMeta = None
+        srcMatches = None; srcMatchMeta = None
+        if self.config.doMeasurement and self.config.doWriteSourceMatches:
+            self.log.info("Matching src to reference catalogue")
+            try:
+                srcMatches, srcMatchMeta = self.matchSources(inputExposure, sources)
+                normalizedSrcMatches = afwTable.packMatches(srcMatches)
+                normalizedSrcMatches.table.setMetadata(srcMatchMeta)
+                dataRef.put(normalizedSrcMatches, self.dataPrefix + "srcMatch")
+            except Exception as e:
```
jdswinbank: Is there a more specific exception we could catch instead, or is this really a catch-anything case?

python/lsst/pipe/tasks/calibrate.py
jdswinbank: @PaulPrice 
> I don't recall the intention, and I don't see this feature being used anywhere in the HSC stack. Nevertheless, I think it's a feature that might be useful, and should be included in a separate commit.

laurenam: Making it a separate commit would be rearranging the HSC history, so if we are keeping this feature, I think the commit structure should remain as is.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -129,6 +139,10 @@ def validate(self):
         pexConfig.Config.validate(self)
         if self.doPhotoCal and not self.doAstrometry:
             raise ValueError("Cannot do photometric calibration without doing astrometric matching")
+        if self.requireAstrometry and not self.doAstrometry:
+            raise ValueError("Astrometric solution required, but not activated")
```
jdswinbank: The two lines above should really be part of 53ae929 rather than ce2dc85, but it's a trivial point.

python/lsst/pipe/tasks/processImage.py
PaulPrice: I expect this is due to changes to astrometry on the LSST side.  Suggest removing the comment.

python/lsst/pipe/tasks/calibrate.py
jdswinbank: I'd suggest keeping it, but adding a note to the commit message noting that it's included. Not perfect, but better than dropping it or silently including it.

laurenam: Will do.

python/lsst/pipe/tasks/processImage.py
PaulPrice: This is indeed intended to catch anything.

python/lsst/pipe/tasks/processImage.py
laurenam: Indeed.  It's loadAndMatch that calls useKnownWcs now: https://github.com/lsst/meas_astrom/blob/60db4916ce4f3a3d19740ff81a384e2cdcc06f24/python/lsst/meas/astrom/anetAstrometry.py#L310
I also agree the comment should just be removed.

python/lsst/pipe/tasks/processCcd.py
```diff
@@ -72,15 +129,112 @@ def run(self, sensorRef):
...
-        result = self.process(sensorRef, postIsrExposure)
-        result.postIsrExposure = postIsrExposure
-        return result
+        result = self.process(sensorRef, calExposure, idFactory=idFactory)
+
+        # combine the differential background we estimated while detecting the main src catalog
+        # with the background estimated in the calibrate step
+        for bg in result.backgrounds:
+            backgrounds.append(bg)
+        result.backgrounds = backgrounds
```
jdswinbank: This seems convoluted, and I'm not sure if it needs to be. Is the ordering of the backgrounds within the list important? If so, can we add a comment to say so? If not, can we just `result.backgrounds.extend(backgrounds)` and then use `result.backgrounds` in the calls to `write` and `restoreBackgrounds` below?

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -531,16 +569,17 @@ def extractMagArrays(self, matches, filterName, sourceKeys):
...
     @pipeBase.timeMethod
-    def run(self, exposure, matches):
+    def run(self, exposure, matches, doSelectUnresolved=True):
         """!Do photometric calibration - select matches to use and (possibly iteratively) compute
         the zero point.
 
-        \param[in]  exposure   Exposure upon which the sources in the matches were detected.
-        \param[in]  matches    Input lsst.afw.table.ReferenceMatchVector
+        \param[in]  exposure            Exposure upon which the sources in the matches were detected.
+        \param[in]  matches             Input lsst.afw.table.ReferenceMatchVector
```
jdswinbank: Whitespace.

python/lsst/pipe/tasks/calibrate.py
laurenam: Again, that would re-write the HSC history, so I'm inclined to leave it as is.

PR #25, opened 2016-01-13
Review for DM-4797: Correct call to superclass.

jdswinbank:


natelust:
Both the fix, and the additional unit test look appropriate to me.

PR #29, opened 2016-02-22
Review for DM-5161: Support a full background model when detecting cosmic rays 

jdswinbank:


python/lsst/pipe/tasks/repair.py
```diff
@@ -234,22 +234,40 @@ def cosmicRay(self, exposure, keepCRs=None):
         except Exception:
             pass
 
-        mi = exposure.getMaskedImage()
-        bg = afwMath.makeStatistics(mi, afwMath.MEDIAN).getValue()
+        exposure0 = exposure            # initial value of exposure
+        binSize = self.config.cosmicray.background.binSize
+        nx, ny = exposure.getWidth()/binSize, exposure.getHeight()/binSize
+        if nx*ny <= 1:
```
jdswinbank: I see the logic here, but: what happens if we call ``estimateBackground`` even when ``nx*ny`` is less than 1?

* If it does just use the median, can we just call it here? It simplifies the logic since we don't need to handle this case specially.
* If it doesn't just use the median, what's the justification for using the same technique here as in ``estimateBackground`` for small bins, and a different technique for large bins?

python/lsst/pipe/tasks/repair.py
```diff
@@ -234,22 +234,40 @@ def cosmicRay(self, exposure, keepCRs=None):
...
+            bg = afwMath.makeStatistics(exposure.getMaskedImage(), afwMath.MEDIAN).getValue()
+            bkgd = None
+        else:
+            exposure = exposure.Factory(exposure, True)
+            bkgd, exposure = measAlg.estimateBackground(exposure, self.config.cosmicray.background,
+                                                      subtract=True)
```
jdswinbank: Wonky indentation.

python/lsst/pipe/tasks/repair.py
```diff
@@ -234,22 +234,40 @@ def cosmicRay(self, exposure, keepCRs=None):
...
             pass
 
-        mi = exposure.getMaskedImage()
-        bg = afwMath.makeStatistics(mi, afwMath.MEDIAN).getValue()
+        exposure0 = exposure            # initial value of exposure
+        binSize = self.config.cosmicray.background.binSize
+        nx, ny = exposure.getWidth()/binSize, exposure.getHeight()/binSize
+        if nx*ny <= 1:
+            bg = afwMath.makeStatistics(exposure.getMaskedImage(), afwMath.MEDIAN).getValue()
+            bkgd = None
```
jdswinbank: Could we find better names for these (``medianBg`` and ``modelBg``, maybe)?

python/lsst/pipe/tasks/repair.py
```diff
@@ -234,22 +234,40 @@ def cosmicRay(self, exposure, keepCRs=None):
...
-            crs = measAlg.findCosmicRays(mi, psf, bg, pexConfig.makePolicy(self.config.cosmicray), keepCRs)
+            crs = measAlg.findCosmicRays(exposure.getMaskedImage(),
+                                         psf, bg, pexConfig.makePolicy(self.config.cosmicray), keepCRs)
+            if bkgd:
+                # Add back background image
+                img = exposure.getMaskedImage()
+                img += bkgd.getImageF()
+                del img
+                # Replace original image with CR subtracted image
+                exposure0.getMaskedImage().assign(exposure.getMaskedImage())
```
jdswinbank: Why not ``exposure0.setMaskedImage(exposure.getMaskedImage())``? That naively seems more obvious. Is there a performance or memory management consideration? I think you end up copying the pixels in both cases, right?

python/lsst/pipe/tasks/repair.py
laurenam: The way I see it, the current logic is much simpler for the `nx*ny<=1` case.  This is because the `estimateBackground` function returns an exposure and a `lsst.afw.math.mathLib.Background` object (computed based on the config values, for which the default is indeed set to MEDIAN here).  The variable `bg` fed into `findCosmicRays` is a double, so I would still need to compute that with `afwMath.makeStatistics(bg.getImageF(), afwMath.MEDIAN).getValue()` after having run `bg, exposure = measAlg.estimateBackground(exposure, self.config.cosmicray.background, subtract=False)` and note that I have set `subtract=False` in the above as we do not actually want a background subtracted copy of the image here (although I suppose we could do that as in the `nx*ny>1` case, but that seems like an extra operation that is not required for the majority of cases).

I can confirm that the MEDIAN sky level is the same for the case I gave above as it is in the current implementation.  

In all, I am inclined to leave the current logic.

python/lsst/pipe/tasks/repair.py
laurenam: Fixed.

python/lsst/pipe/tasks/repair.py
laurenam: Good idea.

python/lsst/pipe/tasks/repair.py
laurenam: I was just following the changes made in DM-4102.  Looking at the difference between the two options, `assign` allows for bounding box and image origin inputs.  As these are not required here, I will make the change as you suggest.

PR #30, opened 2016-02-29
Tickets/dm 4866

rearmstr:


tests/testCoadds.py
```diff
@@ -130,6 +132,32 @@ def tearDown(self):
...
 
+    def getCalexpIds(self, butler):
+        return butler.mapper.index['calexp']['visit'][None]
+
+    def addMaskPlanes(self):
+        butler = lsst.daf.persistence.Butler(DATAREPO_ROOT)
+        # Get the dataId for each calexp in the repository
+        calexpDataIds = self.getCalexpIds(butler)
+        # Loop over each of the calexp and add the CROSSTALK and NOT_DEBLENDED mask planes
+        for ID in calexpDataIds:
```
rearmstr: Is there a good reason to capitalize "ID" here and below?

tests/testCoadds.py
```diff
@@ -175,6 +203,14 @@ def comparePsfs(self, a, b):
             for aParam, bParam in zip(aFuncParams, bFuncParams):
                 self.assertEqual(aParam, bParam)
 
+    @unittest.skip("Remove test until DM-5174 is complete")
+    def testMasksRemoved(self):
```
rearmstr: I know this is waiting on another ticket, but doesn't this need more code to actually work?  Does it make sense to include it in this ticket if the later ticket will need to change it?

tests/testCoadds.py
natelust: No good reason except you commonly see Id capitalized in general text. I can change it though

tests/testCoadds.py
natelust: Well the code that will need to change is one line that this ticket added and is documented on the ticket. I would rather not loose the work of writing this test, and need to reduplicate it later. Since this is currently not doing anything it should be save to leave in. If in the future the other ticket is marked invalid, it should be clear enough to remove this bit of code too.

tests/testCoadds.py
```diff
@@ -130,6 +132,32 @@ def tearDown(self):
...
 
+    def getCalexpIds(self, butler):
+        return butler.mapper.index['calexp']['visit'][None]
+
+    def addMaskPlanes(self):
+        butler = lsst.daf.persistence.Butler(DATAREPO_ROOT)
+        # Get the dataId for each calexp in the repository
+        calexpDataIds = self.getCalexpIds(butler)
+        # Loop over each of the calexp and add the CROSSTALK and NOT_DEBLENDED mask planes
+        for Id in calexpDataIds:
```
PaulPrice: `Id` looks like a class.
Don't use `id` (which would clobber the builtin function).
Give the iteration over `calexpDataIds`, using `dataId` would be ideal.

jdswinbank:
Has been merged.

PR #31, opened 2016-03-03
DM-4692: Refactor ProcessCcdTask and sub-tasks

PaulPrice:


PaulPrice:
Third commit ("Refactor ProcessCcdTask and sub-tasks") is far too large (1600+ lines added).  Each of the bullet items should be a separate commit as far as possible.

PaulPrice:
All commit messages should be in the imperative (e.g,. not "removed", but "remove").

examples/calibrateTask.py
```diff
@@ -27,18 +27,24 @@
...
+from lsst.afw.geom import PointD
+from lsst.meas.algorithms import LoadReferenceObjectsTask
+from lsst.meas.astrom import createMatchMetadata
+
+from lsst.pipe.tasks.characterizeImage import CharacterizeImageTask
 from lsst.pipe.tasks.calibrate import CalibrateTask
 
 np.random.seed(1)
 
+FilterName = "r"
```
PaulPrice: Shouldn't have a leading capital, which is reserved for class names.

examples/calibrateTask.py
```diff
@@ -27,18 +27,24 @@
...
-from lsst.meas.astrom import AstrometryTask
+import lsst.pipe.base as pipeBase
+from lsst.daf.butlerUtils import ExposureIdInfo
+import lsst.afw.display as afwDisplay
+import lsst.afw.table as afwTable
+import lsst.afw.image as afwImage
+from lsst.pex.config import Config
+from lsst.afw.coord import Coord
+from lsst.afw.geom import PointD
+from lsst.meas.algorithms import LoadReferenceObjectsTask
```
PaulPrice: Changing these imports has nothing to do with the goal of this ticket or the declared contents of the commits. Making such changes just generates churn (and makes my job as a reviewer harder).  I would like to see this undone.  If it absolutely must be done, then it should be done in a separate commit.

examples/calibrateTask.py
```diff
@@ -52,81 +58,97 @@ def loadData(pixelScale=1.0):
...
-    """An override for CalibrateTask's astrometry task"""
-    def __init__(self, *args, **kwargs):
-        super(MyAstrometryTask, self).__init__(*args, **kwargs)
+class MyAstrometryTask(pipeBase.Task):
+    """An override for CalibrateTask's astrometry task that fakes a solution"""
+    ConfigClass = Config
+    _defaultName = "astrometry"
+
+    def __init__(self, schema=None, **kwargs):
+        pipeBase.Task(**kwargs)
```
PaulPrice: Pointless.

PaulPrice: Oh, maybe not.  Document why it's here (to accept `schema`).

examples/calibrateTask.py
```diff
@@ -52,81 +58,97 @@ def loadData(pixelScale=1.0):
...
-        # Perform the "match"
-        #
-        matches = []
-        md = dafBase.PropertyList()
-        for m, s in zip(refCat, sourceCat):
-            matches.append(afwTable.ReferenceMatch(m, s, 0.0))
+    def loadAndMatch(self, exposure, sourceCat):
+        """!Fake loading and matching
+
+        Copy the source catalog to a reference catalog and producing a match list
```
PaulPrice: produce

examples/calibrateTask.py
```diff
@@ -52,81 +58,97 @@ def loadData(pixelScale=1.0):
...
+        """
+        wcs = exposure.getWcs()
+        refSchema = LoadReferenceObjectsTask.makeMinimalSchema(
+            filterNameList = [FilterName],
```
PaulPrice: @timj says the space around the equals is disallowed by PEP-8.  I thought we allow it for this style (kwargs on separate lines).  What's the answer?

timj: Well, http://developer.lsst.io/en/latest/coding/python_style_guide.html#spaces-must-not-be-used-around-for-default-parameter indicates that in the function definition there should not be spaces so I'm not sure why we would want them when we call the function. I'm having trouble finding the bit in the style guide where we talk about kwargs on separate lines. We are in a bit of a limbo where life would be much easier if we defined our style guide relative to PEP8.

examples/calibrateTask.py
```diff
@@ -52,81 +58,97 @@ def loadData(pixelScale=1.0):
...
-    #
-    exposure = loadData(config.initialPsf.pixelScale)
-    result = calibrateTask.run(exposure)
 
-    exposure0, exposure = exposure, result.exposure
-    sources = result.sources
+    # load the data
+    # Exposure ID and the number of bits required for exposure IDs are usually obtained from a data repo,
+    # but here we pick reasonable values (there are 64 bits to share between exposure IDs and source IDs).
+    exposure = loadData()
```
PaulPrice: Why are you changing the `pixelScale` from 0.185 to the default 1.0?

examples/calibrateTask.py
```diff
@@ -52,81 +58,97 @@ def loadData(pixelScale=1.0):
...
-        with ds9.Buffering():
-            for s in sources:
-                xy = s.getCentroid()
-                ds9.dot('+', *xy, ctype=ds9.CYAN if s.get("flags_negative") else ds9.GREEN, frame=frame)
+    # calibrate the exposure
+    calRes = calibrateTask.calibrate(exposure=charRes.exposure, exposureIdInfo=exposureIdInfo)
+    if display:
+        displayFunc(calRes.exposure, calRes.sourceCat, frame=2)
+
+def displayFunc(exposure, sourceCat, frame):
```
PaulPrice: Why did you need to pull this out into its own function?  Please don't make unnecessary changes.

PaulPrice: Oh, I see --- you're calling it twice now.

Please add a quick docstring.

examples/detectAndMeasureExample.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python
+
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
```
PaulPrice: It's 2016.

examples/detectAndMeasureExample.py
```diff
@@ -0,0 +1,104 @@
...
+from lsst.afw.detection import GaussianPsf    
+import lsst.afw.image as afwImage
+from lsst.meas.astrom import displayAstrometry
+from lsst.meas.algorithms import estimateBackground
+from lsst.pipe.tasks.calibrate import DetectAndMeasureTask
+from lsst.pipe.tasks.repair import RepairTask
+
+np.random.seed(1)
+
+FilterName = "r"
```
PaulPrice: Leading capital letter is reserved for class names.

examples/detectAndMeasureExample.py
```diff
@@ -0,0 +1,104 @@
...
+    # Load sample input from disk
+    mypath = lsst.utils.getPackageDir('afwdata')
+    imFile = os.path.join(mypath, "CFHT", "D4", "cal-53535-i-797722_small_1.fits")
+
+    exposure = afwImage.ExposureF(imFile)
+    # add a filter
+    afwImage.Filter.define(afwImage.FilterProperty(FilterName, 600, True))
+    exposure.setFilter(afwImage.Filter(FilterName))
+    # add a simple Gaussian PSF model
+    psfModel = GaussianPsf(11, 11, psfSigma)
```
PaulPrice: Why hard-code the size and not the sigma, when the two should be related.

examples/detectAndMeasureExample.py
```diff
@@ -0,0 +1,104 @@
...
+    estBg, exposure = estimateBackground(
+        exposure = exposure,
+        backgroundConfig = backgroundConfig,
+        subtract = True,
+    )
+
+    # detect and measure
+    damRes = detectAndMeasureTask.run(exposure=exposure, exposureIdInfo=exposureIdInfo)
+    if display:
+        displayAstrometry(frame=2, exposure=damRes.exposure, sourceCat=damRes.sourceCat, pause=False)
```
PaulPrice: Why frame 2?

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -21,151 +21,80 @@
...
+from lsstDebug import getDebugFrame
 import lsst.pex.config as pexConfig
-import lsst.afw.math as afwMath
-import lsst.afw.table as afwTable
-import lsst.meas.algorithms as measAlg
 import lsst.pipe.base as pipeBase
-from lsstDebug import getDebugFrame
-from lsst.afw.display import getDisplay
-from lsst.meas.base import BasePlugin, SingleFrameMeasurementTask, MeasureApCorrTask
+import lsst.afw.table as afwTable
```
PaulPrice: This import of lsst.afw.table has been moved.  Please leave it where it was unless there's a good reason (and then do it on a separate commit).

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -21,151 +21,80 @@
...
+        dtype = str,
+        default = ("calib_psfCandidate", "calib_psfUsed", "calib_psfReserved"),
+        doc = "Fields to copy from the icSource catalog to the output catalog for matching sources. "
+        "Missing fields are logged (at debug level) and ignored. "
+        "If detectAndMeasure.doMeasureApCorr is True and detectAndMeasure cannot determine its own "
+        "suitable candidates, then this list must include "
+        "config.detectAndMeasure.measureApCorr.inputFilterFlag."
     )
-
-    def validate(self):
```
PaulPrice: Why has the validation been removed?

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -21,151 +21,80 @@
...
-        self.detection.returnOriginalFootprints = True
-        self.initialMeasurement.algorithms.names -= ["base_ClassificationExtendedness"]
-        self.initialMeasurement.doApplyApCorr = "no" # no aperture correction data yet
-        self.measurement.doApplyApCorr = "yes"
-        initflags = [x for x in self.measurePsf.starSelector["catalog"].badStarPixelFlags]
-        self.measurePsf.starSelector["catalog"].badStarPixelFlags.extend(initflags)
-        self.background.binSize = 1024
+        pexConfig.Config.setDefaults(self)
+        # aperture correction should already be measured
+        self.detectAndMeasure.doMeasureApCorr = False
```
PaulPrice: Why is measuring the aperture correction part of `detectAndMeasure`?  That it has to be disabled here may indicate that the model isn't quite right.  Maybe it shouldn't be part of `detectAndMeasure`, or maybe it should be an API parameter rather than a config parameter.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -174,468 +103,432 @@ def setDefaults(self):
...
+        <dt>MAGZERO <dd>Measured zeropoint (DN per second)
+    </dl>
+
+    Exposure metadata
+    <dl>
+        <dt>MAGZERO_RMS  <dd>MAGZERO's RMS == return.sigma
+        <dt>MAGZERO_NOBJ <dd>Number of stars used == return.ngood
+        <dt>COLORTERM1   <dd>?? (always 0.0)
+        <dt>COLORTERM2   <dd>?? (always 0.0)
+        <dt>COLORTERM3   <dd>?? (always 0.0)
```
PaulPrice: What do you mean "??"?

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
```
PaulPrice: It's 2016.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        dtype = bool,
+        default = True,
+        doc = "Write icExp and icExpBackground in addition to icSrc? Ignored if doWrite False.",
+    )
+    psfIterations = pexConfig.RangeField(
+        dtype = int,
+        default = 2,
+        min = 1,
+        doc = "Number of iterations of detect sources, measure sources, estimate PSF. "
+            "If useSimplePsf='all_iter' then 2 should be plenty; otherwise more may be wanted.",
```
PaulPrice: `useSimplePsf` is a straight boolean, and its docs doesn't mention this possibility.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+    detectAndMeasure = pexConfig.ConfigurableField(
+        target = DetectAndMeasureTask,
+        doc = "Detect and measure sources, for measuring PSF",
+    )
+    useSimplePsf = pexConfig.Field(
+        dtype = bool,
+        default = True,
+        doc = "Replace the existing PSF model with a simplified version that has the same sigma "
+            "at the start of each PSF determination iteration? Doing so makes PSF determination "
+            "converge more robustly and quickly.",
```
PaulPrice: Why is this so?  What's your criteria for convergence?
A 'complicated' PSF doesn't really enter into the iteration much --- detection is done with a Gaussian regardless of the PSF, and the only other thing I can think of where the PSF might come into it is measurement or aperture correction, but I don't see how that would change the PSF.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+    )
+    installSimplePsf = pexConfig.ConfigurableField(
+        target = InstallGaussianPsfTask,
+        doc = "Install a simple PSF model",
+    )
+    astrometry = pexConfig.ConfigurableField(
+        target = AstrometryTask,
+        doc = "Task to load and match reference objects. Only used if measurePsf can use matches. "
+            "Warning: matching will only work well if the initial WCS is accurate enough "
+            "to give good matches (roughly: good to 3 arcsec across the CCD).",
```
PaulPrice: This is far too specific.  Surely that warning is appropriate for any and all astrometric matching?
And how can you possibly say it needs to be good to 3 arcsec?  Surely it depends on the matching radius in the matcher?

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+    )
+    repair = pexConfig.ConfigurableField(
+        target = RepairTask,
+        doc = "Remove cosmic rays",
+    )
+
+    def setDefaults(self):
+        pexConfig.Config.setDefaults(self)
+        # just detect bright stars; includeThresholdMultipler=10 seems large,
+        # but these are the values we have been using
```
PaulPrice: No need for the commentary on the values.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        self.detectAndMeasure.detection.thresholdValue = 5.0
+        self.detectAndMeasure.detection.includeThresholdMultiplier = 10.0
+        # do not deblend, as it makes a mess
+        self.detectAndMeasure.doDeblend = False
+        # measure and apply aperture correction; note: measuring and applying aperture
+        # correction are disabled until the final measurement, after PSF is measured
+        self.detectAndMeasure.doMeasureApCorr = True
+        self.detectAndMeasure.measurement.doApplyApCorr = "yes"
```
PaulPrice: This seems to be a common refrain.  Maybe `doApplyApCorr` should default to "yes".

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        # minimal set of measurements needed to determine PSF
+        self.detectAndMeasure.measurement.plugins.names = [
+            "base_PixelFlags",
+            "base_SdssCentroid",
+            "base_SdssShape",
+            "base_GaussianFlux",
+            "base_PsfFlux",
+            "base_CircularApertureFlux",
+        ]
```
PaulPrice: Why limit the set of plugins?  All plugins need to have aperture corrections measured.  If they're not measured here, then they can't be corrected later.  So I think you need to be far more inclusive, at least for the fluxes.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+
+    def validate(self):
+        if self.detectAndMeasure.doMeasureApCorr and not self.measurePsf:
+            raise RuntimeError("Must measure PSF to measure aperture correction, "
+                "because flags determined by PSF measurement are used to identify "
+                "sources used to measure aperture correction")
```
PaulPrice: The flag to use for aperture correction is configurable, so you shouldn't make commentary on that here --- it breaks encapsulation.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+     - @ref pipe_tasks_characterizeImage_Purpose
+     - @ref pipe_tasks_characterizeImage_Initialize
+     - @ref pipe_tasks_characterizeImage_IO
+     - @ref pipe_tasks_characterizeImage_Config
+     - @ref pipe_tasks_characterizeImage_Debug
+     - @ref pipe_tasks_characterizeImage_Example
+
+    @section pipe_tasks_characterizeImage_Purpose  Description
+
+    Given an exposure with defects repaired (masked and interpolated over, e.g. as output by IsrTask):
```
PaulPrice: If the defects have been repaired, why is a "final repair" done here?  And why is that not mentioned below?

examples/calibrateTask.py
PaulPrice: But why does it need to accept `schema`?

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+            and the exposure and background arguments must be None;
+            if False the exposure must be provided.
+            True is intended for running as a command-line task, False for running as a subtask
+
+        @return same data as the characterize method
+        """
+        self._frame = self._initialFrame # reset debug display frame
+        self.log.info("Processing %s" % (dataRef.dataId))
+
+        if doUnpersist:
```
PaulPrice: Same comment as for `CalibrateTask`'s `doUnpersist`.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+
+        exposureIdInfo = dataRef.get("expIdInfo")
+
+        charRes = self.characterize(
+            exposure = exposure,
+            exposureIdInfo = exposureIdInfo,
+            background = background,
+        )
+
+        if self.config.doWrite:
```
PaulPrice: Put the following in its own method, to keep this top-level method as slim as possible.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        - Iterate the following config.psfIterations times, or once if config.doMeasurePsf false:
+            - detect and measure sources and estimate PSF (see detectMeasureAndEstimatePsf for details)
+        - interpolate over cosmic rays
+        - perform final measurement
+
+        @param[in,out] exposure  exposure to characterize (an lsst.afw.image.ExposureF or similar).
+            The following changes are made:
+            - update or set psf
+            - set apCorrMap
+            - update detection and cosmic ray mask planes
```
PaulPrice: Give the actual names of the mask planes so the user doesn't have to guess.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+            - subtract background and interpolate over cosmic rays
+        @param[in] exposureIdInfo  ID info for exposure (an lsst.daf.butlerUtils.ExposureIdInfo)
+        @param[in] background  model of background model already subtracted from exposure
```
PaulPrice: "model of background model"

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        - exposure: characterized exposure; image is repaired by interpolating over cosmic rays,
+            mask is updated accordingly, and the PSF model is set
+        - sourceCat: detected sources (an lsst.afw.table.SourceCatalog)
+        - background: model of background subtracted from exposure (an lsst.afw.math.BackgroundList)
+        - psfCellSet: spatial cells of PSF candidates (an lsst.afw.math.SpatialCellSet)
+        """
+        self._frame = self._initialFrame # reset debug display frame
+
+        if not self.config.doMeasurePsf and not exposure.hasPsf():
+            raise pipeBase.TaskError("exposure has no PSF model and config.doMeasurePsf false")
```
PaulPrice: This seems like something that the user would want a traceback for (so they can find where it's coming from), but using `TaskError` suppresses that.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        image = exposure.getMaskedImage().getImage()
+        image -= estBg.getImageF()
+        background.append(estBg)
+
+        psfIterations = self.config.psfIterations if self.config.doMeasurePsf else 1
+        for i in range(psfIterations):
+            if i > 1:
+                # restore original mask so that detections and cosmic rays
+                # are only marked by the final iteration
+                exposure.getMaskedImage().getMask()[:] = originalMask
```
PaulPrice: Why not use `setMask`, which I think removes the pixel copy?

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+
+        # subtract an initial estimate of background level
+        estBg = estimateBackground(
+            exposure = exposure,
+            backgroundConfig = self.config.background,
+            subtract = False, # this makes a deep copy, which we don't want
+        )[0]
+        image = exposure.getMaskedImage().getImage()
+        image -= estBg.getImageF()
+        background.append(estBg)
```
PaulPrice: I would rather leave `subtract=True` and put up with the inefficiency until it's fixed rather than duplicate code.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+                exposure = exposure,
+                exposureIdInfo = exposureIdInfo,
+                background = background,
+            )
+
+            psf = dmeRes.exposure.getPsf()
+            psfSigma = psf.computeShape().getDeterminantRadius()
+            psfDimensions = psf.computeImage().getDimensions()
+            medBackground = np.median(dmeRes.background.getImage().getArray())
+            self.log.info("iter %s; PSF sigma=%0.2f, dimensions=%s; median background=%0.2f" % \
```
PaulPrice: Remove the backslash.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+     - @ref pipe_tasks_characterizeImage_IO
+     - @ref pipe_tasks_characterizeImage_Config
+     - @ref pipe_tasks_characterizeImage_Debug
+     - @ref pipe_tasks_characterizeImage_Example
+
+    @section pipe_tasks_characterizeImage_Purpose  Description
+
+    Given an exposure with defects repaired (masked and interpolated over, e.g. as output by IsrTask):
+    - detect and measure bright sources
+    - repair cosmic rays
```
PaulPrice: This seems to neglect repairing defects.  There's a reason both of those are grouped in the `RepairTask` --- both (should) depend on the PSF (the interpolation doesn't currently because the implementation uses a hard-coded PSF), and so both should be run with the final PSF.

python/lsst/pipe/tasks/characterizeImage.py
PaulPrice: Add units for PSF sigma (pixels or arcsec?).
I don't think the PSF dimensions are of general interest and printing them here is probably left over from your debugging.
Why is the median background printed?  Is it ever significantly different from zero?  I suggest removing it.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        background.append(estBg)
+
+        psfIterations = self.config.psfIterations if self.config.doMeasurePsf else 1
+        for i in range(psfIterations):
+            if i > 1:
+                # restore original mask so that detections and cosmic rays
+                # are only marked by the final iteration
+                exposure.getMaskedImage().getMask()[:] = originalMask
+
+            dmeRes = self.detectMeasureAndEstimatePsf(
```
PaulPrice: dmeRes seems a little heavy.  How about `result`?

python/lsst/pipe/tasks/characterizeImage.py
PaulPrice: Move the logging into the `detectMeasureAndEstimatePsf` method to keep this top-level method as simple as possible.  (Need to pass the iteration number in.)

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+            exposureIdInfo = exposureIdInfo,
+            sourceCat = dmeRes.sourceCat,
+            allowApCorr = True,  # the default; listed for emphasis
+        )
+        self.display("measure", exposure=dmeRes.exposure, sourceCat=dmeRes.sourceCat)
+
+        return dmeRes
+
+    @pipeBase.timeMethod
+    def detectMeasureAndEstimatePsf(self, exposure, exposureIdInfo, background):
```
PaulPrice: Why is this called "EstimatePsf", when in other places we refer to measuring the PSF?
I think naming this method "iterate" or "iterateOnce" might make it a bit easier.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        """!Perform one iteration of detect, measure and estimate PSF
+
+        Performs the following operations:
+        - if config.doMeasurePsf:
+            - install a simple PSF model (replacing the existing one, if need be)
+        - interpolate over cosmic rays with keepCRs=True
```
PaulPrice: When is the final CR removal with `keepCRs=False` done?

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        - estimate background and subtract it from the exposure
```
PaulPrice: I don't see any explicit background estimation here.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        @param[in,out] exposure  exposure to characterize (an lsst.afw.image.ExposureF or similar)
+            The following changes are made:
+            - update or set psf
+            - update detection and cosmic ray mask planes
+            - subtract background
+        @param[in] exposureIdInfo  ID info for exposure (an lsst.daf.butlerUtils.ExposureIdInfo)
+        @param[in,out] background  model of background model already subtracted from exposure
+            (an lsst.afw.math.BackgroundList).
+
+        @return pipe_base Struct containing these fields, all from the final iteration
```
PaulPrice: This method doesn't do any iteration.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        - sourceCat  detected sources (an lsst.afw.table.SourceCatalog)
+        - background  model of background subtracted from exposure (an lsst.afw.math.BackgroundList)
+        - psfCellSet  spatial cells of PSF candidates (an lsst.afw.math.SpatialCellSet)
+        """
+        # install a simple PSF model, if wanted
+        if self.config.doMeasurePsf:
+            if self.config.useSimplePsf or not exposure.hasPsf():
+                self.installSimplePsf.run(exposure=exposure)
+        elif not exposure.hasPsf():
+            raise pipeBase.TaskError("exposure has no PSF model and config.doMeasurePsf false")
```
PaulPrice: Again, don't think `TaskError` is appropriate here because it suppresses the backtrace.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        self.display("repair_iter", exposure=exposure)
+
+        damRes = self.detectAndMeasure.run(
+            exposure = exposure,
+            exposureIdInfo = exposureIdInfo,
+            background = background,
+            allowApCorr = False,  # wait for final PSF
+        )
+
+        measPsfRes = pipeBase.Struct(
```
PaulPrice: Again, would prefer `None` instead of a `Struct` with `None` values.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+            if self.config.useSimplePsf or not exposure.hasPsf():
+                self.installSimplePsf.run(exposure=exposure)
+        elif not exposure.hasPsf():
+            raise pipeBase.TaskError("exposure has no PSF model and config.doMeasurePsf false")
+
+        # run repair, but do not interpolate over cosmic rays (do that elsewhere, with the final PSF model)
+        self.repair.run(exposure=exposure, keepCRs=True)
+        self.display("repair_iter", exposure=exposure)
+
+        damRes = self.detectAndMeasure.run(
```
PaulPrice: `damRes`?  How about `measResults`?

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        )
+        if self.config.doMeasurePsf:
+            if self.measurePsf.usesMatches:
+                matches = self.astrometry.loadAndMatch(
+                    exposure = exposure,
+                    sourceCat = damRes.sourceCat,
+                ).matches
+            else:
+                matches = None
+            measPsfRes = self.measurePsf.run(
```
PaulPrice: `psfResults`

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+                sources = damRes.sourceCat,
+                matches = matches,
+            )
+        self.display("measure_iter", exposure=exposure, sourceCat=damRes.sourceCat)
+
+        return pipeBase.Struct(
+            exposure = exposure,
+            sourceCat = damRes.sourceCat,
+            background = damRes.background,
+            psfCellSet = measPsfRes.cellSet,
```
PaulPrice: Why do we return the `psfCellSet`?  Is it ever going to be used?  If so, what might it be used for?
Are there more useful outputs from `measurePsf` that should be included?

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -0,0 +1,493 @@
...
+        """
+        val = getDebugFrame(self._display, itemName)
+        if not val:
+            return
+
+        displayAstrometry(
+                exposure = exposure,
+                sourceCat = sourceCat,
+                frame = self._frame,
+                pause = False,
```
PaulPrice: I think you do want the `pause`, because of the iteration and the many things being displayed.  Otherwise, the user's going to lose track of what's being displayed.  I think there also needs to be some title on the display to help disambiguate.

python/lsst/pipe/tasks/detectAndMeasure.py
```diff
@@ -0,0 +1,255 @@
...
+        dtype = bool,
+        default = False,
+        doc = "Compute aperture corrections and set ApCorrMap in exposure?",
+    )
+    measureApCorr = pexConfig.ConfigurableField(
+        target = MeasureApCorrTask,
+        doc = "subtask to measure aperture corrections",
+    )
+
+    def setDefaults(self):
```
PaulPrice: Pointless; remove.

python/lsst/pipe/tasks/detectAndMeasure.py
```diff
@@ -0,0 +1,255 @@
...
+    @section pipe_tasks_detectAndMeasure_Initialize  Task initialisation
+
+    @copydoc \_\_init\_\_
+
+    @section pipe_tasks_detectAndMeasure_IO  Invoking the Task
+
+    The main method is `run`.
+
+    Measurement is rather complicated, so that code lives in a method `measure`.
+    It would be wise to replace this by a higher-level measurement task.
```
PaulPrice: Isn't that this one?

python/lsst/pipe/tasks/detectAndMeasure.py
```diff
@@ -0,0 +1,255 @@
...
+        @return pipe_base Struct containing these fields:
+        - exposure: input exposure (as modified in the course of runing)
+        - sourceCat: source catalog (an lsst.afw.table.SourceTable)
+        - background: background model (input as modified, or a new model if input is None);
+            an lsst.afw.math.BackgroundList
+        """
+        if background is None:
+            background = BackgroundList()
+
+        sourceIdFactory = IdFactory.makeSource(exposureIdInfo.expId, exposureIdInfo.unusedBits)
```
PaulPrice: Move these three lines to their own method ("makeTable"?) to keep the main method as thin as possible.

python/lsst/pipe/tasks/detectAndMeasure.py
```diff
@@ -0,0 +1,255 @@
...
+        sourceCat = detRes.sources
+        if detRes.fpSets.background:
+            background.append(detRes.fpSets.background)
+
+        if self.config.doDeblend:
+            self.deblend.run(
+                exposure = exposure,
+                sources = sourceCat,
+                psf = exposure.getPsf(),
+            )
```
PaulPrice: This style makes the code look much longer than it needs to.  Can't you just put the kwargs all on the same line?

python/lsst/pipe/tasks/detectAndMeasure.py
```diff
@@ -0,0 +1,255 @@
...
+
+        @param[in,out] exposure  exposure to process. Background must already be subtracted
+            to reasonable accuracy, else detection will fail.
+            Set apCorrMap if measuring aperture correction.
+        @param[in]     exposureIdInfo  ID info for exposure (an lsst.daf.butlerUtils.ExposureIdInfo)
+        @param[in,out] background  background model to be modified (an lsst.afw.math.BackgroundList),
+            or None to create a new background model
+        @param[in] allowApCorr  allow measuring and applying aperture correction?
+        """
+        if self.config.doMeasureApCorr and allowApCorr:
```
PaulPrice: How about removing some of the unnecessary indentation:

    if not self.config.doMeasureApCorr or not allowApCorr:
        self.measurement.run(...)
        return
    # perform measurements before aperture correction
    self.measurement.run(...)
    # and so on

python/lsst/pipe/tasks/measurePsf.py
```diff
@@ -27,14 +27,13 @@
 import lsst.meas.algorithms.utils as maUtils
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
-import lsst.afw.table as afwTable
 
 class MeasurePsfConfig(pexConfig.Config):
-    starSelector = measAlg.starSelectorRegistry.makeField("Star selection algorithm", default="secondMoment")
+    starSelector = measAlg.starSelectorRegistry.makeField("Star selection algorithm", default="objectSize")
```
PaulPrice: Hooray!

python/lsst/pipe/tasks/processCcd.py
```diff
@@ -19,241 +19,152 @@
...
-        doc = "Perform ISR?",
-    )
     isr = pexConfig.ConfigurableField(
         target = IsrTask,
-        doc = "Instrumental Signature Removal",
+        doc = """Task to perform instrumental signature removal or load a post-ISR image; ISR consists of:
+            - assemble raw amplifier images into an exposure with image, variance and mask planes
+            - perform bias subtraction, flat fielding, etc.
+            - mask known bad pixels
+            - provide a preliminary WCS
```
PaulPrice: Is it worth putting "provide a preliminary PSF" in ISR?

python/lsst/pipe/tasks/processCcd.py
```diff
@@ -19,241 +19,152 @@
...
-        # with the background estimated in the calibrate step
-        for bg in result.backgrounds:
-            backgrounds.append(bg)
-        result.backgrounds = backgrounds
+    def __init__(self, **kwargs):
+        pipeBase.CmdLineTask.__init__(self, **kwargs)
+        self.makeSubtask("isr")
+        self.makeSubtask("charImage")
+        self.makeSubtask("calibrate",
+            icSourceSchema = self.charImage.schema,
```
PaulPrice: Why can't this go on the previous line?  It's such a waste otherwise.

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env python
+from __future__ import absolute_import, division, print_function
+#
+# LSST Data Management System
+# Copyright 2008-2013 LSST Corporation.
```
PaulPrice: 2016

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
```
PaulPrice: Modern version of this notice uses `https://...`

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+from lsst.meas.algorithms.testUtils import plantSources
+from lsst.pipe.tasks.detectAndMeasure import DetectAndMeasureTask
+
+try:
+    obsTestDir = getPackageDir("obs_test")
+    InputDir = os.path.join(obsTestDir, "data", "input")
+except Exception:
+    InputDir = None
+
+class ProcessCcdTestCase(lsst.utils.tests.TestCase):
```
PaulPrice: Why is this called `ProcessCcdTestCase` when it's supposed to be testing `DetectAndMeasureTask`?

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+    def testBasics(self):
+        """Test detection and measurement on simple synthesized data
+        """
+        bbox = Box2I(Point2I(256, 100), Extent2I(128, 127))
```
PaulPrice: Box for what?

PaulPrice: Please use a much larger XY0 (e.g., 10000 or more) so any problems have a more difficult time hiding.

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+        numY = 5
+        coordList = self.makeCoordList(
+            bbox = bbox,
+            numX = numX,
+            numY = numY,
+            minCounts = minCounts,
+            maxCounts = maxCounts,
+            sigma = starSigma,
+        )
+        kwid = 11 # kernel width
```
PaulPrice: `kwid` is an awful variable name.  I was thinking it was a keyword identifier or something.  Please spell out "width".

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+        config = DetectAndMeasureTask.ConfigClass()
+        task = DetectAndMeasureTask(config=config, schema=schema)
+
+        butler = Butler(root=InputDir)
+        dataRef = butler.dataRef("calexp", dataId=dict(visit=1))
+        wcs = dataRef.get("raw").getWcs()
+        exposure.setWcs(wcs)
+        exposureIdInfo = dataRef.get("expIdInfo")
+        taskRes = task.run(exposure=exposure, exposureIdInfo=exposureIdInfo)
+        self.assertEqual(len(taskRes.sourceCat), numX * numY)
```
PaulPrice: No spaces either side of `*`.

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+        schema = taskRes.sourceCat.schema
+        centroidFlagKey = schema.find("slot_Centroid_flag").getKey()
+        parentKey = schema.find("parent").getKey()
+        psfFluxFlagKey = schema.find("slot_PsfFlux_flag").getKey()
+        psfFluxKey = schema.find("slot_PsfFlux_flux").getKey()
+        for src in taskRes.sourceCat:
+            self.assertFalse(src.get(centroidFlagKey))  # centroid found
+            self.assertEqual(src.get(parentKey), 0)     # not debelended
```
PaulPrice: Is this a statement about the nature of the image, because deblending should have been performed (it defaults to `True`, and I don't see it disabled)?

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+            self.assertFalse(src.get(psfFluxFlagKey))   # flux measured
+            self.assertGreater(src.get(psfFluxKey), 4000)  # flux sane
+
+    def makeCoordList(self, bbox, numX, numY, minCounts, maxCounts, sigma):
+        """Make a coordList for plantSources
+
+        Coords are uniformly spaced in a rectangular grid, with linearly increasing counts
+        """
+        dX = bbox.getWidth() / float(numX)
```
PaulPrice: No spaces either side of `/` or `*` (below).

PaulPrice: Is the cast required, given `from future import division`?

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+        dY = bbox.getHeight() / float(numY)
+        minX = bbox.getMinX() + (dX / 2.0)
```
PaulPrice: Remove parens --- they don't add anything for anyone with very basic knowledge of the order of operations (it's not an obscure case).

tests/testDetectAndMeasure.py
```diff
@@ -0,0 +1,134 @@
...
+        minY = bbox.getMinY() + (dY / 2.0)
+        dCounts = (maxCounts - minCounts) / (numX * numY - 1)
+    
+        coordList = []
+        counts = minCounts
+        for i in range(numX):
+            x = minX + (dX * i)
```
PaulPrice: Remove parens, no space either side of `*`.

I have a hard time thinking REO wrote this....

tests/testDetectAndMeasure.py
timj: For reference: http://developer.lsst.io/en/latest/coding/python_style_guide.html#binary-operators-should-be-surrounded-by-a-single-space-except-for

tests/testProcessCcd.py
```diff
@@ -55,7 +55,7 @@ def testProcessCcd(self):
             dataId = dict(visit=1)
             dataIdStrList = ["%s=%s" % (key, val) for key, val in dataId.iteritems()]
             fullResult = ProcessCcdTask.parseAndRun(
-                args = [InputDir, "--output", outPath, "--id"] + dataIdStrList,
+                args = [InputDir, "--output", outPath, "--clobber-config", "--id"] + dataIdStrList,
```
PaulPrice: Why is `--clobber-config` required?  It shouldn't be used unless necessary.

tests/testProcessCcd.py
```diff
@@ -110,26 +110,26 @@ def testProcessCcd(self):
...
                     print("psf Ixx = %r, Iyy = %r, Ixy = %r" % (psfIxx, psfIyy, psfIxy))
 
-                    self.assertAlmostEqual(bgMean, 327.61695151065607, places=7)
-                    self.assertAlmostEqual(bgStdDev, 0.40081096398879795, places=7)
+                    self.assertAlmostEqual(bgMean, 327.73706832701379, places=7)
+                    self.assertAlmostEqual(bgStdDev, 0.478942723145386, places=7)
                     self.assertEqual(len(icSrc), 28)
-                    self.assertEqual(len(sources), 188)
-                    self.assertEqual(numGoodPix, 1953514)
+                    self.assertEqual(len(src), 183)
```
PaulPrice: Why is it good that you're finding fewer sources?

tests/testProcessCcd.py
```diff
@@ -110,26 +110,26 @@ def testProcessCcd(self):
...
+                    self.assertAlmostEqual(imMean,      1.6992922142160001, places=7)
+                    self.assertAlmostEqual(imStdDev,  163.66265133203555, places=7)
                     self.assertAlmostEqual(varMean,   212.5416436210289, places=7)
                     self.assertAlmostEqual(varStdDev, 121.25084470503148, places=7)
-                    self.assertAlmostEqual(psfIxx, 2.790601924907123, places=7)
-                    self.assertAlmostEqual(psfIyy, 2.1989110429762513, places=7)
-                    self.assertAlmostEqual(psfIxy,  0.16599586021207952, places=7)
+                    self.assertAlmostEqual(psfIxx, 2.205197497931509, places=7)
+                    self.assertAlmostEqual(psfIyy, 2.0893651175583723, places=7)
+                    self.assertAlmostEqual(psfIxy,  -0.07647011192069166, places=7)
```
PaulPrice: That's a HUGE change (half a pixel in width, and different orientation)!  You should verify that it's correct.

python/lsst/pipe/tasks/characterizeImage.py
RobertLuptonTheGood: If we know which bits are set (and we should) we can clear them instead of saving a copy of the mask.  More memory and cache efficient; I think you can say `exposure.getMaskedImage().getMask()[:] &= ~bitsToClear`

PR #34, opened 2016-03-08
DM-5348 relative to DM-4692

r-owen:
Improve the support for camera-specific ISR tasks in ProcessCcdTask
by getting the dataset type from the ISR task's config. This allows
support for ISR tasks that read data other than "raw", such as
DecamNullIsrTask and SdssNullIsrTask, eliminating the need for camera-
specific variants of ProcessCcdIsr in these cases.


parejkoj:
Now that I've looked at pipeBase.ArgumentParser, I understand why we derived from argparse. Looks good, and thanks for giving me a chance to learn that lesson.

PR #36, opened 2016-03-09
Review of DM-2754: Documentation for multiband coadd processing.

jdswinbank:


doc/multiBand.dox
```diff
@@ -0,0 +1,73 @@
...
+/**
+\page pipeTasks_multiBand High-level Overview of Multi-Band Coadd Processing
+
+The goal of coadd measurement is to consistently detect and measure sources across coadds constructed in 
+multiple filter bands in a consistent and reasonable manner. Specifically, we would like to measure the same 
+(deblended) sources across all photometric bands. There are several advantages to imposing this requirement. 
+We get the same deblending in all the bands. This allows us to run forced photometry on consistently deblended 
+objects which gives us reasonable colors. It also permits us to obtain measurements of flux in the case of 
+drop-out objects such as Lyman-break galaxies. Our strategy for meeting the consistent deblending requirement 
+can be broken up into a number of simple steps.
```
jdswinbank: This introductory paragraph could usefully be extended a bit by incorporating some of the material from the [HSC equivalent](http://hsca.ipmu.jp/hscsphinx/pipeline/multiband.html). In particular, the HSC paragraphs labelled "forced photometry" and "model magnitude issue" go into a (little!) bit more detail than you've done, and I think you should capture that.

doc/multiBand.dox
```diff
@@ -0,0 +1,73 @@
...
+<DL> 
+  <DT> i. \ref DetectCoaddSourcesTask_ "DetectCoaddSourcesTask": Detect Sources in Individual Coadds</DT>
```
jdswinbank: Some of the tasks mentioned here seem to get linked to their documentation, but not all of them. I'm not sure if that's just an artefact of the way I built the documentation, or if it's because not all of the tasks are actually documented, or something else.

doc/multiBand.dox
```diff
@@ -0,0 +1,73 @@
...
+multiple filter bands in a consistent and reasonable manner. Specifically, we would like to measure the same 
+(deblended) sources across all photometric bands. There are several advantages to imposing this requirement. 
+We get the same deblending in all the bands. This allows us to run forced photometry on consistently deblended 
+objects which gives us reasonable colors. It also permits us to obtain measurements of flux in the case of 
+drop-out objects such as Lyman-break galaxies. Our strategy for meeting the consistent deblending requirement 
+can be broken up into a number of simple steps.
+<DL> 
+  <DT> i. \ref DetectCoaddSourcesTask_ "DetectCoaddSourcesTask": Detect Sources in Individual Coadds</DT> 
+  <DD> This step is the last step in the process of constructing a coadd and only temporarily belongs in the 
+  Multi-Band Coadd processing pipeline. After scaling the variance plane in a coadd to correct for the
```
jdswinbank: I'm not sure what it means to say this "only temporarily belongs". Isn't a source detection step always necessary to start the ball rolling?

Also: I'd avoid using the term "pipeline". LDM-151 (the DM Applications Design) uses that term for some quite specific pieces of software, and this isn't one of them. Maybe prefer "procedure" or similar.

doc/multiBand.dox
```diff
@@ -0,0 +1,73 @@
...
+  each source comes from. This produces a cross-band master source catalog. During the merging process, we 
+  cull garbage peaks around bright sources using the criteria described in \ref CullPeaksConfig 
+  "CullPeaksConfig".
+  \par Inputs:
+    deepCoadd_det{tract,patch,filter}: SourceCatalog (only parent Footprints)
+  \par Outputs:
+    deepCoadd_mergeDet{tract,patch}: SourceCatalog (only parent Footprints)
+  \par Data Unit: 
+    tract, patch</DD>
+  <DT> iii. \ref MeasureMergedCoaddSourcesTask_ "MeasureMergedCoaddSourcesTask": De-Blend & Measure Sources </DT>
```
jdswinbank: (Throughout) we normally write "deblend" rather than "de-blend".

doc/multiBand.dox
```diff
@@ -0,0 +1,73 @@
...
+  <DT> iii. \ref MeasureMergedCoaddSourcesTask_ "MeasureMergedCoaddSourcesTask": De-Blend & Measure Sources </DT> 
+  <DD> The master catalog is used to de-blend & measure sources in each coadd. This ensures that all bands 
+  will have the same sources. \ref SourceDeblendTask_ "SourceDeblendTask" is used to deblend sources. The 
+  deblender has to retain all peaks from the master catalog, deblending peaks that do not correspond to 
+  sources in the given coadd (dropouts) as PSFs. The \ref SingleFrameMeasurementTask_ "SingleFrameMeasurementTask" 
+  is used to measure the properties of the de-blended children. \ref SetPrimaryFlagsTask_ "SetPrimaryFlagsTask" 
+  is run to set 'is-primary' and related flags and \ref PropagateVisitFlagsTask_ "PropagateVisitFlagsTask" is 
+  used to propagte flags from individual visits to the coadd (eg. was the source used for PSF determination 
+  in the original visit etc...). At this point, the \ref ANetAstrometryTask_ "ANetAstrometryTask" can be run 
+  (optional, not run by default) to match the sources to a reference catalog.
```
jdswinbank: Why would one want this match to a reference catalogue?

doc/multiBand.dox
```diff
@@ -0,0 +1,73 @@
...
+  \par Outputs:
+    deepCoadd_ref{tract,patch}: SourceCatalog
+  \par Data Unit: 
+    tract, patch</DD>
+  <DT> v. Run Forced Photometry on Coadds</DT>
+  <DD> Forced photometry is run on the coadd for each band using reference centroids and shapes but using 
+  footprints from the previous de-blending step i.e. the footprints are band specific.</DD>
+  <DT> vi. Run Forced Photometry on Exposures</DT>
+  <DD> Forced photometry is run on the individual calibrated exposures for each band for each visit using the 
+  reference measurements (but no deblending).</DD>
```
jdswinbank: The measurement on exposures holds exactly the same parameters fixed as on coadds, doesn't it? (Do correct me if not). If so, can you use the same wording for the above two entries to make that clear.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -111,12 +116,117 @@ def setDefaults(self):
...
+\section pipe_tasks_multiBand_DetectCoaddSourcesTask_Config       Configuration parameters
+
+See \ref DetectCoaddSourcesConfig
+
+\section pipe_tasks_multiBand_DetectCoaddSourcesTask_Debug		Debug variables
+
+The \link lsst.pipe.base.cmdLineTask.CmdLineTask command line task\endlink interface supports a
+flag \c -d to import \b debug.py from your \c PYTHONPATH; see \ref baseDebug for more about \b debug.py files.
+
+DetectCoaddSourcesTask has no debug variables of its own because it relegates all the work to
```
jdswinbank: "delegates" rather than "relegates".

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -111,12 +116,117 @@ def setDefaults(self):
...
+## \copybrief DetectCoaddSourcesTask
+## \}
 
 class DetectCoaddSourcesTask(CmdLineTask):
-    """Detect sources on a coadd
-
-    This operation is performed separately in each band.  The detections from
-    each band will be merged before performing the measurement stage.
+    """!
+\anchor DetectCoaddSourcesTask
```
jdswinbank: Ah, here's a problem with the links: the name of the anchor here doesn't match the reference above (the latter ends with a ``_``). Looks like this problem recurs in several places in this commit.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -99,6 +99,11 @@ def getShortFilterName(name):
 ##############################################################################################################
 
 class DetectCoaddSourcesConfig(Config):
+    """!
+\anchor DetectCoaddSourcesConfig
```
jdswinbank: (Throughout) is it necessary to have the documentation out-dented to the leftmost column? It doesn't seem to be the norm in other tasks.

doc/multiBand.dox
```diff
@@ -0,0 +1,73 @@
...
+objects which gives us reasonable colors. It also permits us to obtain measurements of flux in the case of 
+drop-out objects such as Lyman-break galaxies. Our strategy for meeting the consistent deblending requirement 
+can be broken up into a number of simple steps.
+<DL> 
+  <DT> i. \ref DetectCoaddSourcesTask_ "DetectCoaddSourcesTask": Detect Sources in Individual Coadds</DT> 
+  <DD> This step is the last step in the process of constructing a coadd and only temporarily belongs in the 
+  Multi-Band Coadd processing pipeline. After scaling the variance plane in a coadd to correct for the 
+  effects of warping, sources are detected using the \ref SourceDetectionTask_ "SourceDetectionTask". This 
+  produces sources with footprints in each band. 
+  \par Inputs:
```
jdswinbank: Is it useful to list all the inputs and outputs etc on this overview page? I suspect what the casual reader is looking for is a description of the basic logic; if they need this level of detail, they can dive into the task documentation. Including it here just means there's another place to have to remember to update when things change!

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -111,12 +116,117 @@ def setDefaults(self):
...
+footprints. Subsequent tasks in the multi-band processing pipeline will merge sources across bands and, 
+eventually, perform forced photometry. Command-line usage of DetectCoaddSourcesTask expects a data reference 
+to the coadd to be processed. A list of the available optional arguments can be obtained by calling 
+detectCoaddSources.py with the --help command line argument:
+\code
+detectCoaddSources.py --help
+\endcode
+
+To demonstrate usage of the DetectCoaddSourcesTask in the larger context of multi-band processing, we will 
+process the HSC-I & -R band coadds generated from HSC engineering test data provided in the ci-hsc package. To
```
jdswinbank: Please make sure you get the package name correct -- ``ci_hsc`` rather than ``ci-hsc`.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -111,12 +116,117 @@ def setDefaults(self):
...
+    deepCoadd{tract,patch,filter}: ExposureF
+  \par Outputs:
+    deepCoadd_det{tract,patch,filter}: SourceCatalog (only parent Footprints)
+    \n deepCoadd_calexp_det{tract,patch,filter}: ExposureF
+    \n deepCoadd_calexp_detBackground{tract,patch,filter}: BackgroundList
+  \par Data Unit:
+    tract, patch, filter
+
+DetectCoaddSourcesTask delegates most of its work to \ref SourceDetectionTask_ "SourceDetectionTask". You can 
+replace this subtask if you wish.
```
jdswinbank: I believe the usual idiom is "retarget", rather than "replace".

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -111,12 +116,117 @@ def setDefaults(self):
...
+We correct the coadd variance by scaling the variance plane in the coadd to match the observed variance - this 
+is not the right thing to do but is simple and works well in practice. We must also detect sources and 
+generate footprints for the parent sources - this is delegated to the existing SourceDetectionTask.
+
+  \par Inputs:
+    deepCoadd{tract,patch,filter}: ExposureF
+  \par Outputs:
+    deepCoadd_det{tract,patch,filter}: SourceCatalog (only parent Footprints)
+    \n deepCoadd_calexp_det{tract,patch,filter}: ExposureF
+    \n deepCoadd_calexp_detBackground{tract,patch,filter}: BackgroundList
```
jdswinbank: What are these outputs? Specifically, a quick squint at the source suggests the ``_calexp_det`` is the input exposure with its variance plane scaled as discussed above. Is that correct? If so, say so.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -111,12 +116,117 @@ def setDefaults(self):
...
+
+We want to detect sources on coadded images in prep for eventually running forced photometry consistently 
+across multiple filter bands. Two things must be achieved to finish coadd processing. Coadding individual 
+visits requires each exposure to be warped. This introduces covariance in the noise properties across pixels. 
+We correct the coadd variance by scaling the variance plane in the coadd to match the observed variance - this 
+is not the right thing to do but is simple and works well in practice. We must also detect sources and 
+generate footprints for the parent sources - this is delegated to the existing SourceDetectionTask.
+
+  \par Inputs:
+    deepCoadd{tract,patch,filter}: ExposureF
```
jdswinbank: Name depends on config, I think -- it's only ``deepCoadd`` if ``DetectCoaddSourcesConfig.coaddName`` is ``deep``. (Also applies to the below)

doc/multiBand.dox
jdswinbank: (Fixed link above to point to public version of the documentation.)

PR #37, opened 2016-03-10
Review of DM-4841: choose highest S/N band as the reference band when priority band S/N is small

jdswinbank:


jdswinbank:
First line of commit message should be no more than 50 characters.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -697,6 +697,19 @@ class MergeMeasurementsConfig(MergeSourcesConfig):
     pseudoFilterList = ListField(dtype=str, default=["sky"],
                                  doc="Names of filters which may have no associated detection\n"
                                      "(N.b. should include MergeDetectionsConfig.skyFilterName)")
+    snName = Field(dtype=str, default="base_PsfFlux",
+                       doc="Name of flux measurement for calculating the S/N when choosing the "
+                           "reference band.")
```
jdswinbank: Can we justify the choices of default parameters, here and below? (Why PSF flux? why is ``minSN`` 10? why is ``minSNDiff`` 3?). Perhaps they're just what seems to work in practice, or perhaps there's more detailed consideration that the user should be aware of.

Since this is a port from HSC, if it's not straightforward to answer that question, we shouldn't block the merge but should consider filing a documentation debt ticket to come back to it later.

Also, please fix the indentation of ``doc`` (by matching it to ``dtype``, you can actually fit the whole ``doc`` argument on a single line).

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -697,6 +697,19 @@ class MergeMeasurementsConfig(MergeSourcesConfig):
...
+    minSN = Field(dtype=float, default=10.,
+                  doc="If the S/N from the priority band is below this value (and the S/N "
+                      "is larger than minSNDiff compared to the priority band), use the band with "
+                      "the largest S/N as the reference band."
+                  )
+    minSNDiff = Field(dtype=float, default=3.,
+                  doc="If the difference in S/N between another band and the priority band is larger "
+                      "than this value (and the S/N in the priority band is less than minSN) "
+                      "use the band with the largest S/N as the reference band"
+                  )
```
jdswinbank: Same comment re ``)``.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -697,6 +697,19 @@ class MergeMeasurementsConfig(MergeSourcesConfig):
...
                                  doc="Names of filters which may have no associated detection\n"
                                      "(N.b. should include MergeDetectionsConfig.skyFilterName)")
+    snName = Field(dtype=str, default="base_PsfFlux",
+                       doc="Name of flux measurement for calculating the S/N when choosing the "
+                           "reference band.")
+    minSN = Field(dtype=float, default=10.,
+                  doc="If the S/N from the priority band is below this value (and the S/N "
+                      "is larger than minSNDiff compared to the priority band), use the band with "
+                      "the largest S/N as the reference band."
+                  )
```
jdswinbank: For consistency with the code above, please move the ``)`` onto the previous line.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -763,33 +780,68 @@ def mergeCatalogs(self, catalogs, patchRef):
         # This first zip iterates over all the catalogs simultaneously, yielding a sequence of one
         # record for each band, in order.
         for n, orderedRecords in enumerate(zip(*orderedCatalogs)):
```
jdswinbank: It's not strictly part of this change, but since ``n`` is unused, can we get rid of the enumeration here?

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -763,33 +780,68 @@ def mergeCatalogs(self, catalogs, patchRef):
...
-            # Now we iterate over those record-band pairs, until we find the one with the right flag set.
+            # Now we iterate over those record-band pairs, keeping track of the priority and largest S/N band
```
jdswinbank: I think a few more comments scattered through this code would be useful: while there's nothing desperately complicated going on, it took me a few moments of head-scratching to grok the logic.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -763,33 +780,68 @@ def mergeCatalogs(self, catalogs, patchRef):
...
-                hasPseudoFilter = False
-                for pseudoFilterKey in self.pseudoFilterKeys:
-                    for inputRecord, flagKeys in zip(orderedRecords, orderedKeys):
+                parent = (inputRecord.getParent() == 0 and inputRecord.get(flagKeys.footprint))
+                child = (inputRecord.getParent() != 0 and inputRecord.get(flagKeys.peak))
+
+                if not (parent or child):
+                    for pseudoFilterKey in self.pseudoFilterKeys:
                         if inputRecord.get(pseudoFilterKey):
                             self.log.logdebug(msg + "; it is a sky object")
```
jdswinbank: ``msg`` is undefined.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -763,33 +780,68 @@ def mergeCatalogs(self, catalogs, patchRef):
...
-            outputRecord = mergedCatalog.addNew()
-            outputRecord.assign(inputRecord, self.schemaMapper)
-            if flagKeys:
-                outputRecord.set(flagKeys.output, True)
+                if inputRecord.get(self.fluxFlagKey) or inputRecord.get(self.fluxErrKey) == 0:
+                    sn = 0.
+                else:
+                    sn = inputRecord.get(self.fluxKey)/inputRecord.get(self.fluxErrKey)
+                if numpy.isnan(sn) or sn < 0.:
+                    sn = 0
```
jdswinbank: Why floating point zero above and integer here?

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -763,33 +780,68 @@ def mergeCatalogs(self, catalogs, patchRef):
...
+                if (parent or child) and priorityRecord is None:
+                    priorityRecord = inputRecord
+                    priorityFlagKeys = flagKeys
+                    prioritySN = sn
+                if sn > maxSN:
```
jdswinbank: Since we initialized ``maxSN`` to be ``-1``, this will be true even when ``sn`` is ``0``, e.g. if the flux measurement has been flagged or is NaN. Is that the intention?

python/lsst/pipe/tasks/multiBand.py
rearmstr: We can justify the choice of PSF flux because it is in practice the most robust.  I justified the minSN number for HSC based on the data, but the minSNDiff is completely unjustified.  I don't know if this information should be recorded somewhere, but it would be useful to look at these numbers again more generally.  I'll create a ticket for them.

Fixed the indentation.

python/lsst/pipe/tasks/multiBand.py
rearmstr: Fixed.

python/lsst/pipe/tasks/multiBand.py
rearmstr: Fixed.

python/lsst/pipe/tasks/multiBand.py
rearmstr: Fixed.

python/lsst/pipe/tasks/multiBand.py
rearmstr: Removed.

python/lsst/pipe/tasks/multiBand.py
rearmstr: Fixed.

python/lsst/pipe/tasks/multiBand.py
rearmstr: Hmm, I guess this could be a problem if all S/N measurements were either < 0 or NaN.  Then it would not have the expected behavior.  I think it is better to initialize `maxSN` to 0.

PR #42, opened 2016-03-24
DM-1579: HTM indexed files.

SimonKrughoff:


tests/testHtmIndex.py
```diff
@@ -0,0 +1,213 @@
...
+        default_config.is_photometric_name = 'is_phot'
+        default_config.is_resolved_name = 'is_res'
+        default_config.is_variable_name = 'is_var'
+        default_config.id_name = 'id'
+        default_config.extra_col_names = ['val1', 'val2', 'val3']
+        default_config.header_lines = 1
+        default_config.colnames = ['id', 'ra', 'dec', 'a', 'a_err', 'b', 'b_err', 'is_phot', 'is_res',
+                                   'is_var', 'val1', 'val2', 'val3']
+        default_config.delimiter = '|'
+        #this also tests changing the delimiter
```
timj: Need a space between the `#` and the text.

tests/testHtmIndex.py
```diff
@@ -0,0 +1,213 @@
...
+        retVal = IngestIndexedReferenceTask.parseAndRun(args=[input_dir, "--output", self.out_path+"/output_override",
```
timj: This test is testing that things don't crash? Is there any way to test that the right thing happened?

SimonKrughoff: I do test that the catalogs produced by the loader are identical to the same catalogs chosen by selecting a circular aperture from the input data.

This is certainly a high level test.  I could be testing more of the low level methods, but it gives some confidence that it returns identical results to the hand chosen catalogs at a variety of sky positions.

tests/testHtmIndex.py
SimonKrughoff: Thanks.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -133,6 +133,135 @@ def run(self, parsedCmd):
...
+        self.indexer = HtmIndexer(self.config.level)
+
+    def create_indexed_catalog(self, files):
+        rec_num = 0
+        first = True
+        for filename in files:
+            names = True
+            if self.config.colnames:
+                names = self.config.colnames
+            arr = numpy.genfromtxt(filename, dtype=None, skip_header=self.config.header_lines,
```
PaulPrice: Why text input?  Our catalogs are typically large, and in FITS.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -133,6 +133,135 @@ def run(self, parsedCmd):
...
+                # dealing with a string like thing.  Need to get type and size.
+                at_type = afwTable.aliases[str]
+                at_size = dtype[name].itemsize
+                return schema.addField(name, type=at_type, size=at_size)
+            else:
+                at_type = afwTable.aliases[dtype[name].type]
+                return schema.addField(name, at_type)
+
+        for item in mag_column_list:
+            key = schema.addField(item+'_flux', float)
```
PaulPrice: Space around operator `+`.

Here and below.

PaulPrice: `key` is unused.  Why are you dropping it?  You need it if this is not going to be dreadfully slow.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -133,6 +133,135 @@ def run(self, parsedCmd):
...
+    def _set_flags(self, record, row):
+        names = record.schema.getNames()
+        for flag in self._flags:
+            if flag in names:
+                attr_name = 'is_{}_name'.format(flag)
+                record.set(flag, bool(row[getattr(self.config, attr_name)]))
+
+    def _set_mags(self, record, row):
+        for item in self.config.mag_column_list:
+            record.set(item+'_flux', fluxFromABMag(row[item]))
```
PaulPrice: Space around operator `+`.

Here and below.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -133,6 +133,135 @@ def run(self, parsedCmd):
...
+                                dataId=dataId)
+                first = False
+            pixel_ids = set(index_list)
+            shards = []
+            for pixel_id in pixel_ids:
+                dataId = self.make_data_id(pixel_id)
+                catalog = self.get_catalog(dataId, schema)
+                els = numpy.where(index_list == pixel_id)
+                #Just in case someone has only one line in the file.
+                for row in numpy.atleast_1d(arr)[els]:
```
PaulPrice: Going row by row might be slow, especially since you're using field names and not keys.  Can you optimise this?  Some of our tables are VERY large, and this design doesn't permit parallelism so you need to be as fast as possible.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -133,6 +133,135 @@ def run(self, parsedCmd):
...
+                    record = catalog.addNew()
+                    rec_num = self._fill_record(record, row, rec_num)
+                self.butler.put(catalog, self.config.ref_dataset_name, dataId=dataId)
+
+    @staticmethod
+    def make_data_id(pixel_id):
+        return {'pixel_id':pixel_id}
```
PaulPrice: Space after `:`.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -133,6 +133,135 @@ def run(self, parsedCmd):
...
+
+    @staticmethod
+    def compute_coord(row, ra_name, dec_name):
+        return afwCoord.IcrsCoord(row[ra_name]*afwGeom.degrees,
+                                  row[dec_name]*afwGeom.degrees)
+
+    def _set_flags(self, record, row):
+        names = record.schema.getNames()
+        for flag in self._flags:
```
PaulPrice: Where is `self._flags` set?  I don't see it anywhere.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -133,6 +133,135 @@ def run(self, parsedCmd):
...
+class IngestIndexedReferenceTask(pipeBase.CmdLineTask):
+    ConfigClass = IngestIndexedReferenceConfig
+    RunnerClass = IngestReferenceRunner
+    _DefaultName = 'IngestIndexedReferenceTask'
+
+    _flags = ['photometric', 'resolved', 'variable']
+
+    @classmethod
+    def _makeArgumentParser(cls):
+        parser = pipeBase.ArgumentParser(name=cls._DefaultName)
```
PaulPrice: I think you want `InputOnlyArgumentParser`, even though you're writing outputs, so the outputs don't end up in a single output.

python/lsst/pipe/tasks/htmIndexEngine.py
PaulPrice: Oh, it's a class variable.  Sorry.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -0,0 +1,352 @@
...
+
+class IngestIndexedReferenceConfig(pexConfig.Config):
+    ref_dataset_name = pexConfig.Field(
+            dtype = str,
+            default = 'cal_ref_cat',
+            doc = 'String to pass to the butler to retrieve persisted files.',
+    )
+    ra_name = pexConfig.Field(
+            dtype = str,
+            default = '',
```
PaulPrice: Why an empty string `default`, which is useless?  Leaving the `default` out forces the user to set this.  Alternatively, "ra" is probably what most people would choose.  But an empty string is the worst choice.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -0,0 +1,352 @@
...
+    )
+    is_variable_name = pexConfig.Field(
+            dtype = str,
+            default = '',
+            doc = 'Name of column stating if the object is measured to be variable (optional).'
+    )
+    id_name = pexConfig.Field(
+            dtype = str,
+            default = '',
+            doc = 'Name of column to use as an identifier (optional).'
```
PaulPrice: If it's optional, don't set a `default`, and use `optional=True`.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -0,0 +1,352 @@
...
+    level = pexConfig.Field(
+            dtype = int,
+            default = 8,
+            doc = 'Default HTM level.  Level 8 gives ~0.08 sq deg per trixel.',
+    )
+
+    def validate(self):
+        pexConfig.Config.validate(self)
+        if not (self.ra_name and self.dec_name and self.mag_column_list):
+            raise ValueError("RA column name, Dec column name and at least one magnitude column must be"+\
```
PaulPrice: For a more helpful error message, use the field names in the `Config`, and not what they mean, e.g., `ra_name` instead of `RA column name`.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -0,0 +1,352 @@
...
+                             " supplied.")
+        if len(self.mag_err_column_map) > 0 and not len(self.mag_column_list) == len(self.mag_err_column_map):
+            raise ValueError("If magnitude errors are provided, all magnitudes must have an error column")
+
+class IngestReferenceRunner(pipeBase.TaskRunner):
+    """!Task runner for the reference catalog ingester
+    """
+    def run(self, parsedCmd):
```
PaulPrice: This implementation removes all support for `multiprocessing`, which is good because the design doesn't permit parallelism, but you need to tell the user that (and catch any attempt to set `-j`).
This also removes the traceback squashing, which may be less desirable.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -0,0 +1,352 @@
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
```
PaulPrice: It's 2016.  Ditto for other files.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -0,0 +1,352 @@
...
+
+class HtmIndexer(object):
+    def __init__(self, depth=8):
+        """!Construct the indexer object
+
+        @param[in] depth  depth of the hierarchy to construct
+        """
+        self.htm = esutil.htm.HTM(depth)
+        # HACK need to call intersect first otherwise it segfaults
+        _ = self.htm.intersect(1., 2., 0.00001)
```
PaulPrice: No need to hold the return value.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+class HtmIndexer(object):
+    def __init__(self, depth=8):
+        """!Construct the indexer object
+
+        @param[in] depth  depth of the hierarchy to construct
+        """
+        self.htm = esutil.htm.HTM(depth)
+        # HACK need to call intersect first otherwise it segfaults
```
PaulPrice: Has this been reported upstream?

PaulPrice:
I think several of these commits could be squashed together without losing anything.

PaulPrice:
Despite the filename choices, I think there's a lot in here that's not HTM-specific.  I think it's worth making it more generic, so I can choose which indexing scheme I want to use.  For example, you could make an abstract base class for the `Indexer` and allow the user to select (via `Config`) which `Indexer` subclass to use.

tests/testHtmIndex.py
```diff
@@ -0,0 +1,213 @@
...
+        extra_col3 = numpy.array([get_word(num) for num in numpy.random.randint(11, size=size)])
+
+        dtype = numpy.dtype([('id', float), ('ra_icrs', float), ('dec_icrs', float), ('a', float),
+                             ('a_err', float), ('b', float), ('b_err', float), ('is_phot', int),
+                             ('is_res', int), ('is_var', int), ('val1', float), ('val2', float),
+                             ('val3', '|S11')])
+
+        arr = numpy.array(zip(ident, ra, dec, a_mag, a_mag_err, b_mag, b_mag_err, is_photometric, is_resolved,
+                           is_variable, extra_col1, extra_col2, extra_col3), dtype=dtype)
+        numpy.savetxt(out_path+"/ref.txt", arr, delimiter=",",
```
PaulPrice: I think this demonstrates my concern about having the ingestion read CSV rather than FITS.  Converting FITS catalogs covering 3pi steradians to 21 mag to CSV is going to be very slow and consume a lot of disk space.

PaulPrice:
As I've thought about this some more, I've concluded that the ingestion code is going to have to be parallelised in order to ingest large (all sky) catalogs.  That doesn't mean that the parallelisation needs to be done as part of this ticket, but I would suggest that the ingestion should be broken up now into suitable pieces for implementing the parallelisation later.

I see two operations that can be parallelised:
1. Assigning each source an index: can be parallelised over input files, writing temporary files that join sources to indices.
2. Writing each source to the output: can be parallelised over indices, writing the final product.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -36,4 +47,9 @@ def get_pixel_ids(self, ctrCoord, radius):
         return pixel_id_list, is_on_boundary
 
     def index_points(self, ra_list, dec_list):
+        """!Generate trixel ids for each row in an input file
+
+        @param[in] coord_list  List of coord objects to index
```
djreiss: Should list "ra_list", "dec_list" as input parameters in docstring.

python/lsst/pipe/tasks/htmIndexEngine.py
```diff
@@ -206,10 +242,19 @@ def _set_mags(self, record, row):
...
+        @param[in,out] record  SourceCatalog record to modify
+        @param[in] row  dict like object containing the column values
+        """
         for extra_col in self.config.extra_col_names:
             record.set(extra_col, row[extra_col])
 
     def _fill_record(self, record, row, rec_num):
+        """!Fill a record to put in the persisted indexed catalogs
+
+        @param[in,out] record  afwTable.SourceRecord in a reference catalog to fill.
```
djreiss: Add `rec_num` to docstring

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: Text input is the less trivial of the two.  I figured a FITS reader could be added once this is in.  I could do it here if it's a no-go without it.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: O.K.  Good point.  I'll use the keys.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: I don't know how to translate more than one row from a numpy array into a source catalog.  If there's a vectorized way to do this, I'd be happy to know.

BTW, I just changed to using keys.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: Can you expand on this?  I didn't know about InputOnlyArgumentParser.  I'm not reading the inputs from a repository, but I am writing to it.  That seems like the right way to go, but let me know if I'm wrong.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: Good point.  I'm forcing people to set these in the validate step, but I'll take out the default.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: Done.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: Good point.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: True.

python/lsst/pipe/tasks/htmIndexEngine.py
SimonKrughoff: Yes.  I talked with Erin about it, but it is unresolved at the moment.  There is an [issue](https://github.com/esheldon/esutil/issues/12).

tests/testHtmIndex.py
SimonKrughoff: I agree.  I think it would be trivial to add, but I'd rather do that as another ticket so I can close this.  Please push back if that is not acceptable to you.

python/lsst/pipe/tasks/htmIndexEngine.py
PaulPrice: The input mechanism isn't abstracted, so changing the input mechanism isn't straight-forward, and would involve copying a lot of code.

python/lsst/pipe/tasks/htmIndexEngine.py
PaulPrice: Can you add multiple rows at once (be sure to use `reserve` to allocate contiguous memory), and then set the columns for those rows?

python/lsst/pipe/tasks/htmIndexEngine.py
PaulPrice: `InputOnlyArgumentParser` doesn't insist that you provide a `--rerun` or `--output`.  Since one probably wants the reference catalog available regardless of the output, that means putting them in at the top-level, so no `--rerun` or `--output`.

tests/testHtmIndex.py
PaulPrice: I don't think the code is useful if it's not fast, but the reference catalog I usually use is very large (151 GB for PS1 astrometry_net_data).

SimonKrughoff: I get your point, but I disagree it's not useful if it's not fast.  It will be very useful for me when I run simulations and I have to make a new (small) reference catalog for every different run I do.

PR #50, opened 2016-05-16
add DipoleFitTask as default dipole measurement task

SimonKrughoff:
Added config parameter doMeasurementNew (default=True) which if
set to False will fall back on *only* running the previous
DipoleMeasurementTask measurements. If True, it will run
both.

### DIFF
--- a/python/lsst/pipe/tasks/imageDifference.py
+++ b/python/lsst/pipe/tasks/imageDifference.py
@@ -33,10 +33,10 @@ import lsst.meas.astrom as measAstrom
 from lsst.pipe.tasks.registerImage import RegisterTask
 from lsst.meas.algorithms import SourceDetectionTask, PsfAttributes, SingleGaussianPsf, \
     ObjectSizeStarSelectorTask
-from lsst.ip.diffim import ImagePsfMatchTask, DipoleMeasurementTask, DipoleAnalysis, \
+from lsst.ip.diffim import ImagePsfMatchTask, DipoleAnalysis, \
     SourceFlagChecker, KernelCandidateF, cast_KernelCandidateF, makeKernelBasisList, \
     KernelCandidateQa, DiaCatalogSourceSelectorTask, DiaCatalogSourceSelectorConfig, \
-    GetCoaddAsTemplateTask, GetCalexpAsTemplateTask
+    GetCoaddAsTemplateTask, GetCalexpAsTemplateTask, DipoleFitTask
 import lsst.ip.diffim.diffimTools as diffimTools
 import lsst.ip.diffim.utils as diUtils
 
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -36,7 +36,7 @@
 from lsst.ip.diffim import ImagePsfMatchTask, DipoleMeasurementTask, DipoleAnalysis, \
```
SimonKrughoff: I think the DipoleMeasurementTask import is now not used.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -70,6 +70,7 @@ class ImageDifferenceConfig(pexConfig.Config):
     doMatchSources = pexConfig.Field(dtype=bool, default=True,
         doc="Match diaSources with input calexp sources and ref catalog sources")
     doMeasurement = pexConfig.Field(dtype=bool, default=True, doc="Measure diaSources?")
+    doMeasurementNew = pexConfig.Field(dtype=bool, default=True, doc="Measure diaSources using new algorithm?")
```
SimonKrughoff: These names are too similar and will cause confusion.  Maybe rename `doMeasurementNew` to `doDipoleFitting`.  That seems more descriptive.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -518,7 +526,10 @@ def run(self, sensorRef, templateIdList=None):
 
             if self.config.doMeasurement:
                 self.log.info("Running diaSource measurement")
-                self.measurement.run(diaSources, subtractedExposure)
+                if not self.config.doMeasurementNew:
+                    self.measurement.run(diaSources, subtractedExposure)
+                else:
+                    self.measurement.run(diaSources, subtractedExposure, exposure, templateExposure)
```
SimonKrughoff: This suggests that running with different arguments will result in different fields being filled in (or even different schema).  Is it possible to have two different tasks, so we can simply delete one run call when we decide to remove the old way of doing things?

python/lsst/pipe/tasks/imageDifference.py
djreiss: OK, removed.

python/lsst/pipe/tasks/imageDifference.py
djreiss: OK, done. Also changed docstring to: "Fit dipoles using new algorithm?"
Note this option will be removed in the future once DipoleMeasurement is slated for removal.

python/lsst/pipe/tasks/imageDifference.py
djreiss: Sorry, can you explain? What do you mean by two different tasks? The plan is to simply remove lines 529-531 (and the doDipoleFitting option) when we decide to remove the old measurement method. I should note that doDipoleFitting will actually run both tasks by default (due to its config).

SimonKrughoff: Know that I'm not pushing back hard.  Just explaining.

O.K. I guess I was balking at the extra level of indirection.  I.e. the same call does different things (and more things) depending on what arguments are passed.  I would have preferred to see a call to the old measurement and then optionally a call to the new measurement.

Do we actually need to run both algorithms going forward?  The fitter seems like the "right" algorithm.

djreiss: I would like to do some further tests on real data (and/or Ian's simulated images) before officially pulling the plug on the old algorithm. This change is making those tests possible (DM-5412), and we could remove the old one immediately afterwards (new ticket to be filed).

SimonKrughoff: O.K.  Good enough.  I think I was misunderstanding you before.

PR #53, opened 2016-05-28
Tickets/dm 5798

r-owen:


examples/photoCalTask.py
```diff
@@ -30,8 +30,8 @@
 import sys
 import lsst.afw.table              as afwTable
 import lsst.afw.image              as afwImage
-from lsst.meas.astrom import AstrometryTask
 from lsst.pipe.tasks.photoCal import PhotoCalTask
+import lsst.meas.astrom as measAstrom
```
r-owen: No big deal, but why not just import the one additional symbol from meas.astrom?

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -231,7 +237,8 @@ def __init__(self, icSourceSchema=None, **kwargs):
             See also config field `icSourceFieldsToCopy`.
         @param[in,out] kwargs  other keyword arguments for lsst.pipe.base.CmdLineTask
         """
-        pipeBase.Task.__init__(self, **kwargs)
+        butler = kwargs.pop('butler')
```
r-owen: Please use butler in the list of arguments, instead. It's makes the argument list much easier to understand. I do realize you are trying to make it required. You can list it before icSourceSchema with no default, or list it after icSourceSchema with default = None and then raise an exception if butler is None. I would probably do the former, for clarity (despite the fact that it changes the argument order, so you'll have to check all callers and make sure they are providing the schema by name).

 Also please list that argument in the constructor doc string and add a @params icSourceSchema and butler.

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -238,13 +242,15 @@ def DebugInfo(name):
...
     _DefaultName = "imageCharacterization"
+    RunnerClass = pipeBase.ButlerInitializedTaskRunner
 
     def __init__(self, schema=None, **kwargs):
         """!Construct a CharacterizeImageTask
 
         @param[in,out] schema  initial schema (an lsst.afw.table.SourceTable), or None
         @param[in,out] kwargs  other keyword arguments for lsst.pipe.base.CmdLineTask
         """
+        butler = kwargs.pop('butler')
```
r-owen: Same comment as for the calibration task constructor.

python/lsst/pipe/tasks/processCcd.py
```diff
@@ -124,13 +124,15 @@ class ProcessCcdTask(pipeBase.CmdLineTask):
...
     _DefaultName = "processCcd"
 
     def __init__(self, **kwargs):
+        butler = kwargs.pop('butler')
         pipeBase.CmdLineTask.__init__(self, **kwargs)
         self.makeSubtask("isr")
-        self.makeSubtask("charImage")
-        self.makeSubtask("calibrate", icSourceSchema=self.charImage.schema)
+        self.makeSubtask("charImage", butler=butler)
+        self.makeSubtask("calibrate", icSourceSchema=self.charImage.schema, butler=butler)
```
r-owen: If you do decide to list butler before icSourceSchema in the arg list of the calibrate task, I suggest using the same order in calling, even though it doesn't really matter. It just makes the code a bit easier to read.

PR #54, opened 2016-06-01
DM-5877: Use Afterburners to clean up aperture correction logic

PaulPrice:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -25,88 +25,117 @@
...
-            "config.detectAndMeasure.measureApCorr.inputFilterFlag. "
             "Ignored if icSourceCat is not provided."
     )
     matchRadiusPix = pexConfig.Field(
-        dtype = float,
-        default = 3,
-        doc = "Match radius for matching icSourceCat objects to sourceCat objects (pixels)",
+        dtype=float,
+        default=3,
+        doc="Match radius for matching icSourceCat objects to sourceCat objects (pixels)",
```
PaulPrice: Please don't include the above changes removing the spaces around the `=` unless it's in a separate commit.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -226,9 +255,6 @@ def __init__(self, icSourceSchema=None, **kwargs):
         """!Construct a CalibrateTask
 
         @param[in] icSourceSchema  schema for icSource catalog, or None.
-            If measuring aperture correction and the task detectAndMeasure cannot determine
-            its own suitable candidates, then this argument must be specified.
```
PaulPrice: Not sure these lines can be removed without explanation.  Can this value still be `None`?  Under what conditions may it be `None`?

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -255,17 +281,25 @@ def __init__(self, icSourceSchema=None, **kwargs):
                     self.schemaMapper.addMapping(schemaItem.getKey())
 
             if missingFieldNames:
-                raise RuntimeError(
-                    "isSourceCat is missing fields %s specified in icSourceFieldsToCopy" %
-                     (missingFieldNames,))
+                raise RuntimeError("isSourceCat is missing fields {} specified in icSourceFieldsToCopy"
+                                   .format(missingFieldNames))
```
PaulPrice: Cleanup, should not be part of this commit.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -255,17 +281,25 @@ def __init__(self, icSourceSchema=None, **kwargs):
...
+        self.makeSubtask('detection', schema=self.schema)
+
+        self.algMetadata = dafBase.PropertyList()
+
+        if self.config.doDeblend:
+            self.makeSubtask("deblend", schema=self.schema)
+        self.makeSubtask('measurement', schema=self.schema, algMetadata=self.algMetadata)
+        if self.config.doApCorr:
+            self.makeSubtask('applyApCorr', schema=self.schema)
+        self.makeSubtask('afterburners', schema=self.schema)
```
PaulPrice: This has got to be a common pattern (hence the previous use of `detectAndMeasure`).  Can we factor it out somehow?

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -321,20 +352,20 @@ def run(self, dataRef, exposure=None, background=None, icSourceCat=None, doUnper
...
-                background = calRes.background,
-                sourceCat = calRes.sourceCat,
-                astromMatches = calRes.astromMatches,
-                matchMeta = matchMeta,
+                dataRef=dataRef,
+                exposure=calRes.exposure,
+                background=calRes.background,
+                sourceCat=calRes.sourceCat,
+                astromMatches=calRes.astromMatches,
+                matchMeta=matchMeta,
```
PaulPrice: Cleanup, should be in separate commit.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -379,8 +429,8 @@ def calibrate(self, exposure, exposureIdInfo, background=None, icSourceCat=None)
         if self.config.doAstrometry:
             try:
                 astromRes = self.astrometry.run(
-                    exposure = exposure,
-                    sourceCat = sourceCat,
+                    exposure=exposure,
+                    sourceCat=sourceCat,
```
PaulPrice: Move to separate commit.

python/lsst/pipe/tasks/calibrate.py
natelust: I meant to talk to you about this while you were here. Jim and I discussed it at length with Russel. The problem is that there are many cases where you do very similar but not the same operations with these tasks. Having them all together then necessitated having a complex series of config options and function parameters to control the execution flow. We decided a bit of duplication was acceptable to simplify the execution logic. I have a proposal for a type of mixin that would reduce the boiler plate for much of this at some point in the future. It might be worth talking about at the Aug meeting. If it is ok with you, the verbosity will remain for this ticket, and will hopefully be addressed in the furture.

python/lsst/pipe/tasks/calibrate.py
natelust: Updated the text, to reflect the behavior of the code

python/lsst/pipe/tasks/calibrate.py
natelust: changed

python/lsst/pipe/tasks/calibrate.py
natelust: check

python/lsst/pipe/tasks/calibrate.py
natelust: check

python/lsst/pipe/tasks/calibrate.py
natelust: check

PR #57, opened 2016-06-07
Review for DM-2837

pschella:


tests/testColorterm.py
```diff
@@ -54,7 +54,9 @@
 class ColortermTestCase(unittest.TestCase):
     """A test case for MaskedImage"""
     def setUp(self):
-        self.sources = dict(g=0.0, r=0.0, true_g=-0.00928), dict(g=0.0, r=-1.0, true_g=-0.09168)
+        self.sources = (dict(g=0.0, r=0.0, fluxErr_g=0.0, fluxErr_r=0.0, true_g=-0.00928, true_fluxErr_g=0.0),
+                        dict(g=0.0, r=-1.0, fluxErr_g=1.0, fluxErr_r=1.0, true_g=-0.09168, 
+                                                                          true_fluxErr_g=0.92129230974756315))
```
pschella: Where do all these hard-coded values come from?

tests/testColorterm.py
```diff
@@ -102,6 +104,13 @@ def testTransformMags(self):
         for s in self.sources:
             self.assertEqual(ct.transformMags(s[ct.primary], s[ct.secondary]), s["true_g"])
 
+    def testPropagateFluxErrors(self):
+        """Check if we can propagate flux errors"""
+
+        ct = self.colorterms.getColorterm("g", photoCatName="ham")
+        for s in self.sources:
+            self.assertEqual(ct.propagateFluxErrors(s["fluxErr_g"], s["fluxErr_r"]), s["true_fluxErr_g"])
```
pschella: Use `assertAlmostEqual` for `float`

tests/testPhotoCal.py
```diff
@@ -42,6 +43,15 @@
 Log(Log.getDefaultLog(), "meas.astrom.sip", Log.WARN)
 Log(Log.getDefaultLog(), "astrometricSolver", Log.WARN)
 
+testColorterms = ColortermLibrary(data={
+    "test*": ColortermDict(data={
```
pschella: Probably correct. But why is that `*` here?

tests/testPhotoCal.py
```diff
@@ -71,11 +81,34 @@ def setUp(self):
...
+        self.schema = self.matches[0].second.schema
+
+        self.config = PhotoCalConfig()
+
+        # The test and associated data have been prepared on the basis that we
+        # use the PsfFlux to perform photometry.
+        self.config.fluxField = "base_PsfFlux_flux"
+
+        self.config.doWriteOutput = False    # schema is fixed because we already loaded the data
+
```
pschella: A bit too many blank lines. This can improve readability, but in this case I would say it reduces it due to fewer lines with content fitting on a screen.

tests/testPhotoCal.py
```diff
@@ -87,38 +120,77 @@ def getAstrometrySolution(self, loglvl = Log.INFO):
         )
 
     def testGetSolution(self):
-        res = self.getAstrometrySolution(loglvl=Log.DEBUG)
-        self.assertTrue(res is not None)
-        self.assertGreater(len(res.matches), 50)
+        self.assertTrue(self.res is not None)
```
pschella: Use `assertIsNot`

tests/testPhotoCal.py
```diff
@@ -87,38 +120,77 @@ def getAstrometrySolution(self, loglvl = Log.INFO):
...
 
         print 'Test1'
 
-        logLevel = Log.DEBUG
-        log = Log(Log.getDefaultLog(),
-                  'testPhotoCal',
-                  logLevel)
+        task = PhotoCalTask(config=self.config, schema=self.schema)
+        pCal = task.run(exposure=self.exposure, matches=self.matches)
+        print "Ref flux fields list =", pCal.arrays.refFluxFieldList
```
pschella: Don't know if this is common. I would expect it should either be logged as info or silenced. But again I don't know what is common practice in tests.

tests/testPhotoCal.py
```diff
@@ -87,38 +120,77 @@ def getAstrometrySolution(self, loglvl = Log.INFO):
...
-        pCal = task.run(exposure=self.exposure, matches=matches)
+        self.config.applyColorTerms = True
+        self.config.colorterms = testColorterms
+        self.config.photoCatName = "test"
+        zeroPointOffset = testColorterms.data['test*'].data['i'].c0
+
+        task = PhotoCalTask(config=self.config, schema=self.schema)
+        pCal = task.run(exposure=self.exposure, matches=self.matches)
         print "Ref flux fields list =", pCal.arrays.refFluxFieldList
         refFluxField = pCal.arrays.refFluxFieldList[0]
```
pschella: Only used once, so just refer to it directly.

tests/testPhotoCal.py
PaulPrice: The colorterm loading allows matching with wildcards, but that functionality doesn't appear to be tested here (`self.config.photoCatName = "test" `) so it should be removed or the test should be changed to use a `photoCatName = "testFooBar"` or similar.

tests/testColorterm.py
jdswinbank: `true_fluxErr_g` should be below `g=0.0`.

tests/testPhotoCal.py
```diff
@@ -34,6 +34,7 @@
 import lsst.utils.tests            as utilsTests
 from lsst.pex.logging import Log
 from lsst.pipe.tasks.photoCal import PhotoCalTask, PhotoCalConfig
+from lsst.pipe.tasks.colorterms import Colorterm, ColortermDict, ColortermLibrary, ColortermNotFoundError
```
jdswinbank: `ColortermNotFoundError` is unused.

tests/testPhotoCal.py
```diff
@@ -71,11 +81,34 @@ def setUp(self):
...
     def tearDown(self):
         del self.srcCat
         del self.conf
         del self.exposure
         del self.refObjLoader
+        del self.res
+        del self.matches
+        del self.log
+        del self.schema
+        del self.config
```
jdswinbank: I don't think you need to delete `self.config`

PR #59, opened 2016-06-09
Tickets/dm 5932

r-owen:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -65,6 +65,9 @@ class ImageDifferenceConfig(pexConfig.Config):
         doc="Use a simple gaussian PSF model for pre-convolution (else use fit PSF)? "
             "Ignored if doPreConvolve false.")
     doDetection = pexConfig.Field(dtype=bool, default=True, doc="Detect sources?")
+    doDecorrelation = pexConfig.Field(dtype=bool, default=False,
+        doc="Perform diffim decorrelation correction similar to Kaiser? If True, also updates the "
```
r-owen: "similar to Kaiser" is not very informative. A more formal reference would be very helpful. As a very picky point, I suggest "update" instead of "updates" for consistency with the first sentence.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -516,6 +519,17 @@ def run(self, sensorRef, templateIdList=None):
             mask  = subtractedExposure.getMaskedImage().getMask()
             mask &= ~(mask.getPlaneBitMask("DETECTED") | mask.getPlaneBitMask("DETECTED_NEGATIVE"))
 
+        # Compute ALZC correction kernel from matching kernel
+        # Here we use a constant kernel, just compute it for the center of the image.
+        # Investigating correction for spatially varying kernels is a subject of other tickets.
```
r-owen: I suggest removing this comment. The user can look up the definition of doDecorrelation above, and that's all you are doing here. Plus the log message is a clue. Details of the algorithm belong in the docs for "performExposureDecorrelation", since that may evolve over time, leaving the information here outdated.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -516,6 +519,17 @@ def run(self, sensorRef, templateIdList=None):
...
+        if self.config.doDecorrelation:
+            self.log.info("Running decorrelation correction.")
+            from lsst.ip.diffim.imageDecorrelation import performExposureDecorrelation as doDecor
```
r-owen: I suggest importing at the top with everything else.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -516,6 +519,17 @@ def run(self, sensorRef, templateIdList=None):
...
+            subtractedExposure, _ = doDecor(templateExposure, exposure, subtractedExposure,
+                                            subtractRes.psfMatchingKernel, self.log)
+            self.log.info("Finished running decorrelation correction.")
```
r-owen: I suggest removing this log message. One is enough.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -516,6 +519,17 @@ def run(self, sensorRef, templateIdList=None):
             mask  = subtractedExposure.getMaskedImage().getMask()
             mask &= ~(mask.getPlaneBitMask("DETECTED") | mask.getPlaneBitMask("DETECTED_NEGATIVE"))
 
+        # Compute ALZC correction kernel from matching kernel
+        # Here we use a constant kernel, just compute it for the center of the image.
+        # Investigating correction for spatially varying kernels is a subject of other tickets.
+        if self.config.doDecorrelation:
+            self.log.info("Running decorrelation correction.")
```
r-owen: Super picky, but "decorrelation correction" sounds like a double-negative to me. You are decorrelating, which corrects for over-correlating, right? Perhaps "running decorrelation" or "running image decorrelation"?

python/lsst/pipe/tasks/imageDifference.py
djreiss: OK, changed to "Perform diffim decorrelation to undo pixel correlation due to convolution?"

python/lsst/pipe/tasks/imageDifference.py
djreiss: Removed.

python/lsst/pipe/tasks/imageDifference.py
djreiss: OK.

python/lsst/pipe/tasks/imageDifference.py
djreiss: OK.

python/lsst/pipe/tasks/imageDifference.py
djreiss: True enough. Fixed.

PR #60, opened 2016-06-10
Tickets/dm 6459

n8pease:


tests/testCoadds.py
```diff
@@ -144,11 +144,11 @@ def tearDown(self):
...
 
     def getCalexpIds(self, butler):
-        if len(butler.repository._parents) is not 1:
+        if len(butler.inputs) is not 1:
             raise RuntimeError("This test assumes one input repository, can not continue with %s"
-                               % butler.repository._parents)
+                               % butler.repository.inputs)
 
-        return butler.repository._parents[0]._mapper.index['calexp']['visit'][None]
+        return butler.inputs[0].repo._mapper.index['calexp']['visit'][None]
```
ktlim: It would be nicer if this didn't have to know about the mapper internals (or that there's even a mapper at all) and could just use a standard Butler interface.  But it isn't a big deal, as this is just a test.

n8pease: I agree that the mappers (and mapper internals) should be private. I think that's acknowledged by everyone and all agree that this is an acceptable violation of that since it's in a test.
In this case, the index parameter is unique to lsst.pipe.tasks.mocks.simpleMapper.SimpleMapper. One way I can think of to do it would be to create a dataset type for the parameter and have the mapper special case that dataset type to return the value for, in this case, self.index['calexp']['visit'][None]
(I think it would probably want RFCing, which makes it larger than the scope of this story, IMO)

python/lsst/pipe/tasks/ingest.py
```diff
@@ -436,7 +436,7 @@ def isBadId(self, info, badIdList):
     def run(self, args):
         """Ingest all specified files and add them to the registry"""
         filenameList = sum([glob(filename) for filename in args.files], [])
-        root = args.butler.repository._mapper.root
+        root = args.input #args.butler.repository._mapper.root
```
TallJimbo: Unless the old version is somehow descriptive, should just delete the comment.

PR #61, opened 2016-06-14
Update run method call of SourceDeblendTask

parejkoj:
The run method of SourceDeblendTask has been updated in the previous
commit to extract the psf from the provided exposure. This commit
updates all calls of the run method to match the new interface.


python/lsst/pipe/tasks/detectAndMeasure.py
```diff
@@ -189,7 +189,6 @@ def run(self, exposure, exposureIdInfo, background=None, allowApCorr=True):
             self.deblend.run(
                 exposure = exposure,
```
parejkoj: This doesn't need to be 3 lines any more: it'll should fit on a single line just fine.

PR #63, opened 2016-06-23
Add DecorrelateALKernelTask as a subtask

djreiss:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -167,6 +170,8 @@ def setDefaults(self):
         # DiaSource Detection
         self.detection.thresholdPolarity = "both"
         self.detection.thresholdValue = 5.5
+        if self.doDecorrelation:
+            self.detection.thresholdValue = 5.0
```
r-owen: Please add a comment explaining why you change this value.

Also, since you only run decorrelation if doSubtract is True, I suggest you apply the same test here. That makes the "doCorrelate" config parameter a bit easier to explain (e.g. "ignored if doSubtract False")

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -108,6 +107,10 @@ class ImageDifferenceConfig(pexConfig.Config):
         target=ImagePsfMatchTask,
         doc="Warp and PSF match template to exposure, then subtract",
     )
+    decorrelate = pexConfig.ConfigurableField(
+        target=ImageDecorrelationTask,
+        doc="Decorrelate image difference",
+    )
```
r-owen: Please add a comment noting that it will be ignored if doSubtract is False (other than affecting the default value of detection.thresholdValue, which I think should be fixed).

Also, a brief explanation of WHY you only decorrelate if doSubtract is true would be nice. I assume it is because there is no noise to decorrelate.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -36,10 +36,9 @@
 from lsst.ip.diffim import ImagePsfMatchTask, DipoleAnalysis, \
     SourceFlagChecker, KernelCandidateF, cast_KernelCandidateF, makeKernelBasisList, \
     KernelCandidateQa, DiaCatalogSourceSelectorTask, DiaCatalogSourceSelectorConfig, \
-    GetCoaddAsTemplateTask, GetCalexpAsTemplateTask, DipoleFitTask
+    GetCoaddAsTemplateTask, GetCalexpAsTemplateTask, DipoleFitTask, ImageDecorrelationTask
```
RobertLuptonTheGood: Sorry, I only just noticed that you called this, "ImageDecorrelationTask".  This name is misleading, and should be changed.  It decorrelates the effects of an A&L kernel, but doesn't deal with general correlations (e.g. due to warping).

python/lsst/pipe/tasks/imageDifference.py
djreiss: Good points. Done and done.

python/lsst/pipe/tasks/imageDifference.py
djreiss: Do you actually mean a comment? Or add it to the docstring? I will add a comment.

python/lsst/pipe/tasks/imageDifference.py
djreiss: OK. I will rename it to DecorrelateALKernelTask, unless anyone can think of a better name.

python/lsst/pipe/tasks/imageDifference.py
r-owen: I meant please add the explanation it to the doc string. In general if a config field needs explanation, that explanation should go into the doc string.

PR #64, opened 2016-06-28
add fits reader

SimonKrughoff:


tests/testIndexReference.py
```diff
@@ -0,0 +1,168 @@
...
+        for badHdu in [0, 3, 4]:
+            config = FitsReaderTask.ConfigClass()
+            config.hdu = badHdu
+            task = FitsReaderTask(config=config)
+            with self.assertRaises(Exception):
+                task.readFile(FitsPath)
+
+
+class MyMemoryTestCase(lsst.utils.tests.MemoryTestCase):
+    pass
```
SimonKrughoff: Is it legal to have a memory test that just passes?

tests/testIndexReference.py
```diff
@@ -0,0 +1,168 @@
+from __future__ import absolute_import, division, print_function
```
SimonKrughoff: There is already a test that tests some of the functionality of the indexer.  It might be worthwhile thinking about combining them.

tests/testIndexReference.py
timj: Yes. Take a look at https://developer.lsst.io/coding/python_testing.html

tests/testIndexReference.py
r-owen: I think it is important to have tests of the tasks themselves. I'll file a ticket to add a test of the file loader to that file. You do have a test that tests the indexer, and I am making the assumption that if the task behaves, then it will behave when used by the indexer.

PR #67, opened 2016-07-21
Fix dipoleFitting to use correct matched template exposure

djreiss:
Fix mixup of var1 and var2 for science and template

### DIFF
--- a/python/lsst/pipe/tasks/imageDifference.py
+++ b/python/lsst/pipe/tasks/imageDifference.py
@@ -527,7 +527,7 @@ class ImageDifferenceTask(pipeBase.CmdLineTask):
         # If doSubtract is False, then subtractedExposure was fetched from disk (above), thus it may have
         # already been decorrelated. Thus, we do not do decorrelation if doSubtract is False.
         if self.config.doDecorrelation and self.config.doSubtract:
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -561,7 +561,11 @@ def run(self, sensorRef, templateIdList=None):
                 if not self.config.doDipoleFitting:
                     self.measurement.run(diaSources, subtractedExposure)
                 else:
-                    self.measurement.run(diaSources, subtractedExposure, exposure, templateExposure)
+                    if self.config.doSubtract:
```
parejkoj: I'm not clear on why "doSubtract" means we pass the matchedExposure here?

parejkoj:
Just the one question. This seems straightforward otherwise.

python/lsst/pipe/tasks/imageDifference.py
djreiss: If `doSubract == True` then we have a matchedExposure and we can use that rather than computing it from `exposure - subtractedExposure`. This is noted in the `DipoleFitTask.run()` docstring.

PR #68, opened 2016-07-22
Tickets/dm 6352

r-owen:


SimonKrughoff:
Did you confirm that the example runs with the new changes?

r-owen:
I only changed one example (the only one I found that needed it) and I did confirm that it runs afterwards. It did not run before, due to bit rot.

tests/testFindAstrometryNetDataDir.py
```diff
@@ -29,7 +29,7 @@ def setupAstrometryNetDataDir(name, rootDir=None, verbose=False):
 class TestAstrometryNetDataDirDiscovery(unittest.TestCase):
 
     def test_photocal(self):
-        datapath = setupAstrometryNetDataDir('photocal')
+        datapath = setupAstrometryNetDataDir('t2155')
```
SimonKrughoff: This seems weird to me.  Why do you need a build tag to specify the a.net data directory?

tests/testPhotoCal.py
```diff
@@ -114,7 +117,6 @@ def tearDown(self):
     def getAstrometrySolution(self, loglvl=Log.INFO):
         astromConfig = measAstrom.AstrometryTask.ConfigClass()
         astrom = measAstrom.AstrometryTask(config=astromConfig, refObjLoader=self.refObjLoader)
-        # use solve instead of run because the exposure has the wrong bbox
```
SimonKrughoff: Why delete this comment?  It seems somewhat helpful, and if it's not, better to update the comment than delete it, no?

examples/photoCalTask.py
```diff
@@ -75,19 +70,25 @@ def run():
     exposure, srcCat = loadData()
     schema = srcCat.getSchema()
     #
+    # Create the reference catalog loader
+    #
+    refCatDir = os.path.join(lsst.utils.getPackageDir('meas_astrom'), 'tests', 'data', 'sdssrefcat')
```
SimonKrughoff: The data look like they are in `pipe_tasks` not `meas_astrom`.  Does this example run with these changes?

tests/testFindAstrometryNetDataDir.py
r-owen: This isn't a build tag, it's the name of a directory inside tests/astrometry_net_data. No other test was using the a.net version of "photocal" and this one test didn't care which set of a.net data it used, so I switched it to use the other available a.net data and discarded the now-unused "photocal" data.

tests/testPhotoCal.py
r-owen: The comment is dead wrong. `AstrometryTask.solve` does not support specifying a different bbox, so I made sure the bbox of the exposure was correct. There is no strong reason to prefer `solve` over `run`, but `run` is a thin layer that calls `solve` or `loadAndMatch`, depending on a config flag, and what I wanted was `solve`. In general I think it better to call `AstrometryTask.solve` or `loadAndMatch` rather than run, if you are sure what you want the task to do, and the task is documented as such.

examples/photoCalTask.py
r-owen: That data appears in both packages, and is identical in both. I'm not thrilled about this duplication, but in general we prefer test data in packages to be self-contained, so that modifying test data in one package won't mysteriously break tests in other packages. And indeed both packages had duplicate a.net versions of this data (`tests/astrometry_net_data/photocal`). I was able to delete one of those, and the other will move to the new a.net wrapper package as part of separating moving the a.net wrapper code into its own package.

jdswinbank:
DM-6352 is long merged.

PR #69, opened 2016-08-04
DM-6999: Logging framework migration

hchiang2:


bin.src/makeCoaddTempExp.py
```diff
@@ -20,9 +20,5 @@
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
-import lsst.pex.logging as pexLog
 from lsst.pipe.tasks.makeCoaddTempExp import MakeCoaddTempExpTask
-
-pexLog.Trace.setVerbosity('lsst.coadd', 3)
-pexLog.Trace.setVerbosity('lsst.ip.diffim', 1)
```
hchiang2: @r-owen  , I removed these because I thought they were accidentally left here. Are they here intentionally?  If so I will add them back in the `lsst.log` form.

r-owen: That's a great question and I'm afraid I don't know the answer. I had assumed they provided an appropriate level of verbosity, but perhaps it is too much.

RobertLuptonTheGood: I don't think tasks should be messing with trace levels.  It's up to the user to turn tracing on.

hchiang2: Agree.  I'll keep this commit, with those verbosity setting removed. 

If coadd and ip_diffim need to be more verbose, we probably should change the logging records to a higher level in coadd and ip_diffim.  (I can file a ticket if that is wanted)

PR #71, opened 2016-08-10
Review for DM-6635: typo in CoaddSrcTransformTask

jdswinbank:


tests/testTransform.py
```diff
@@ -266,11 +269,74 @@ def testInterface(self):
...
+        # Persist a coadd measurement config.
+        # We disable all measurement plugins so that there's no actual work
+        # for the TransformTask to do.
+        measCfg = MeasureMergedCoaddSourcesConfig()
+        measCfg.measurement.plugins.names = []
+        self.butler.put(measCfg, self.MEASUREMENT_CONFIG_DATASET)
+
+        # The record the type of coadd on which our supposed measurements have
+        # been carried out: we need to check this was propagated to the
+        # transformation task.
```
fred3m: Remove extra "the".

tests/testTransform.py
```diff
@@ -266,17 +272,75 @@ def testInterface(self):
...
+class MemoryTestCase(utilsTests.MemoryTestCase):
+    pass
+
+
+def setup_module(module):
+    utilsTests.init()
 
 if __name__ == "__main__":
-    run(True)
+    setup_module(sys.modules[__name__])
```
fred3m: I don't think that this line is necessary. It is my understanding that this block will be removed once the conversion to pytest is complete but the dependence on the unused "sys" package might remain. I would recommend either moving `import sys` to this block or just calling `utilsTests.init()` directly.

timj: There is a philosophical debate as to whether we are better off duplicating the content of `setup_module` or having one call the other. There is currently no mandate to remove the `__main__` from the code (people want python to be able to run them from the command line). Maybe when people are comfortable always using pytest we will revisit the topic.

PR #73, opened 2016-08-27
Remove 'flags=sourceWriteFlags' from dataRef.put in CalibrateTask.

wmwv:
The Butler 'put' does not support a 'flags' option to pass down
to the underlying catalogs.  This has meant that CalibrateTask
has been broken, but we don't fully test CalibrateTask.

This pull request is just a one-line change to remove the flags option.


jdswinbank:
Merged in 2111a3b012c0eabe3569fee3776f91aa286fe592.

PR #76, opened 2016-09-09
DM-6815: Update LSST full-stack processing configuration to match best practice from HSC

PaulPrice:


python/lsst/pipe/tasks/coaddBase.py
```diff
@@ -66,7 +66,7 @@ class CoaddBaseConfig(pexConfig.Config):
     doApplyUberCal = pexConfig.Field(
         dtype = bool,
         doc = "Apply meas_mosaic ubercal results to input calexps?",
-        default = False
+        default = True
```
PaulPrice: I'm concerned this will break cameras other than HSC running with meas_mosaic.

PR #77, opened 2016-09-15
DM-7292: Port to Python 3

fred3m:


tests/testProcessCcd.py
```diff
@@ -183,9 +184,12 @@ def assertCatalogsEqual(self, catalog1, catalog2, skipCols=()):
         self.assertEqual(catalog1.schema, catalog2.schema)
         self.assertEqual(len(catalog1), len(catalog2))
         d = catalog1.schema.extract("*")
+        def fixNaN(x):
+            if x!=x:
+                return "NaN"
         fixNaN = lambda x: x if x == x else "NaN"
```
laurenam: Should this lambda be removed with the introduction of the fixNaN function?

fred3m: Yes, I think I was in the middle of making this change when something interrupted me and since it didn't break anything I must have forgotten about it. Great catch.

python/lsst/pipe/tasks/mocks/mockCoadd.py
```diff
@@ -197,8 +197,8 @@ def buildInputImages(self, butler, obsCatalog=None, truthCatalog=None, tract=0):
                 if status:
                     simSrcRecord = simSrcCatalog.addNew()
                     simSrcRecord.setCoord(truthRecord.getCoord())
-                    simSrcRecord.setL(self.objectIdKey, truthRecord.getId())
-                    simSrcRecord.setL(self.exposureIdKey, obsRecord.getId())
+                    simSrcRecord.setI(self.objectIdKey, truthRecord.getId())
+                    simSrcRecord.setI(self.exposureIdKey, obsRecord.getId())
```
natelust: Have you verified that integers can hold the values we use, and won't be too short?

tests/testProcessCcd.py
```diff
@@ -137,19 +137,19 @@ def testProcessCcd(self):
                     print("variance mean = %r, stdDev = %r" % (varMean, varStdDev))
                     print("psf Ixx = %r, Iyy = %r, Ixy = %r" % (psfIxx, psfIyy, psfIxy))
 
-                    self.assertAlmostEqual(bgMean, 191.51595080958367, places=7)
-                    self.assertAlmostEqual(bgStdDev, 0.22492169148323429, places=7)
+                    self.assertAlmostEqual(bgMean, 191.51595080958367, places=3)
+                    self.assertAlmostEqual(bgStdDev, 0.22492169148323429, places=3)
```
natelust: 7 places to 3 places is a huge change! Do differences in software stacks really produce different numbers at the 10^-(3/4) level? If so that is worrying. It implies we might have reproducibility issues in the future. Is this test able to be tightened up to ~5? What are the magnitude of differences we are seeing between versions? Is this related to changing to ints places? Are we losing precision in calculations? I know these are kind of big questions to ask, but making changes to unit tests to make things work scares me unless we can really justify why the change is nessisary.

tests/testProcessCcd.py
```diff
@@ -133,7 +133,13 @@ def testProcessCcd(self):
...
                     print("variance mean = %r, stdDev = %r" % (varMean, varStdDev))
                     print("psf Ixx = %r, Iyy = %r, Ixy = %r" % (psfIxx, psfIyy, psfIxy))
-
+                    
+                    # TODO: these thresholds are larger than they ought to be because the 
+                    # background modeling differs in py2 and py3.
+                    # The values asserted here are the precise result that py2 gives. The precision below 
+                    # should be tightened and this comment removed in the course of DM-8017.
+                    # Also see: https://community.lsst.org/t/difference-in-py2-and-py3-background-models/1240
+
```
natelust: I guess disregard the past comment as I got to this last. As there is a ticket to do this work, I see no need to do the work on this ticket.

python/lsst/pipe/tasks/mocks/mockCoadd.py
timj: There is much confusion over `L` vs `I` in the C++ interface layer. Python (2 or 3) both use at least 64-bits for normal `int` type but in most cases C++ treats `I` as a 32-bit integer and that would indeed be too short for our 64-bit IDs.

See the discussion on [RFC-227](https://jira.lsstcorp.org/browse/RFC-227).

PR #78, opened 2016-09-28
Update for exposure time and date not in Calib

r-owen:
Update all code that used exposure time or date in Calib,
switching to VisitInfo instead.

### DIFF
--- a/python/lsst/pipe/tasks/mocks/mockObservation.py
+++ b/python/lsst/pipe/tasks/mocks/mockObservation.py
@@ -93,6 +93,8 @@ class MockObservationTask(lsst.pipe.base.Task):
         @param[in] camera: camera geometry (an lsst.afw.cameraGeom.Camera)
         @param[in] catalog: catalog to which to add observations (an ExposureCatalog);
             if None then a new catalog is created.
+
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/tasks/mocks/mockObservation.py
```diff
@@ -100,7 +100,16 @@ def run(self, butler, n, tractInfo, camera, catalog=None):
             if not catalog.getSchema().contains(self.schema):
                 raise ValueError("Catalog schema does not match Task schema")
         visit = 1
+
         for position, pa in self.makePointings(n, tractInfo):
+            visitInfo = lsst.afw.image.makeVisitInfo(
+                exposureTime = self.config.expTime,
+                date = lsst.daf.base.DateTime.now(),
+                boresightRaDec = position,
+                # once we know what pa is, we can do something like the following
```
PaulPrice: It not clear to me what this comment means.

I don't like leaving code in but commented-out, even if there's an explanation.

r-owen:
Good catch. I changed the commented out code to a @todo in the doc string. Rebased and pushed.

PR #81, opened 2016-10-14
Tickets/dm 7976

r-owen:


tests/testCoaddInputs.py
```diff
@@ -0,0 +1,256 @@
...
+from lsst.afw.coord import Coord, IcrsCoord
+import lsst.afw.geom
+from lsst.afw.geom import degrees
+from lsst.afw.geom.polygon import Polygon
+import lsst.afw.image
+from lsst.afw.detection import GaussianPsf
+from lsst.afw.math import ChebyshevBoundedField
+from lsst.pipe.tasks.coaddInputRecorder import CoaddInputRecorderTask
+
+SaveCoadd = False  # if True then save coadd even if test passes (always saved if a test fails)
```
TallJimbo: Could this lead to problems if the test fails for some new reason, and then confusingly continues to fail because the committed test data file has been overwritten by something incorrect?

tests/testCoaddInputs.py
```diff
@@ -0,0 +1,256 @@
...
+        self.coadd.setWcs(exp0.getWcs())
+
+        coaddInputs = coaddInputRecorder.makeCoaddInputs()
+        for exp in self.exposures:
+            coaddInputRecorder.addVisitToCoadd(coaddInputs=coaddInputs, coaddTempExp=exp, weight=1.0/numExp)
+        self.coadd.getInfo().setCoaddInputs(coaddInputs)
+
+
+class CoaddInputsTestCase(lsst.utils.tests.TestCase):
+    def setUp(self):
```
TallJimbo: Need an empty line before this one, right?

tests/testCoaddInputs.py
```diff
@@ -0,0 +1,256 @@
...
+from lsst.pipe.tasks.coaddInputRecorder import CoaddInputRecorderTask
+
+SaveCoadd = False  # if True then save coadd even if test passes (always saved if a test fails)
+
+
+class MockExposure(object):
+    """Make a simple mock exposure suitable to put in a coadd
+
+    The metadata is set, but not the pixels
+    """
```
TallJimbo: Reads line a method doc, not a class doc; maybe "Helper class for making simple mock exposures..."  I also think you need an empty line after it.

tests/testCoaddInputs.py
```diff
@@ -0,0 +1,256 @@
...
+        self.coadd.writeFits(coaddPath)
+        coadd = lsst.afw.image.ExposureF(coaddPath)
+        coaddInputs = coadd.getInfo().getCoaddInputs()
+        self.assertCoaddInputsOk(coaddInputs, version=self.version)
+        if not SaveCoadd:
+            os.unlink(coaddPath)
+        else:
+            print("SaveCoadd true; saved coadd as: %r" % (coaddPath,))
+
+    def xtestReadV1Coadd(self):
```
TallJimbo: Unintentionally disabled?

tests/testCoaddInputs.py
```diff
@@ -0,0 +1,256 @@
...
+            self.assertPsfsNearlyEqual(visit.getPsf(), exp.getPsf())
+            self.assertPairsNearlyEqual(visit.getCalib().getFluxMag0(), (1.1e12, 2.2e10))
+            self.assertEqual(len(visit.getApCorrMap()), 3)
+            self.assertEqual(set(visit.getApCorrMap().keys()), set(expInfo.getApCorrMap().keys()))
+            self.assertFloatsAlmostEqual(np.array(visit.getValidPolygon().getVertices()),
+                                         np.array(expInfo.getValidPolygon().getVertices()))
+            if version > 1:
+                self.assertEqual(visit.getVisitInfo(), expInfo.getVisitInfo())
+            else:
+                self.assertFalse(visit.getVisitInfo().getDate().isValid())
```
TallJimbo: Just a reminder that if you do change the code in afw to make this an empty pointer, you'll want to change this line too (to check that `getVisitInfo() is None`).

tests/testCoaddInputs.py
```diff
@@ -0,0 +1,256 @@
...
+    def xtestReadV1Coadd(self):
+        """Read a coadd that contains version 1 CoaddInputs"""
+        coaddPath = self.getCoaddPath(version=1)
+        coadd = lsst.afw.image.ExposureF(coaddPath)
+        coaddInputs = coadd.getInfo().getCoaddInputs()
+        self.assertCoaddInputsOk(coaddInputs, version=1)
+
+    def assertCoaddInputsOk(self, coaddInputs, version):
+        self.assertIsNotNone(coaddInputs)
+        visitTable = coaddInputs.visits
```
TallJimbo: If you're going to test just one of these, it should be `ccds`, not `visits`; most of these quantities in the `visits` table come from the `coaddTempExp`, which is *probably* better than setting them to None but might still be confusing in some cases (e.g. "the visit WCS is the same as the coadd WCS").  The `ccds` table is what corresponds to what is in a `calexp`, and that's what most people will want to look at and hence what we want to make sure doesn't regress.

tests/testCoaddInputs.py
r-owen: Yes, I did realize that. And it improves the test.

PR #83, opened 2016-10-20
DM-7129: temporary coadd covariance experiments

TallJimbo:
This PR is for review comments only; it should not be merged to master.

### DIFF
--- a/python/lsst/pipe/tasks/makeCoaddTempExp.py
+++ b/python/lsst/pipe/tasks/makeCoaddTempExp.py
@@ -63,7 +63,7 @@ class MakeCoaddTempExpTask(CoaddBaseTask):
     """
     ConfigClass = MakeCoaddTempExpConfig
     _DefaultName = "makeCoaddTempExp"
-    
+
     def __init__(self, *args, **kwargs):
         CoaddBaseTask.__init__(self, *args, **kwargs)
         self.makeSubtask("warpAndPsfMatch")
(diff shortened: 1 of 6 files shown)


tests/testCovariance.py
```diff
@@ -0,0 +1,425 @@
...
+some cases where it's easy and/or particularly important to do so (e.g. CoaddPsf); we mostly
+just check that everything runs, and that the results make enough sense to let us proceed
+to the next step.
+
+NOTE: if this test fails with what looks like a failure to load a FITS file, try changing
+the REUSE_DATAREPO variable below to False, as sometimes this error message indicates a
+different problem that's revealed when we're not trying to cache the mock data between
+tests (but set REUSE_DATAREPO back to True when done debugging, or this test will be very
+slow).
+"""
```
TallJimbo: While I know this is just a temporary copy, I think it'd be worthwhile to delete all the tests that aren't actually testing anything new here (and shrink this description accordingly).

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -164,8 +183,22 @@ def createTempExp(self, calexpRefList, skyInfo, visitId=0):
                 calExpRef = calExpRef.butlerSubset.butler.dataRef("calexp", dataId=calExpRef.dataId,
                                                                   tract=skyInfo.tractInfo.getId())
                 calExp = self.getCalExp(calExpRef, bgSubtracted=self.config.bgSubtracted)
-                exposure = self.warpAndPsfMatch.run(calExp, modelPsf=modelPsf, wcs=skyInfo.wcs,
-                                                    maxBBox=skyInfo.bbox).exposure
+                covName = 'covar'
+                for key, val in calExpRef.dataId.items():
+                    covName += '_%s_%s'%(key, val)
```
TallJimbo: I don't actually see this being used anywhere.  Did you intend to write the covariance out with the butler, and then decide not to?  I do think that'd be worth supporting, as it's a lot easier to explore the results after the processing is complete, especially if the processing is at all slow.  And I think you'll need it anyway when you move on to coadding the covariances.

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -58,28 +59,46 @@ class MakeCoaddTempExpConfig(CoaddBaseTask.ConfigClass):
     )
 
 
+class covView(object):
+    def __init__(self, destImage, covImage):
```
AstroVPK: Fixed....

python/lsst/pipe/tasks/makeCoaddTempExp.py
AstroVPK: I was not sure if writing the covariance matrix for an individual exposure made sense but thought I'd leave the name code to do so in there anyway. I was planning on only writing out the per-visit covariance matrix.

tests/testCovariance.py
AstroVPK: Done...

python/lsst/pipe/tasks/makeCoaddTempExp.py
AstroVPK: I've added a small amount of extra functionality to store the name of the covariance matrix in the the CovView object but I think that until I have the covariance matrix for the full visit, I'm going to defer figuring out how to persist the covariance matrices for the individual exposures.

PR #86, opened 2016-11-22
DM-8088: When making PSF-matched coadds, warp first then PSF-match

jmeyers314:


python/lsst/pipe/tasks/warpAndPsfMatch.py
```diff
@@ -51,31 +59,63 @@ def __init__(self, *args, **kwargs):
...
+        modelPsf : :cpp:class: `lsst.meas.algorithms.KernelPsf`
+            target PSF to which to match (or None)
+        maxBBox : :cpp:class:`lsst::afw::geom::Box2I` or None
+            maximum allowed parent bbox of warped exposure;
             if None then the warped exposure will be just big enough to contain all warped pixels;
             if provided then the warped exposure may be smaller, and so missing some warped pixels;
             ignored if destBBox is not None
-        @param destBBox: exact parent bbox of warped exposure (an afwGeom.Box2I or None);
+        destBBox: :cpp:class: `lsst::afw::geom::Box2I` or None
+                exact parent bbox of warped exposure (an );
```
jmeyers314: Some junk at the end of the line here...

python/lsst/pipe/tasks/warpAndPsfMatch.py
```diff
@@ -38,6 +41,11 @@ class WarpAndPsfMatchConfig(pexConfig.Config):
         dtype=afwMath.Warper.ConfigClass,
         doc="warper configuration",
     )
+    reverseWarpAndMatch = pexConfig.Field(
```
jmeyers314: It occurs to me that `reverseWarpAndMatch` only makes sense if you know what the nominal order of warp and match is.  Perhaps something like `matchFirst` or `matchThenWarp` would be more clear?

PR #88, opened 2016-11-29
Review for DM-8390: calibrateTask.py example broken

jdswinbank:


examples/calibrateTask.py
```diff
@@ -56,7 +56,10 @@ def loadData(pixelScale=1.0):
     imFile = os.path.join(mypath, "CFHT", "D4", "cal-53535-i-797722_small_1.fits")
 
     visitInfo = afwImage.makeVisitInfo(exposureTime=1.0)
-    exposure = afwImage.ExposureF(imFile, visitInfo)
+    exposureInfo = afwImage.ExposureInfo()
+    exposureInfo.setVisitInfo(visitInfo)
+    exposure = afwImage.ExposureF(imFile)
+    exposure.setInfo(exposureInfo)
```
r-owen: I'm going to suggest modifying the exposure to have EXPTIME in HDU0, so that VisitInfo picks it up automatically. It seems crazy that we're using such outdated image data in afwdata. But I realize that may be too much for fixing a lone example. So...if you prefer your approach (and it is more expedient) then I suggest letting the FITS constructor construct the ExposureInfo and then set the VisitInfo of it. Admittedly you run a small risk due to trying to read raw data using a ctor that prefers LSST-type headers, but it will pick up more metadata that way and it's simpler. Something like the following, where you might want to break the last line into 2 or 3 lines if you feel it enhances readability:
```
exposure = afwImage.ExposureF(imFile)
exposure.getInfo().setVisitInfo(afwImage.makeVisitInfo(exposureTime=1.0))
```

examples/calibrateTask.py
```diff
@@ -117,15 +120,14 @@ def loadAndMatch(self, exposure, sourceCat):
 
 def run(display=False):
     #
-    # Create the task using a butler constructed using the obs_test repo
+    # Create the tasks
     #
-    butler = dafPersistence.Butler(os.path.join(lsst.utils.getPackageDir("obs_test"), "data", "input"))
```
r-owen: I'm pretty sure you can also delete the line that imports `dafPersistence`. (If so flake8 should show you that; please do run that).

PR #90, opened 2017-01-03
Make HDU number available to ingest.

ctslater:
This is used for ingesting MEF files (such as from Decam).

### DIFF
--- a/python/lsst/pipe/tasks/ingest.py
+++ b/python/lsst/pipe/tasks/ingest.py
@@ -78,7 +78,10 @@ class ParseTask(Task):
                 break
             ext = self.getExtensionName(md)
             if ext in extnames:
-                infoList.append(self.getInfoFromMetadata(md, info=phuInfo.copy()))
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/ingest.py
```diff
@@ -78,7 +78,10 @@ def getInfo(self, filename):
                 break
             ext = self.getExtensionName(md)
             if ext in extnames:
-                infoList.append(self.getInfoFromMetadata(md, info=phuInfo.copy()))
+                hduInfo = self.getInfoFromMetadata(md, info=phuInfo.copy())
+                # We need the HDU number when registering MEF files.
+                hduInfo["hdu"] = extnum - 1
```
hsinfang: This makes `hdu` a very special dataId key name.  Have you checked if this could affect other obs packages?  
I wonder if we want this to be added to some higher-level documentations.

timj: HDU is very FITS-centric for a generic butler...

PR #94, opened 2017-01-27
DM-8461: Wrap pipe_tasks with pybind11

kfindeisen:
These mostly adapt the code to changes in the python API. The [DM-9112](https://jira.lsstcorp.org/browse/DM-9112) workaround is provided as a separate commit so that it can easily be reverted when that issue is resolved.


ups/pipe_tasks.table
```diff
@@ -1,10 +1,10 @@
+setupRequired(afw)
```
r-owen: Good change

r-owen: I suggest making meas_base required, as well, since it is used by mocks.

Also, while you are in here, please add `daf_persistence` as required (used in mocks/simpleMapper.py)

tests/SConscript
```diff
@@ -2,4 +2,5 @@
 import os
 from lsst.sconsUtils import scripts, env
 
-scripts.BasicSConscript.tests()
+# testCoadds.py must be entirely skipped, and Scons doesn't like that
+scripts.BasicSConscript.tests(ignoreList=['testCoadds.py'])
```
r-owen: Please supplement this code comment with some explanation of why testCoadds.py can no longer be run.

tests/testCoaddInputs.py
```diff
@@ -207,6 +207,7 @@ def getCoaddPath(self, version):
         return os.path.join(self.dataDir,
                             "testCoaddInputs_coadd_with_version_{}_data.fits".format(version))
 
+    @unittest.skip("Depends on BaseRecord get support, which will be changed in DM-8716.")
```
r-owen: I'm quite worried about this. I thought we had already properly wrapped BaseRecord.get.

I filed DM-8716 as a request to clean up ugly code, not to add functionality. (I am also hoping that `get` and `__getitem__` can be made do the same thing, instead of almost the same thing. But the small differences are inherited from SWIG)

tests/SConscript
r-owen: I now see the skips in the test itself, which answers the question, but it's not obvious from reading the message. Perhaps something like {{All tests in testCoadds.py are skipped and Scons doesn't like that}}.

It's a pity about Scons not liking that; it'd be much safer to leave the skips in place

tests/testCoaddInputs.py
kfindeisen: These skips were [requested](https://lsstc.slack.com/archives/dm-pybind11-porting/p1484959013000153) by @TallJimbo, in response to my asking why these methods used getL, setD, etc.

tests/SConscript
timj: I am very confused. Why does scons care if tests are skipped? SconsUtils just runs the tests and looks at the exit status. I didn't think that exit status was bad if all tests are skipped (assuming you are skipping them properly).

kfindeisen: @timj with `pytest`:

    tests/testCoadds.py ssssssssssss

    ============================================================== 12 skipped in 1.52 seconds ===============================================================

with `scons tests`:

    running tests/testCoadds.py... failed
    Failed test output:
    tests/testCoadds.py

    [redacted]/lsstsw2/miniconda/lib/python2.7/site-packages/astropy/config/configuration.py:687: ConfigurationMissingWarning: Configuration defaults will be used due to OSError:Could not find unix home directory to search for astropy config dir on None
      warn(ConfigurationMissingWarning(msg))
    [redacted]/lsstsw2/miniconda/lib/python2.7/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
      warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
    Traceback (most recent call last):
      File "tests/testCoadds.py", line 468, in <module>
        setup_module(sys.modules[__name__])
      File "[redacted]/lsstsw2/miniconda/lib/python2.7/unittest/case.py", line 62, in skip_wrapper
        raise SkipTest(reason)

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -585,8 +585,9 @@ def copyIcSourceFields(self, icSourceCat, sourceCat):
             self.log.warn("copyIcSourceFields doing nothing because icSourceFieldsToCopy is empty")
             return
 
-        closest = False  # return all matched objects
-        matches = afwTable.matchXy(icSourceCat, sourceCat, self.config.matchRadiusPix, closest)
+        mc = afwTable.MatchControl()
+        mc.findOnlyClosest = False # return all matched objects
```
r-owen: Please run flake8 linter on the changed code. In this case I expect it to complain about needing two spaces before the `#`.

python/lsst/pipe/tasks/mocks/mockCoadd.py
```diff
@@ -277,7 +277,8 @@ def buildMockCoadd(self, butler, truthCatalog=None, skyMap=None, tract=0):
         tractInfo = skyMap[tract]
         tractWcs = tractInfo.getWcs()
         for patchRef in self.iterPatchRefs(butler, tractInfo):
-            exposure = patchRef.get(self.config.coaddName + "Coadd")
+            # TODO: pybind11 remove `immediate=True` once DM-9112 is resolved
```
r-owen: At risk of being pedantic: there is no actual need to remove `immediate=True` and I suggest alway using it (except for code that actually wants caching, which so far we don't have). That suggest you could delete the comment. On the other hand, I can see an argument for providing a comment for those reading through git history, in which case please consider a slight reword, such as `# immediate=True is required until DM-9112 if fixed`, with or without a `TODO`.

I see at least one other such instance below.

tests/testCoadds.py
```diff
@@ -308,15 +309,15 @@ def testPsfInstallation(self, tract=0):
...
+            savedPsf = coaddExp.getPsf()
             newPsf = lsst.meas.algorithms.CoaddPsf(ccdCat, coaddExp.getWcs())
             self.assertEqual(savedPsf.getComponentCount(), len(ccdCat))
             self.assertEqual(newPsf.getComponentCount(), len(ccdCat))
             for n, record in enumerate(ccdCat):
-                self.assertTrue(lsst.afw.table.io.comparePersistablePtrs(savedPsf.getPsf(n), record.getPsf()))
-                self.assertTrue(lsst.afw.table.io.comparePersistablePtrs(newPsf.getPsf(n), record.getPsf()))
-                self.assertTrue(lsst.afw.table.io.comparePersistablePtrs(savedPsf.getWcs(n), record.getWcs()))
-                self.assertTrue(lsst.afw.table.io.comparePersistablePtrs(newPsf.getWcs(n), record.getWcs()))
+                self.assertIs(savedPsf.getPsf(n), record.getPsf())
```
r-owen: Nice!

python/lsst/pipe/tasks/mocks/mockCoadd.py
kfindeisen: Leaving aside the question of original intent, I think it's good to have tests that use `immediate=False` in order to cover the case where the object is a proxy. Hence why I feel this commit should be treated as a workaround rather than a permanent change.

PR #101, opened 2017-02-20
DM-9520: Remove calibrateTask.py example from Stack (for now) 

mssgill:
In accordance with https://jira.lsstcorp.org/browse/DM-9520, removed
the calibrateTask example which exercises both charImg and calib, and
then removed the relevant part of the docstrings in both these codes
that refer to it.


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -235,30 +235,10 @@ def DebugInfo(name):
...
-
-    Create the task. Note that we're using a custom AstrometryTask (because we don't have a valid
-    astrometric catalogue handy); see \ref calibrate_MyAstrometryTask.
-    @skip CalibrateTask.ConfigClass
-    @until config=config
-
-    We're now ready to process the data. This occurs in two steps, optionally displaying the data after each:
-    - Characterize the image: measure bright sources, fit a background and PSF, and repairs cosmic rays
-    - Calibrate the exposure: measure faint sources, fit an improved WCS and photometric zero-point
+    # Example description used to live here, removed 2-20-2017 by MSSG
```
timj: It's probably better if you refer to the ticket here rather than the person.

mssgill: Good point, i did that (just put in the whole Jira tikt URL, seemed easiest way to get to it if needed later.)

PR #106, opened 2017-03-09
DM-9747 Fix build error in obs_lsstSim

r-owen:


python/lsst/pipe/tasks/scaleZeroPoint.py
```diff
@@ -20,6 +19,8 @@
 # the GNU General Public License along with this program.  If not,
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
+from __future__ import division
```
r-owen: Please use our standard: `from __future__ import absolute_import, division, print_function`

PR #110, opened 2017-03-29
Set the PSF-matched coadd's PSF to the model PSF used

yalsayyad:
MakeCoaddTempExp does not require users to know the dimensions of
the model PSF and automatically adjusts the dimensions to match those of
the Warped PSFs. AssembleCoadd now applies the
dimension-adjusted PSF rather than the user-supplied.

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -619,7 +619,13 @@ discussed in \ref pipeTasks_multiBand (but note that normally, one would use the
             self.inputRecorder.addVisitToCoadd(coaddInputs, tempExp, weight)
         coaddInputs.visits.sort()
         if self.config.doPsfMatch:
-            psf = self.config.modelPsf.apply()
+            # The modelPsf BBox for a psfMatchedWarp/coaddTempExp was dynamically defined by
+            # ModelPsfMatchTask as the square box bounding its spatially-variable, pre-matched WarpedPsf.
+            # Likewise, set the PSF of a PSF-Matched Coadd to the modelPsf
+            # having the maximum width (sufficient because square)
+            modelPsfList = [tempExp.getPsf() for tempExp in tempExpList]
+            modelPsfWidthList = [modelPsf.computeBBox().getWidth() for modelPsf in modelPsfList]
+            psf = modelPsfList[modelPsfWidthList.index(max(modelPsfWidthList))]
         else:
             psf = measAlg.CoaddPsf(coaddInputs.ccds, coaddExposure.getWcs())
         coaddExposure.setPsf(psf)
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -619,7 +619,11 @@ def assembleMetadata(self, coaddExposure, tempExpRefList, weightList):
             self.inputRecorder.addVisitToCoadd(coaddInputs, tempExp, weight)
         coaddInputs.visits.sort()
         if self.config.doPsfMatch:
-            psf = self.config.modelPsf.apply()
+            # set PSF to modelPsf with maximum dimensions
```
laurenam: In addition to stating what the code is doing here, could you add a brief comment as to the "why"?

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -179,16 +179,17 @@ def createTempExp(self, calexpRefList, skyInfo, visitId=0):
                 if numGoodPix > 0 and not didSetMetadata:
                     coaddTempExp.setCalib(exposure.getCalib())
                     coaddTempExp.setFilter(exposure.getFilter())
+                    # PSF replaced after loop with CoaddPsf if and only if creating direct warp
```
laurenam: To avoid confusion (which I was subject to at first!), perhaps rephrase as "PSF replaced with CoaddPsf after loop..."

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -619,7 +619,11 @@ def assembleMetadata(self, coaddExposure, tempExpRefList, weightList):
...
+            modelPsfList = [tempExp.getPsf() for tempExp in tempExpList]
+            modelPsfDimList = [modelPsf.computeBBox().getDimensions() for modelPsf in modelPsfList]
+            maxDim = max(modelPsfDimList)
```
PaulPrice: I'm surprised `max(dimList)` works.  I can't figure out why it does, but I'm not familiar with pybind and maybe I missed something.

python/lsst/pipe/tasks/makeCoaddTempExp.py
yalsayyad: 👍

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: How much do you need here? Would it be sufficient to add explanation on why maximum dimensions in the commit message or would an inline message like the following help? 

            # modelPsf's BBox was dynamically defined by ModelPsfMatchTask
            # to be the square bounding the largest Warped Psf BBox

laurenam: Both would be great!  If the commit message is descriptive enough, the inline can be kept as brief as possible to provide the basic reasoning.

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: Changed to be clearer and more explicit

PR #111, opened 2017-04-05
DM-10096 Add unit test asserts for SpherePoint, SpherePointList and PointList

r-owen:


tests/testSnapCombine.py
```diff
@@ -184,6 +184,7 @@ class MyMemoryTestCase(lsst.utils.tests.MemoryTestCase):
 def setup_module(module):
     lsst.utils.tests.init()
 
+
```
natelust: squash the two pep8 commits together

PR #113, opened 2017-04-13
Update source to new Footprint API

natelust:


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -868,7 +868,7 @@ def countMaskFromFootprint(mask, footprint, bitmask, ignoreMask):
     bbox.clip(mask.getBBox(afwImage.PARENT))
     fp = afwImage.MaskU(bbox)
     subMask = mask.Factory(mask, bbox, afwImage.PARENT)
-    afwDet.setMaskFromFootprint(fp, footprint, bitmask)
+    footprint.spans.setMask(fp, bitmask)
```
fred3m: Huge 👍 on the improved (and more pythonic) API here! I especially like that you can just call ``footprint.spans`` as opposed to ``footprint.getSpans()``.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -830,10 +830,12 @@ def getSkySourceFootprints(self, mergedList, skyInfo, growDetectedFootprints=0):
...
             x = int(numpy.random.uniform(xmin, xmax))
             y = int(numpy.random.uniform(ymin, ymax))
-            foot = afwDetect.Footprint(afwGeom.PointI(x, y), skySourceRadius, patchBBox)
+            spans = afwGeom.SpanSet.fromShape(int(skySourceRadius),
+                                              offset=(x, y))
+            foot = afwDetect.Footprint(spans, patchBBox)
             foot.setPeakSchema(self.merged.getPeakSchema())
 
-            if not foot.overlapsMask(mask):
+            if not foot.spans.overlaps(afwGeom.SpanSet.fromMask(mask)):
```
fred3m: I don't know that I like this change. I think that ``SpanSet.overlaps`` should be overloaded to accept a mask, much like ``SpanSet.intersect``, to prevent the user from having to unnecessarily cast a ``Mask`` to a ``SpanSet``.

fred3m: With that said, in this particular instance it would probably be better to create an object ``maskSpanSet=afwGeom.SpanSet.fromMask(mask)`` before line 827, to avoid recasting it to the exact same object in each iteration of the loop.

PR #119, opened 2017-04-25
Change doc of used field to calib_photometryUsed.  Add reserved and c…

pgee2000:
…andidate flags.

Flag marks sources used in final calibration, name changed from photometryStandard.

Added the ability to reserve a fraction of candidates selected from the initial match.
Mark those calib_photometryReserved.

Mark the remainder of the candidates calib_photometryCandidate.

### DIFF
--- a/python/lsst/pipe/tasks/photoCal.py
+++ b/python/lsst/pipe/tasks/photoCal.py
@@ -27,6 +27,7 @@ from builtins import input
 from builtins import range
 
 import math
+import random
 import sys
 
 import numpy as np
@@ -67,10 +68,16 @@ class PhotoCalConfig(RefMatchConfig):
         default=22.0,
         doc="Don't use objects fainter than this magnitude",
     )
-    doWriteOutput = pexConf.Field(
-        dtype=bool,
-        default=True,
-        doc="Write a field name astrom_usedByPhotoCal to the schema",
+    reserveFraction = pexConf.Field(
+        dtype=float,
+        doc="Fraction of candidates to reserve from fitting; none if <= 0",
+        default=-1.0,
+    )
+    reserveSeed = pexConf.Field(
+        dtype = int,
+        doc = "This number will be multiplied by the exposure ID "
+        "to set the random seed for reserving candidates",
+        default = 1,
     )
     fluxField = pexConf.Field(
         dtype=str,
@@ -290,14 +297,20 @@ into your debug.py file and run photoCalTask.py with the \c --debug flag.
     def __init__(self, refObjLoader, schema=None, **kwds):
         """!Create the photometric calibration task.  See PhotoCalTask.init for documentation
         """
-        RefMatchTask.__init__(self, refObjLoader, schema=schema, **kwds)
+        RefMatchTask.__init__(self, refObjLoader, schema=None, **kwds)
         self.scatterPlot = None
         self.fig = None
-        if self.config.doWriteOutput:
-            self.outputField = schema.addField("photocal_photometricStandard", type="Flag",
-                                               doc="set if source was used in photometric calibration")
+        if schema is not None:
+            self.usedKey = schema.addField("calib_photometryUsed", type="Flag",
+                                           doc="set if source was used in photometric calibration")
+            self.candidateKey = schema.addField("calib_photometryCandidate", type="Flag",
+                                                doc="set if source was a candidate for use in calibration")
+            self.reservedKey = schema.addField("calib_photometryReserved", type="Flag",
+                                               doc="set if source was reserved, so not used in calibration")
         else:
-            self.outputField = None
+            self.usedKey = None
+            self.candidateKey = None
+            self.reservedKey = None
 
     def getSourceKeys(self, schema):
         """!Return a struct containing the source catalog keys for fields used by PhotoCalTask.
@@ -364,6 +377,10 @@ into your debug.py file and run photoCalTask.py with the \c --debug flag.
 
         if len(matches) == 0:
             raise ValueError("No input matches")
+
+        for m in matches:
+            if self.candidateKey is not None:
+                m.second.set(self.candidateKey, True)
 
         # Only use stars for which the flags indicate the photometry is good.
         afterFlagCutInd = [i for i, m in enumerate(matches) if checkSourceFlags(m.second, sourceKeys)]
@@ -456,8 +473,8 @@ into your debug.py file and run photoCalTask.py with the \c --debug flag.
 
         result = []
         for m in matches:
-            if self.outputField is not None:
-                m.second.set(self.outputField, True)
+            if self.usedKey is not None:
+                m.second.set(self.usedKey, True)
             result.append(m)
         return result
 
@@ -577,7 +594,7 @@ into your debug.py file and run photoCalTask.py with the \c --debug flag.
         )
 
     @pipeBase.timeMethod
-    def run(self, exposure, sourceCat):
+    def run(self, exposure, sourceCat, expId=0):
         """!Do photometric calibration - select matches to use and (possibly iteratively) compute
         the zero point.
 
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/photoCal.py
```diff
@@ -294,10 +306,16 @@ def __init__(self, refObjLoader, schema=None, **kwds):
         self.scatterPlot = None
         self.fig = None
         if self.config.doWriteOutput:
-            self.outputField = schema.addField("photocal_photometricStandard", type="Flag",
+            self.usedKey = schema.addField("calib_photometryUsed", type="Flag",
+                                               doc="set if source was used in photometric calibration")
```
parejkoj: wonky indentation

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -648,21 +670,42 @@ def run(self, exposure, sourceCat):
             frame = None
 
         res = self.loadAndMatch(exposure, sourceCat)
+
+        #from res.matches, reserve a fraction of the population and mark the sources as reserved
+        reserveList = []
```
parejkoj: move `reserveList` inside the `if`, since we don't need it unless we're reserving things.

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -648,21 +670,42 @@ def run(self, exposure, sourceCat):
...
+
+        if self.config.reserveFraction > 0:
+            random.seed(self.config.reserveSeed*expId)
+            reserveList = random.sample(res.matches,
+                                        int((self.config.reserveFraction)*len(res.matches)))
+
+            for cand in reserveList:
```
parejkoj: I'd rather that iterator be `candidate` or `reserved`

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -648,21 +670,42 @@ def run(self, exposure, sourceCat):
...
+                res.matches.remove(cand)
+
+            if reserveList and self.reservedKey is not None:
+                for cand in reserveList:
```
parejkoj: Again, `candidate` or `reserved`

python/lsst/pipe/tasks/photoCal.py
pgee2000: Ok

python/lsst/pipe/tasks/photoCal.py
pgee2000: OK

python/lsst/pipe/tasks/photoCal.py
pgee2000: OK

python/lsst/pipe/tasks/photoCal.py
pgee2000: OK

tests/testPhotoCal.py
```diff
@@ -125,6 +124,31 @@ def _runTask(self):
...
+        cat.extend(self.srcCat, mapper=mapper)
+
+        #   test that by default, no stars are reserved and used < candidates
+        task.run(exposure=self.exposure, sourceCat=cat)
+        candidates = cat.get("calib_photometryCandidate").sum()
+        assert(cat.get("calib_photometryReserved").sum() == 0)
+        assert(cat.get("calib_photometryUsed").sum() <= candidates)
+
+        #   set the reserve fraction, and see if the right proportion are reserved.
+        self.config.doWriteOutput = True
```
parejkoj: This usage here strongly suggests that `doWriteOutput` is the wrong name for that option. If you don't want to change it on this ticket, please file another ticket to change it in the future.

tests/testPhotoCal.py
```diff
@@ -125,6 +124,29 @@ def _runTask(self):
...
+        task.run(exposure=self.exposure, sourceCat=cat)
+        candidates = cat.get("calib_photometryCandidate").sum()
+        assert(cat.get("calib_photometryReserved").sum() == 0)
+        assert(cat.get("calib_photometryUsed").sum() <= candidates)
+
+        #   set the reserve fraction, and see if the right proportion are reserved.
+        self.config.reserveFraction = .3
+        task.run(exposure=self.exposure, sourceCat=cat)
+        reserved = cat.get("calib_photometryReserved").sum()
+        assert(reserved == int(.3 * candidates))
```
parejkoj: Please use `self.assertEqual(reserved, int(.3*candidates))`

tests/testPhotoCal.py
```diff
@@ -125,6 +124,29 @@ def _runTask(self):
...
+        assert(cat.get("calib_photometryUsed").sum() <= (candidates - reserved))
```
parejkoj: As above, `self.assertLessEqual(cat.get(...).sum(), (candidates - reserved))`

tests/testPhotoCal.py
```diff
@@ -125,6 +124,29 @@ def _runTask(self):
...
+        mapper = afwTable.SchemaMapper(self.srcCat.schema, schema)
+        cat = afwTable.SourceCatalog(schema)
+        for name in self.srcCat.schema.getNames():
+            mapper.addMapping(self.srcCat.schema.find(name).key)
+        cat.extend(self.srcCat, mapper=mapper)
+
+        #   test that by default, no stars are reserved and used < candidates
+        task.run(exposure=self.exposure, sourceCat=cat)
+        candidates = cat.get("calib_photometryCandidate").sum()
+        assert(cat.get("calib_photometryReserved").sum() == 0)
```
parejkoj: `self.assertEqual(cat.get..., 0)`

tests/testPhotoCal.py
```diff
@@ -125,6 +124,29 @@ def _runTask(self):
...
+        assert(cat.get("calib_photometryUsed").sum() <= candidates)
```
parejkoj: `self.assertLessEqual(cat.get().sum(), candidates)`

parejkoj: Though actually, I'd think you'd be able to know exactly how many objects are selected for `used` here: can't you test with that number, so we can know if it ever changes?

pgee2000: I don't understand how I could know this value except by running the calibration and noting the number.  But that's circular.
Also, I think this is just a test that the flag writing mechanism works.  To test whether the number is correct seems or changes over time seems like it is a different problem.

tests/testPhotoCal.py
pgee2000: OK

tests/testPhotoCal.py
pgee2000: OK

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -290,14 +297,20 @@ def DebugInfo(name):
...
-        if self.config.doWriteOutput:
-            self.outputField = schema.addField("photocal_photometricStandard", type="Flag",
-                                               doc="set if source was used in photometric calibration")
+        if not schema is None:
+            self.usedKey = schema.addField("calib_photometryUsed", type="Flag",
+                                           doc="set if source was used in photometric calibration")
+            self.candidateKey = schema.addField("calib_photometryCandidate", type="Flag",
+                                                doc="set if source was used in photometric calibration")
+            self.reservedKey = schema.addField("calib_photometryReserved", type="Flag",
+                                               doc="set if source was reserved in photometric calibration")
```
pgee2000: OK

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -290,14 +297,20 @@ def DebugInfo(name):
...
         self.scatterPlot = None
         self.fig = None
-        if self.config.doWriteOutput:
-            self.outputField = schema.addField("photocal_photometricStandard", type="Flag",
-                                               doc="set if source was used in photometric calibration")
+        if not schema is None:
+            self.usedKey = schema.addField("calib_photometryUsed", type="Flag",
+                                           doc="set if source was used in photometric calibration")
+            self.candidateKey = schema.addField("calib_photometryCandidate", type="Flag",
+                                                doc="set if source was used in photometric calibration")
```
pgee2000: OK

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -290,14 +297,20 @@ def DebugInfo(name):
...
         """!Create the photometric calibration task.  See PhotoCalTask.init for documentation
         """
-        RefMatchTask.__init__(self, refObjLoader, schema=schema, **kwds)
+        RefMatchTask.__init__(self, refObjLoader, schema=None, **kwds)
         self.scatterPlot = None
         self.fig = None
-        if self.config.doWriteOutput:
-            self.outputField = schema.addField("photocal_photometricStandard", type="Flag",
-                                               doc="set if source was used in photometric calibration")
+        if not schema is None:
```
pgee2000: OK

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -300,13 +300,13 @@ def __init__(self, refObjLoader, schema=None, **kwds):
         RefMatchTask.__init__(self, refObjLoader, schema=None, **kwds)
         self.scatterPlot = None
         self.fig = None
-        if not schema is None:
+        if schema is not None:
             self.usedKey = schema.addField("calib_photometryUsed", type="Flag",
                                            doc="set if source was used in photometric calibration")
             self.candidateKey = schema.addField("calib_photometryCandidate", type="Flag",
-                                                doc="set if source was used in photometric calibration")
+                                                doc="set if source was as possible candidate in photometric calibration")
```
parejkoj: "was a possible"

tests/testPhotoCal.py
parejkoj: Running the test once to get the "current" value and then testing on that value is a reasonable approach, but I agree that it's probably better as a separate test. Probably best done as part of SQUASH, actually, once that's up and running.

PR #121, opened 2017-05-09
DM-10267: Port HSC support for PostgreSQL registries to LSST

PaulPrice:


python/lsst/pipe/tasks/ingestPgsql.py
```diff
@@ -0,0 +1,111 @@
...
+                   }
+
+        cur = conn.cursor()
+        cmd = "CREATE TABLE %s (id SERIAL NOT NULL PRIMARY KEY, " % table
+        cmd += ",".join(["%s %s" % (col, typeMap.get(colType.lower(), 'text')) for
+                         col, colType in self.config.columns.items()])
+        if len(self.config.unique) > 0:
+            cmd += ", UNIQUE(" + ",".join(self.config.unique) + ")"
+        cmd += ")"
+        cur.execute(cmd)
```
n8pease: I thought when using dbapi, it was preferred to let the `cursor.execute` method substitute the values into the string, for protection against Bobby Tables. It looks like you're doing the substitutions above. Are you sure you're safe?

PaulPrice: I had [Little Bobby Tables](http://bobby-tables.com) in my mind as I was working, but it doesn't look like the placeholders work when used for table names or types --- just values.

n8pease: aha, yes. I had that experience too.
I'm not sure what the security requirements are in that case. @timj?

timj: I thought we always used `?` placeholders.

timj: does the db library have a method for cleaning external input?

n8pease: when you call `execute`, and pass a string with placeholders in for `operation` and `parameters` in, the parameters will be cleaned before substituted into the `operation` string.
the placeholder character seems to depend on [paramstyle](https://www.python.org/dev/peps/pep-0249/#paramstyle)(?)

PR #124, opened 2017-05-25
Modify tasks to use lsst.afw.display interface, not ds9

pgee2000:
Also, fix doc in photoCal.py, which gave incorrect debug instructions.


jdswinbank:
Rendered obsolete by DM-13432.

PR #131, opened 2017-06-06
DM-10805: Spatially-varying ZOGY option

djreiss:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -85,6 +86,8 @@ class ImageDifferenceConfig(pexConfig.Config):
     doWriteSubtractedExp = pexConfig.Field(dtype=bool, default=True, doc="Write difference exposure?")
     doWriteMatchedExp = pexConfig.Field(dtype=bool, default=False,
                                         doc="Write warped and PSF-matched template coadd exposure?")
+    doWriteWarpedTemplate = pexConfig.Field(dtype=bool, default=False,
```
yalsayyad: If this config only affects/applies-to/works-with the ZogyImagePsfMatchTask subtask, It should be in the configuration of ZogyImagePsfMatchTask.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -319,232 +346,248 @@ def run(self, sensorRef, templateIdList=None):
...
+            # If doSubtract is False, then subtractedExposure was fetched from disk (above),
+            # thus it may have already been decorrelated. Thus, we do not decorrelate if
+            # doSubtract is False.
+            if self.config.doDecorrelation and self.config.doSubtract:
+                decorrResult = self.decorrelate.run(exposure, templateExposure,
+                                                    subtractedExposure,
+                                                    subtractRes.psfMatchingKernel)
+                subtractedExposure = decorrResult.correctedExposure
+
+        # END (if subtractAlgorithm == 'AL')
```
yalsayyad: This comment is lies. 🤥  :) This is the end of `if self.config.doSubtract`. The end of `(if subtractAlgorithm == 'AL')`  is a few lines up at line 563.  This control flow is probably not what you intended, given that the matchedExp gets written twice if `subtractAlgorithm=="ZOGY"`

PaulPrice: You should never need such comments. If you think that it's helpful, that's probably an indication that your block should be factored out into a separate function.

yalsayyad: imageDifferenceTask has been marked for refactoring for a while now.  There's a separate ticket for that.

python/lsst/pipe/tasks/imageDifference.py
djreiss: Actually, it's just added for debugging purposes, and just used for Zogy right now. But both tasks return a `warpedExposure` item in its return Struct.

But I do agree it would be nice to standardize the returned Struct.

python/lsst/pipe/tasks/imageDifference.py
djreiss: You are right. I will move the comment and remove the extra matchedExp writing code.

PR #133, opened 2017-06-26
Remove use/mention of deprecated PsfAttributes

yalsayyad:


python/lsst/pipe/tasks/snapCombine.py
```diff
@@ -216,7 +216,6 @@ def run(self, snap0, snap1, defects=None):
                 diffKern = diffRet.psfMatchingKernel
                 width, height = diffKern.getDimensions()
                 # TBD...
```
djreiss: Also remove this empty "TBD" comment?

PR #136, opened 2017-07-10
DM-11243: Refine PSF-based image selection

PaulPrice:


python/lsst/pipe/tasks/selectImages.py
```diff
@@ -309,11 +311,11 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
             scatterSize = sigmaMad(starSize - psfSize)
 
             valid = True
-            if medianE1 > self.config.maxEllipResidual:
+            if self.config.maxEllipResidual and medianE1 > self.config.maxEllipResidual:
                 self.log.info("Removing visit %s because median e residual too large: %f" %
```
laurenam: Were you able to confirm that `medianE1` is the correct comparison variable here?  All the plots I've seen from Takashi Hamana have `medianE` in the label (but that's not a guarantee of what is actually plotted), but I also note that @rearmstr returns `medianE` in his sample script here:
http://jeeves.astro.princeton.edu/pipermail/hsc_software/all/5853.html

I also wonder if it's worth adding the config value used in the log.info messages (i.e. print explicitly what "too large" is)?

PaulPrice: I'm fairly convinced @rearmstr meant `medianE` instead of `medianE1`, and so I've added a commit fixing that up (see the commit message for my reasoning).
I've also added a commit to include the limit when printing the value.

laurenam:
You may also want to update the criteria (now at line 278).  It's not very specific, so not explicitly wrong as is (so I won't insist!), but clarification wouldn't hurt.

PR #141, opened 2017-07-25
Remove config option to incorrectly `matchThenWarp`

yalsayyad:
in WarpAndPsfMatchTask when generating PSF-matched warps.
Allowing this legacy order of operations is no longer necessary.


yalsayyad:
This PR is much simpler than the diff would have you believe.   All I did was remove the `if` branch of an `if-else` block.

PR #142, opened 2017-08-07
Allow configuring more statistical options for assembleCoadd

yalsayyad:
Specify main stacking statistic in AssembleCoadd and derived tasks
with config parameter `statistic`. For backwards compatibility:
* Setting the deprecated doSigmaClip to True (now False by default)
  ignores `statistic` and uses MEANCLIP.
* The default statistic is MEANCLIP imitating the previous default
  of doSigmaClip=True.

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -34,6 +34,7 @@ import lsst.afw.detection as afwDet
 import lsst.coadd.utils as coaddUtils
 import lsst.pipe.base as pipeBase
 import lsst.meas.algorithms as measAlg
+import lsst.log as log
 from .coaddBase import CoaddBaseTask, SelectDataIdContainer
 from .interpImage import InterpImageTask
 from .matchBackgrounds import MatchBackgroundsTask
@@ -57,19 +58,24 @@ class AssembleCoaddConfig(CoaddBaseTask.ConfigClass):
         length=2,
         default=(2000, 2000),
     )
+    statistic = pexConfig.Field(
+        dtype=str,
+        doc="Main stacking statistic for aggregating over the epochs.",
+        default="MEANCLIP",
+    )
     doSigmaClip = pexConfig.Field(
         dtype=bool,
-        doc="Perform sigma clipped outlier rejection? If False then compute a simple mean.",
-        default=True,
+        doc="Perform sigma clipped outlier rejection with MEANCLIP statistic? (DEPRECATED)",
+        default=False,
     )
     sigmaClip = pexConfig.Field(
         dtype=float,
-        doc="Sigma for outlier rejection; ignored if doSigmaClip false.",
+        doc="Sigma for outlier rejection; ignored if non-clipping statistic selected.",
         default=3.0,
     )
     clipIter = pexConfig.Field(
         dtype=int,
-        doc="Number of iterations of outlier rejection; ignored if doSigmaClip false.",
+        doc="Number of iterations of outlier rejection; ignored if non-clipping statistic selected.",
         default=2,
     )
     scaleZeroPoint = pexConfig.ConfigurableField(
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -57,19 +58,39 @@ class AssembleCoaddConfig(CoaddBaseTask.ConfigClass):
...
+            "VARIANCE": "variance",
+            "IQRANGE": "inter-quartile range",
+            "STDEVCLIP": "N-sigma clipped stdev (N set by sigmaClip)",
+            "VARIANCECLIP": "N-sigma clipped variance (N set by sigmaClip)",
+            "MIN": "minimum",
+            "MAX": "maximum",
+            "SUM": "sum",
+            "MEANSQUARE": "mean of square of  pixel values",
+        }
+    )
```
RobertLuptonTheGood: Why do we need to specify these?  There's code in Statistics that takes a string and converts it to a valid enum;  why don't we use that in the validate method?

Listing statistics here is what caused this problem in the first place.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -555,16 +580,15 @@ def assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgInfoL
         Assemble the coadd using the provided list of coaddTempExps. Since the full coadd covers a patch (a
         large area), the assembly is performed over small areas on the image at a time in order to
         conserve memory usage. Iterate over subregions within the outer bbox of the patch using
-        \ref assembleSubregion to mean-stack the corresponding subregions from the coaddTempExps (with outlier
-        rejection if config.doSigmaClip=True). Set the edge bits the the coadd mask based on the weight map.
+        \ref assembleSubregion to stack the corresponding subregions from the coaddTempExps with the
+        statistic specified. Set the edge bits the the coadd mask based on the weight map.
```
rearmstr: "the the"  - I know it was there before, but it seems like you can fix it also.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1170,13 +1205,21 @@ def buildDifferenceImage(self, skyInfo, tempExpRefList, imageScalerList, weightL
...
+        config = AssembleCoaddConfig()
+        # getattr necessary because subtasks do not survive Config.toDict()
+        configIntersection = {k: getattr(self.config, k)
+                              for k, v in self.config.toDict().items() if (k in config.keys())}
+        config.update(**configIntersection)
+
         # Build the unclipped coadd
-        coaddMean = AssembleCoaddTask.assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList,
-                                               bgModelList, doClip=False)
+        config.statistic = 'MEAN'
```
rearmstr: Haven't you already forced this to be 'MEAN'?  Maybe it is just here for clarity?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -667,7 +691,7 @@ def assembleSubregion(self, coaddExposure, bbox, tempExpRefList, imageScalerList
         matching is enabled, add the background and background variance from each coaddTempExp. Remove mask
         planes listed in config.removeMaskPlanes, Finally, mean-stack
         the actual exposures using \ref afwMath.statisticsStack "statisticsStack" with outlier rejection if
-        config.doSigmaClip=True. Assign the stacked subregion back to the coadd.
+        config.statistics=MEANCLIP. Assign the stacked subregion back to the coadd.
```
rearmstr: Since you have now added more statistics options, I think you need to modify this doc.  It assumes you are doing mean or clipping.

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: Yeah, writing out the allowed options here really sucks, doesn't it.  I liked the idea of the options being exposed in the resulting config files.  What I really want is the stats class to give me a dictionary of available statistics for stacking. I talked this over with @pschella, who says the new API will offer that.  I ended up specifying them explicitly because:
1) Not all statistics area available for stacking: NOTHING, ERROR, ORMASK make no sense in a coadd and stackStatistics raises InvalidParameter errors or segfault (in the case of ORMASK)
2) The only documentation for the options in the `Property enum`  is in code comments. There's no way to retrieve them.

Without exposing the options, I could add to the config's `validate`: 
```
        UNSTACKABLE_STATS = ['NOTHING', 'ERROR', 'ORMASK']
        if not hasattr(afwMath.Property, self.statistic) or self.statistic in UNSTACKABLE_STATS:
            stackableStats = [str(k) for k in afwMath.Property.__members__.keys() 
                         if str(k) not in UNSTACKABLE_STATS]
            raise ValueError("statistic %s is not allowed. Please choose one of %s."
                             % (self.statistic, stackableStats))
```
sufficient solution?

yalsayyad: It'll print out: 

> `ValueError: statistic ORMASK is not allowed. Please choose one of ['NPOINT', 'ERRORS', 'MIN', 'STDEVCLIP', 'SUM', 'VARIANCECLIP', 'MEDIAN', 'MEANCLIP', 'IQRANGE', 'STDEV', 'VARIANCE', 'MEANSQUARE', 'MAX', 'MEAN'].`

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: It was for clarity and safety in case the validation of statistic==MEAN was removed.  I'll add a comment. 

Question: How do you feel about forcing the third and final iteration of `assemble` to use a MEAN? I added that to replicate current behavior.  Of course MEAN should be the default, but it doesn't technically need to be enforced anymore.

python/lsst/pipe/tasks/assembleCoadd.py
RobertLuptonTheGood: That looks fine

PR #144, opened 2017-08-15
DM-11589: Fix pytest FAKE failure

timj:
Here is an attempt at fixing the pytest failure with fake mask planes. I remove the FAKE entry in the test  for `testFakeProcessing.py` in the same manner as used in `testFakes.py`. No idea whether this is reasonable but it does work.

Thanks @danielsf for the kick.

### DIFF
--- a/tests/testFakeProcessing.py
+++ b/tests/testFakeProcessing.py
@@ -125,6 +125,8 @@ class FakeProcessingTestCase(lsst.utils.tests.TestCase):
             else "{}-ProcessCcd".format(OutputName)
         dataId = dict(visit=1)
         dataIdStrList = ["{}={}".format(*item) for item in dataId.items()]
+        mask = None
+        maskPlaneName = "FAKE"
 
         try:
             # Set the configurations for running the fake object piepline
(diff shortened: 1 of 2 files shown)


tests/testFakeProcessing.py
```diff
@@ -116,6 +116,7 @@ def getObsTestConfig(TaskClass):
 
 
 class FakeProcessingTestCase(lsst.utils.tests.TestCase):
+
```
PaulPrice: Separate commit for whitespace change.

tests/testFakeProcessing.py
```diff
@@ -145,12 +148,15 @@ def testFakeProcessing(self):
 
             exposure = pCcdResult.resultList[0].result.calibRes.exposure
             mask = exposure.getMaskedImage().getMask()
-            maskBit = 2**mask.getMaskPlaneDict()['FAKE']
+            maskBit = 2**mask.getMaskPlaneDict()[maskPlaneName]
```
PaulPrice: Not your fault, but better would be:

    maskBit = mask.getPlaneBitMask(maskPlaneName)

timj:
Ok. I see #143 has the test renames in so I'll remove my renaming. @r-owen you should rebase your branch to pick up the test fix.

tests/testFakeProcessing.py
timj: I can't get that to work so I think I'm going to punt on that fix for the moment.

r-owen:
OK Thanks. I though I marked pipe_tasks as mine to rename?

PR #145, opened 2017-08-19
Force testCoadds to run in separate pytest process

timj:
This test creates and configures a single shared butler repository. It takes 200MB and 25 seconds to create it and for the tests to run in parallel each sub process would need that amount of disk space in a temporary directory. This is possible but is much slower than forced single-process execution.

### DIFF
--- a/tests/testCoadds.py
+++ b/tests/testCoadds.py
@@ -39,6 +39,13 @@ the REUSE_DATAREPO variable below to False, as sometimes this error message indi
 different problem that's revealed when we're not trying to cache the mock data between
 tests (but set REUSE_DATAREPO back to True when done debugging, or this test will be very
 slow).
+
+NOTE: This test should not be run with other tests using pytest, and should
+not be discoverable automatically by pytest. The reason for this is that the
+tests rely on 200 MB of data generated on module load, using a single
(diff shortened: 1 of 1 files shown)


jonathansick:
@timj Following up from Slack:

> when the tests get renamed we have to remember not to rename testCoadds.py (you don’t want that one to be discovered automatically)

Could you create a README in `tests/` and/or a module docstring in testCoadds.py that states this?

tests/testCoadds.py
```diff
@@ -39,6 +39,12 @@
 different problem that's revealed when we're not trying to cache the mock data between
 tests (but set REUSE_DATAREPO back to True when done debugging, or this test will be very
 slow).
+
+NOTE: This test should not be run with other tests using pytest, and should
+not be discoverable automatically by pytest.
+The reason for this is that the tests rely on 200 MB of data generated on module load, using a single directory visible to all the tests.
+When run in parallel with pytest-xdist this data will be created for ever sub-process, leading to excessive disk usage and excessive test execution times.
```
jdswinbank: "every".

timj:
Note this can't be merged until lsst/sconsUtils#34 is merged.

PR #146, opened 2017-08-22
Update schema naming conventions to avoid bug in peak culling.

TallJimbo:
Also adds an assert to hopefully keep that bug from happening again.


TallJimbo:
Assert message added.

PR #148, opened 2017-08-31
Implement CompareWarpAssembleCoaddTask with artifact rejection

yalsayyad:
* Add new task to compare PSF-matched
warps and identify and mask artifacts when coadding.
* Control warpType by config parameter `warpType`

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -41,7 +41,7 @@ from .scaleZeroPoint import ScaleZeroPointTask
 from .coaddHelpers import groupPatchExposures, getGroupDataRef
 from lsst.meas.algorithms import SourceDetectionTask
 
-__all__ = ["AssembleCoaddTask", "SafeClipAssembleCoaddTask"]
+__all__ = ["AssembleCoaddTask", "SafeClipAssembleCoaddTask", "CompareWarpAssembleCoaddTask"]
 
 
 class AssembleCoaddConfig(CoaddBaseTask.ConfigClass):
@@ -362,9 +362,11 @@ discussed in \ref pipeTasks_multiBand (but note that normally, one would use the
                 self.log.warn("No valid background models")
                 return
 
+        supplementaryData = self.makeSupplementaryData(dataRef, selectDataList)
         coaddExp = self.assemble(skyInfo, inputData.tempExpRefList, inputData.imageScalerList,
                                  inputData.weightList,
-                                 inputData.backgroundInfoList if self.config.doMatchBackgrounds else None)
+                                 inputData.backgroundInfoList if self.config.doMatchBackgrounds else None,
+                                 supplementaryData=supplementaryData)
         if self.config.doMatchBackgrounds:
             self.addBackgroundMatchingMetadata(coaddExp, inputData.tempExpRefList,
                                                inputData.backgroundInfoList)
@@ -384,6 +386,16 @@ discussed in \ref pipeTasks_multiBand (but note that normally, one would use the
             dataRef.put(coaddExp, self.getCoaddDatasetName(self.warpType))
 
         return pipeBase.Struct(coaddExposure=coaddExp)
+
+    def makeSupplementaryData(self, dataRef, selectDataList):
+        """!
+        \brief Make additional inputs to assemble() specific to subclasses.
+
+        Available to be implemented by subclasses only if they need the
+        coadd dataRef for performing preliminary processing before
+        assembling the coadd.
+        """
+        pass
 
     def getTempExpRefList(self, patchRef, calExpRefList):
         """!
@@ -564,7 +576,7 @@ discussed in \ref pipeTasks_multiBand (but note that normally, one would use the
                                imageScalerList=newScaleList, backgroundInfoList=newBackgroundStructList)
 
     def assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgInfoList=None,
-                 altMaskList=None, mask=None):
+                 altMaskList=None, mask=None, supplementaryData=None):
         """!
         \anchor AssembleCoaddTask.assemble_
 
@@ -583,6 +595,8 @@ discussed in \ref pipeTasks_multiBand (but note that normally, one would use the
         \param[in] bgInfoList: List of background data from background matching, or None
         \param[in] altMaskList: List of alternate masks to use rather than those stored with tempExp, or None
         \param[in] mask: Mask to ignore when coadding
+        \param[in] supplementaryData: pipeBase.Struct with additional data products needed to assemble coadd.
+                        Only used by subclasses that implement makeSupplementaryData and override assemble.
         \return coadded exposure
         """
         tempExpName = self.getTempExpDatasetName(self.warpType)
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -383,6 +386,14 @@ def run(self, dataRef, selectDataList=[]):
 
         return pipeBase.Struct(coaddExposure=coaddExp)
 
+    def makeSupplementaryData(self, dataRef, selectDataList):
+        """!
+        \brief Make additional inputs to assemble() specific to subclasses. 
+
+        To be implemented by subclasses if needed.
```
isullivan: Could you give an example of the sort of supplementary data that could be provided, or that might be needed?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -42,7 +42,7 @@
 from .coaddHelpers import groupPatchExposures, getGroupDataRef
```
isullivan: flake8 complains that afwDetect is imported but not used.

python/lsst/pipe/tasks/assembleCoadd.py
isullivan: (Picky) Watch for trailing whitespace (line 391)

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -562,7 +573,7 @@ def backgroundMatching(self, inputData, refExpDataRef=None, refImageScaler=None)
                                imageScalerList=newScaleList, backgroundInfoList=newBackgroundStructList)
 
     def assemble(self, skyInfo, tempExpRefList, imageScalerList, weightList, bgInfoList=None,
-                 altMaskList=None, mask=None):
+                 altMaskList=None, mask=None, supplementaryData=None):
```
isullivan: Add `supplementaryData` to the documented `param[in]`'s

isullivan: Also, it's not actually used here. Is it only used if you retarget `assemble`?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+
+
+class CompareWarpAssembleCoaddConfig(AssembleCoaddConfig):
+    assembleStaticSkyModel = pexConfig.ConfigurableField(
+        target=AssembleCoaddTask,
+        doc="Task to assemble clean, artifact-free, PSF-matched Coadd to serve as model of "
+            "the static sky",
+    )
+    chiThreshold = pexConfig.Field(
+        doc="Pixels in the chi-image of warp minus the model of the static sky (template coadd)",
```
isullivan: I don't really know what this means. Could you re-word or add to this description?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+    In this task, we correct for this behavior by creating a new badMaskPlane 'CLIPPED' which marks
+    pixels in the individual warps suspected to contain an artifact.
+    We populate this plane on the input warps by comparing PSF-matched warps with a PSF-matched median coadd
+    which serves as a model of the static sky. Any group of pixels that deviates from the PSF-matched
+    median coadd by more than config.chiThreshold sigma, is an artifact candidate. The candidates are then
+    filtered to remove variable sources and sources that are difficult to subtract such as bright stars.
+    This filter is configured using the config parameters temporalThreshold and spatialThreshold.
+    The temporalThreshold is the maximum fraction of epochs that the deviation can
+    appear in and still be considered an artifact. The spatialThreshold is the maximum fraction of pixels in
+    the footprint of the deviation that appear other epochs (where other epochs is defined by the
```
isullivan: "...appear [in] other epochs..." ?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+        """
+        spanSetMaskList = maskSpanSets.artifacts
+        spanSetNoDataList = maskSpanSets.noData
+        altMaskList = []
+        for warpRef, artifacts, noData in zip(tempExpRefList, spanSetMaskList, spanSetNoDataList):
+            warp = warpRef.get(self.getTempExpDatasetName(self.config.warpType), immediate=True)
+            mask = warp.maskedImage.mask
+            maskClipValue = mask.addMaskPlane("CLIPPED")
+            noDataValue = mask.addMaskPlane("NO_DATA")
+            _ = [s.clippedTo(mask.getBBox()).setMask(mask, 2**maskClipValue) for s in artifacts]
```
isullivan: I think a plain `for` loop would make more sense here than assigning a list comprehension to a dummy variable. Same for the next line.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+        return altMaskList
+
+    def _filterArtifacts(self, spanSetList, epochCountImage, maxNumEpochs=None, minPixels=None):
+        if minPixels is None:
+            minPixels = self.config.minPixels
+        maskSpanSetList = []
+        x0, y0 = epochCountImage.getXY0()
+        for i, span in enumerate(spanSetList):
+            y, x = span.indices()
+            if len(y) < self.config.minPixels:
```
isullivan: Shouldn't this just be `minPixels`?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+            spanSetNoDataMaskList.append(spanSetNoDataMask)
+        return pipeBase.Struct(artifacts=spanSetArtifactList,
+                               noData=spanSetNoDataMaskList)
+
+    def computeAltMaskList(self, tempExpRefList, maskSpanSets):
+        """!
+        \brief Apply artifact span set lists to masks
+
+        @param tempExpRefList: List of data references to warps
+        @param imageScalerList: List of image scalers
```
isullivan: Please update the docstring with `maskSpanSets`, and not `imageScalarList`

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+            if self.config.doMaskNegative:
+                arr[numpy.where(arrIn <= -self.config.chiThreshold)] = 1
+        return arr
+
+    def _snrToBinaryIm(self, im):
+        arr = self._snrToBinaryArr(im.array)
+        return afwImage.ImageU(arr.astype(numpy.uint16), xy0=im.getXY0())
+
+    def _makeChiIm(self, maskedImage):
+        chiIm = maskedImage.image.Factory(1./numpy.sqrt(maskedImage.variance.array))
```
isullivan: Do you need to worry about invalid variance values being present (<=0 or nan)?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+        chiIm *= maskedImage.image
+        chiIm.setXY0(maskedImage.getXY0())
+        return chiIm
+
+    def _readAndComputeWarpDiff(self, warpRef, imageScaler, templateCoadd):
+        warp = warpRef.get(self.getTempExpDatasetName('psfMatched'), immediate=True)
```
isullivan: Should the warp type be read from the config?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1362,3 +1374,327 @@ def detectClipBig(self, tempExpClipList, clipFootprints, clipIndices, maskClipVa
...
+            mi = self._readAndComputeWarpDiff(warpRef, imageScaler, templateCoadd)
+            chiIm = self._makeChiIm(mi)
+            chiOneMap = self._snrToBinaryIm(chiIm)
+            epochCountImage += chiOneMap
+
+        self.log.debug("Generating Mask List. Second loop through warps")
+        spanSetArtifactList = []
+        spanSetNoDataMaskList = []
+
+        maxNumEpochs = max(1, self.config.temporalThreshold*len(tempExpRefList))
```
isullivan: This should maybe be rounded to the nearest integer.

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: Yes thats right. If you're writing a new assembleCoadd subclass that needs extra input data that can be computed with the coadd dataRef, you'd override makeSupplementaryData().   In CompareWarpAssemble the template can eventually just be read off disk in  makeSupplementaryData when we have the capability to generate many different coadds, but for now we need to generate a PSF-Matched MEDIAN coadd on the fly. We can't reuse the results of prepareInputs from the main task (Like SafeClip) because those are for the direct warps (and don't apply to the psf-matched warps).  So makeSupplementaryData generates it by running the subtask.

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: No, but I added a comment
# Warp comparison must use PSF-Matched Warps regardless of requested coadd warp type

python/lsst/pipe/tasks/assembleCoadd.py
isullivan: No, that seems fine.

RobertLuptonTheGood: I think that that's standard IEEE754 and can be relied upon.

PR #150, opened 2017-09-19
DM-11951: Fixes in AL Decorrelation and Zogy

djreiss:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -323,7 +323,8 @@ def run(self, sensorRef, templateIdList=None):
                 spatiallyVarying = self.config.doSpatiallyVarying
                 subtractRes = self.subtract.subtractExposures(templateExposure, exposure,
                                                               doWarping=True,
-                                                              spatiallyVarying=spatiallyVarying)
+                                                              spatiallyVarying=spatiallyVarying,
```
r-owen: For simplicity of code, readability and consistency with the new next line, I gently suggest you use `spatiallyVarying=self.config.doSpatiallyVarying,` here and ditch line 323. This same thing appears once more in the code and would benefit from the same treatment.

PR #153, opened 2017-10-24
DM-12108: Add fake sources after wcs update

rearmstr:
If adding fake sources are to be consistent in every frame that
goes into a coadd, they must be added after the wcs has been
updated from that inserted by the telescope. This necessitates that
detection and measurement be rerun with the fake sources inserted

### DIFF
--- a/python/lsst/pipe/tasks/calibrate.py
+++ b/python/lsst/pipe/tasks/calibrate.py
@@ -487,9 +487,6 @@ class CalibrateTask(pipeBase.CmdLineTask):
         table = SourceTable.make(self.schema, sourceIdFactory)
         table.setMetadata(self.algMetadata)
 
-        if self.config.doInsertFakes:
-            self.insertFakes.run(exposure, background=background)
-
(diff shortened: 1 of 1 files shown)


rearmstr:
The commit message says that this change is important for getting coadds to be consistent.  Isn't it also true that you need this change even if you are only running a single image?  The coordinates must be incorrect for the fake sources if you are not using the correct wcs.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -547,6 +544,38 @@ def calibrate(self, exposure, exposureIdInfo=None, background=None,
...
+                )
+            self.catalogCalculation.run(sourceCat)
+
+            if icSourceCat is not None and \
+                    len(self.config.icSourceFieldsToCopy) > 0:
+                self.copyIcSourceFields(icSourceCat=icSourceCat,
+                                        sourceCat=sourceCat)
+
+
+
```
rearmstr: You can remove at least one of these blank lines.

PR #158, opened 2017-11-06
DM-12450: use --reuse-outputs-from instead of doOverwrite options

TallJimbo:
Note that `makeCoaddTempExp` is the only CmdLineTask in this package that actually had an overwrite/clobber option prior to these changes (the rest are all in pipe_drivers), and that's why it's getting special treatment.  While it might be cleanest to have all of these options handled at the same level, I don't want to restructure the code to achieve that now.

### DIFF
--- a/python/lsst/pipe/tasks/makeCoaddTempExp.py
+++ b/python/lsst/pipe/tasks/makeCoaddTempExp.py
@@ -117,8 +117,8 @@ class MakeCoaddTempExpTask(CoaddBaseTask):
     @copydoc \_\_init\_\_
 
     This task has one special keyword argument: passing reuse=True will cause
-    the task to skip the creation of warps that are already present in input
-    or output the repositories.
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -121,7 +116,9 @@ class MakeCoaddTempExpTask(CoaddBaseTask):
 
     @copydoc \_\_init\_\_
 
-    This task has no special keyword arguments.
+    This task has one special keyword argument: passing reuse=True will cause
+    the task to skip the creation of warps that are already present in input
+    or output the repositories.
```
hsinfang: Later you changed to `datasetExists(..., write=True)` so it will only search the output repos?

TallJimbo: Good point; I'll amend that commit to update the comment as well.

PR #159, opened 2017-11-07
Dilate artifact detections before filtering them

yalsayyad:
Detections should be grown before deciding
whether they are artifacts or variable sources.

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -1698,7 +1698,9 @@ class CompareWarpAssembleCoaddTask(AssembleCoaddTask):
         """
 
         self.log.debug("Generating Count Image, and mask lists.")
-        epochCountImage = afwImage.ImageU(templateCoadd.getBBox())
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1708,12 +1710,18 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
...
+                                                afwGeom.Stencil.CIRCLE).clippedTo(coaddBBox)
+                                      for s in spanSetList]
+                slateIm.set(0)
+                for spans in dilatedSpanSetList:
+                    spans.setImage(slateIm, 1, doClip=True)
+                epochCountImage += slateIm
             else:
-                chiIm = afwImage.ImageF(templateCoadd.getBBox(), numpy.nan)
+                chiIm = afwImage.ImageF(coaddBBox, numpy.nan)
                 spanSetList = []
```
natelust: I think this needs to be renamed to dilatedSpanSetList, otherwise there is a chance below that spanSetArtifactList.append will fail, as spanSetList will not exist.

PR #162, opened 2017-11-16
DM-12445: Set appropriate default configs for CompareWarp Coadds

yalsayyad:


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1446,14 +1446,13 @@ class CompareWarpAssembleCoaddConfig(AssembleCoaddConfig):
...
     )
-    temporalThreshold = pexConfig.Field(
-        doc="Unitless fraction of number of epochs to classify as an artifact/outlier source versus"
-            " a source that is intrinsically variable or difficult to subtract cleanly. "
-            "If outlier region in warp-diff Chi-image is mostly (defined by spatialThreshold) "
-            "an outlier in less than temporalThreshold * number of epochs, then mask. "
-            "Otherwise, do not mask.",
-        dtype=float,
-        default=0.075
+    maxNumEpochs = pexConfig.Field(
```
PaulPrice: I would have thought that a fraction would be so much more generally useful than a hard number. I'm worried that a hard number isn't going to scale as you go from n=4 to n=400 inputs.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1769,25 +1766,25 @@ def computeAltMaskList(self, tempExpRefList, maskSpanSets):
 
         return altMaskList
 
-    def _filterArtifacts(self, spanSetList, epochCountImage, maxNumEpochs=None):
+    def _filterArtifacts(self, spanSetList, epochCountImage):
```
PaulPrice: This method needs a docstring.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1769,25 +1766,25 @@ def computeAltMaskList(self, tempExpRefList, maskSpanSets):
...
             y, x = span.indices()
             counts = epochCountImage.array[[y1 - y0 for y1 in y], [x1 - x0 for x1 in x]]
-            idx = numpy.where((counts > 0) & (counts <= maxNumEpochs))
+            idx = numpy.where((counts > 0) & (counts <= self.config.maxNumEpochs))
             percentBelowThreshold = len(idx[0]) / len(counts)
             if percentBelowThreshold > self.config.spatialThreshold:
                 maskSpanSetList.append(span)
         return maskSpanSetList
 
     def _readAndComputeWarpDiff(self, warpRef, imageScaler, templateCoadd):
```
PaulPrice: This method needs a docstring.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1769,25 +1766,25 @@ def computeAltMaskList(self, tempExpRefList, maskSpanSets):
...
 
-    def _filterArtifacts(self, spanSetList, epochCountImage, maxNumEpochs=None):
+    def _filterArtifacts(self, spanSetList, epochCountImage):
         maskSpanSetList = []
         x0, y0 = epochCountImage.getXY0()
         for i, span in enumerate(spanSetList):
             y, x = span.indices()
             counts = epochCountImage.array[[y1 - y0 for y1 in y], [x1 - x0 for x1 in x]]
-            idx = numpy.where((counts > 0) & (counts <= maxNumEpochs))
+            idx = numpy.where((counts > 0) & (counts <= self.config.maxNumEpochs))
```
PaulPrice: You should avoid using `numpy.where` whenever you can (it's expensive), and just use a boolean array. Then, `len(idx[0])` below simply becomes `array.sum()`.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1769,25 +1766,25 @@ def computeAltMaskList(self, tempExpRefList, maskSpanSets):
...
             percentBelowThreshold = len(idx[0]) / len(counts)
             if percentBelowThreshold > self.config.spatialThreshold:
                 maskSpanSetList.append(span)
         return maskSpanSetList
 
     def _readAndComputeWarpDiff(self, warpRef, imageScaler, templateCoadd):
-        # Warp comparison must use PSF-Matched Warps regardless of requested coadd warp type
-        warpName = self.getTempExpDatasetName('psfMatched')
+        # Warp comparison should use PSF-Matched Warps regardless of requested coadd warp type
+        warpName = self.getTempExpDatasetName(self.config.assembleStaticSkyModel.warpType)
```
PaulPrice: It's not a good idea to go reaching through multiple config elements to get what you want: what if I `retarget` the `assembleStaticSkyModel` and name `warpType` something else? You should only access what you control. That suggests putting `warpType` into the config at _this_ level, and passing it down to the `assembleStaticSkyModel` as it needs it.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1454,19 +1454,22 @@ class CompareWarpAssembleCoaddConfig(AssembleCoaddConfig):
...
         default=2
     )
-    spatialThreshold = pexConfig.Field(
+    spatialThreshold = pexConfig.RangeField(
         doc="Unitless fraction of pixels defining how much of the outlier region has to meet the "
             "temporal criteria",
         dtype=float,
-        default=0.5
+        default=0.5,
+        min=0., max=1.
```
PaulPrice: Add `inclusiveMax=True`?

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1454,19 +1454,22 @@ class CompareWarpAssembleCoaddConfig(AssembleCoaddConfig):
...
     )
 
     def setDefaults(self):
         AssembleCoaddConfig.setDefaults(self)
+        self.statistic = 'MEAN'
         self.assembleStaticSkyModel.warpType = 'psfMatched'
-        self.assembleStaticSkyModel.statistic = 'MEDIAN'
+        self.assembleStaticSkyModel.statistic = 'MEANCLIP'
+        self.assembleStaticSkyModel.sigmaClip = 1.5
```
PaulPrice: Why such harsh clipping?

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: Just made a new ticket for this: DM-12692. I had talked briefly with @RobertLuptonTheGood about it. You're right maxNumEpochs won't scale, but neither will a fraction of the maximum number of epochs.  I've seen some steep Nepoch gradients in a patch.  Its also depends on how many repeat pointings line up the artifacts.

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: Hmmm. How would you have done this?  Passed the new config down to the subtask in `validate`?  I configure the defaults for assembleStaticSkyModel subtask (an `AssembleCoaddTask` instance) in the CompareWarpAssembleConfig's `setDefaults`, already requiring that it have a warpType. It's going to be 'psfMatched' in every scenario I can think of (but then again I lack imagination).  I'm debating if its worth making a new config parameter or just leaving it hardcoded.

PaulPrice: We could just leave it hardcoded for now: I also lack the imagination to think as to why anyone would use a different `warpType`, since the algorithm for rejecting artifacts requires PSF-matched inputs.
Or put `warpType` into the `AssembleCoaddTask`'s `Config`, and then pass it to the `AssembleStaticSkyModelTask` constructor when you `makeSubtask`.

python/lsst/pipe/tasks/assembleCoadd.py
PaulPrice: Yeah, it probably needs some combination of a number and a fraction, maybe `max(maxNumEpochs, fracMaxEpochs*num)`?

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: Borrowed from SafeClip. Running a grid of configs is still a to do: DM-12698

PaulPrice: But harsh clipping like this is part of the SafeClip algorithm (we deliberately toss out anything that looks even remotely bad in order to make a very clean model of the sky), whereas I wouldn't have thought it would be for this.

yalsayyad: This is the same idea.  Make a very very clean and naive robust PSF-match coadd to serve as the naive model of the static sky.

PaulPrice: Ah, I see --- this is for the `assembleStaticSkyModel`. Fair enough, then.

yalsayyad: For 2< N<=5 especially, we have to be strict to get the junk out of the naive template:

```
import  lsst.afw.math as afwMath
arr = [1,10,10,20]
sctrl = afwMath.StatisticsControl()
sctrl.setNumSigmaClip(1.5)
sctrl.setNumIter(3)
print afwMath.makeStatistics(arr, afwMath.MEANCLIP, sctrl).getValue()
print afwMath.makeStatistics(arr, afwMath.STDEV, sctrl).getValue()
> 10.0
> 7.76208734813
```

PaulPrice: Perhaps that's due to a shortcoming in our statistics suite (no robust stdev), or because you're not using `STDEVCLIP`:
```
>>> ss = afwMath.makeStatistics(arr, afwMath.MEANCLIP | afwMath.STDEVCLIP | afwMath.NCLIPPED, sctrl)
>>> ss.getValue(afwMath.MEANCLIP)
10.0
>>> ss.getValue(afwMath.STDEVCLIP)
0.0
>>> ss.getValue(afwMath.NCLIPPED)
2.0
```

(`NCLIPPED` is a feature being implemented on DM-9953.)

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: I thought configs frozen by the time we got to making the subtask, but I'll give it try.

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: That's cool.  My point was that if I increase `sctrl.setNumSigmaClip(1.5)` to  `sctrl.setNumSigmaClip(2.6)` the 2 outliers don't get clipped anymore.
```
>>> sctrl.setNumSigmaClip(2.6)
>>> print afwMath.makeStatistics(arr, afwMath.MEANCLIP, sctrl).getValue()
10.25
```

PaulPrice: OK. How common is it that 40% of the inputs have outliers?

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: The config is definitely frozen within `CompareWarpAssembleCoadd.__init__`

```
(Pdb) kwargs['config'].assembleStaticSkyModel.warpType = staticSkyModelWarpType
*** FieldValidationError: ConfigurableField 'assembleStaticSkyModel' failed validation: Cannot modify a frozen Config
For more information read the Field definition at:
  File pex/config/configurableField.py:271 (__deepcopy__)And the Config definition at:
  File pipe/tasks/assembleCoadd.py:1439 (<module>)
```
To clarify: `AssembleStaticSkyModelTask` is not a thing.   It is a vanilla `AssembleCoaddTask` used as a subtask within `CompareWarpAssembleCoaddTask`. 

Only solution I can think of at the moment is:
```    
def validate(self):
        AssembleCoaddConfig.validate(self)
        self.assembleStaticSkyModel.warpType = self.staticSkyModelWarpType
```
which I'm not sure is any better because it smells like code duplication. `staticSkyModelWarpType` is just a copied and pasted `warpType` config from:
https://github.com/lsst/pipe_tasks/blob/master/python/lsst/pipe/tasks/assembleCoadd.py#L55

yalsayyad: Reverting to hardcoded 'psfMatched'. Jenkins running.

PR #172, opened 2017-12-21
ingest.py: read metadata from default HDU by default

PaulPrice:
When the input file has been FITS tile-compressed, hdu=0 (which is the
default) is the PHU which has no useful information. We need to specify
hdu=INT_MIN, but that's not possible in python so we just need a
mechanism to signal to leave off the 'hdu' parameter; choose hdu <= 0.

### DIFF
--- a/python/lsst/pipe/tasks/ingest.py
+++ b/python/lsst/pipe/tasks/ingest.py
@@ -17,6 +17,7 @@ from lsst.pex.config import Config, Field, DictField, ListField, ConfigurableFie
 import lsst.pex.exceptions
 from lsst.pipe.base import Task, InputOnlyArgumentParser
 import lsst.afw.image as afwImage
+from lsst.afw.fits import DEFAULT_HDU
 
 
 class IngestArgumentParser(InputOnlyArgumentParser):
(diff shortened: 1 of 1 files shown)


RobertLuptonTheGood:
I don't think this is right.  There's a magic number to read from default header, but it isn't <0 as @timj wanted to reserve that to read from HDUs counted from the end.  The magic isn't exported to pybind11 (it should be).

							R

PaulPrice:
All right. I'll change this to use the magic number.

python/lsst/pipe/tasks/ingest.py
```diff
@@ -17,6 +17,7 @@
 import lsst.pex.exceptions
 from lsst.pipe.base import Task, InputOnlyArgumentParser
 import lsst.afw.image as afwImage
+from lsst.afw.fits import DEFAULT_HDU
```
RobertLuptonTheGood: OK, but I really don't like these imports of specific symbols from packages.  Not worth changing, but I'd much rather see `import lsst.afw.fits as afwFits` ...  `afwFits.DEFAULT_HDU`

PR #174, opened 2018-01-16
DM-12995: Improve detection thresholds and background with temporary sky objects

PaulPrice:


PaulPrice:
Modified the commit message; please let me know if it's still inaccurate.

TallJimbo:
Afraid not; I'd say it'd be most accurate to just remove the second sentence entirely.  My understanding is that `DynamicDetectionTask` just accounts for the variance lost to covariance due to warping.  What I don't understand is why the original coadd variance rescaling didn't account for that, but I don't know how it determined the scaling factor; maybe it just wasn't using the rescaling that is relevant for detection.

PR #179, opened 2018-01-26
Tickets/dm 13345

yalsayyad:


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -83,6 +83,12 @@ class AssembleCoaddConfig(CoaddBaseTask.ConfigClass):
         doc="Number of iterations of outlier rejection; ignored if non-clipping statistic selected.",
         default=2,
     )
+    calcErrorFromInputVariance = pexConfig.Field(
+        dtype=bool,
+        doc="Calculate coadd variance from input variance by stacking statistic."
+            "Passed to StatisticsControl.setCalcErrorFromInputVariance()",
```
timothydmorton: Does this work?  Two strings on consecutive lines without any explicit concatenation?

PaulPrice: Contained within the function parens, yes.

timothydmorton: Interesting! Never seen that before.

PR #181, opened 2018-01-29
DM-13396: Fix mask propagation on coadds.

TallJimbo:
Pixels with rejections due to most masks from the warps are now indicated with REJECTED.

Pixels that had EDGE set on an input are indicated with SENSOR_EDGE.

Pixels that had CLIPPED set on an input are indicated with CLIPPED (as are pixels that were actually
sigma-clipped).

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -523,7 +523,6 @@ discussed in \ref pipeTasks_multiBand (but note that normally, one would use the
             except Exception as e:
                 self.log.fatal("Cannot compute coadd %s: %s", subBBox, e)
 
-        self.setEdge(coaddMaskedImage.getMask())
         self.setInexactPsf(coaddMaskedImage.getMask())
         # Despite the name, the following doesn't really deal with "EDGE" pixels: it identifies
         # pixels that didn't receive any unmasked inputs (as occurs around the edge of the field).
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -746,8 +732,9 @@ def setInexactPsf(self, mask):
         inexactPsf = mask.getPlaneBitMask("INEXACT_PSF")
         sensorEdge = mask.getPlaneBitMask("SENSOR_EDGE")  # chip gaps
         clipped = mask.getPlaneBitMask("CLIPPED")  # pixels clipped from coadd
+        rejected = mask.getPlaneBitMask("REJECTED")  # pixels clipped from coadd
```
PaulPrice: Fix comment.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -600,8 +599,19 @@ def assembleSubregion(self, coaddExposure, bbox, tempExpRefList, imageScalerList
...
+        coaddExposure.mask.addMaskPlane("CLIPPED")
+        coaddExposure.mask.addMaskPlane("SENSOR_EDGE")
+        # If a pixel is rejected due to a mask value other than EDGE, NO_DATA,
+        # or CLIPPED, set it to REJECTED on the coadd.
+        # If a pixel is rejected due to EDGE, set the coadd pixel to SENSOR_EDGE.
+        # if a pixel is rejected due to CLIPPED, set the coadd pixel to CLIPPED.
+        edge = afwImage.Mask.getPlaneBitMask("EDGE")
+        noData = afwImage.Mask.getPlaneBitMask("NO_DATA")
+        clipped = afwImage.Mask.getPlaneBitMask("CLIPPED")
+        toReject = statsCtrl.getAndMask() & (~noData) & (~edge) & (~clipped)
```
PaulPrice: Thanks for being so careful here. Without excluding those others, this feature wouldn't be useful.

PR #190, opened 2018-03-16
DM-11162: Replace all use of Coord and subclasses with SpherePoint

r-owen:


tests/test_coaddInputs.py
```diff
@@ -97,11 +96,11 @@ def makeExposure(self, universalId):
         exp.setDetector(detector)
 
         expInfo = exp.getInfo()
-        scale = 5.1e-5 * degrees
-        cdMatrix = makeCdMatrix(scale=scale)
-        wcs = makeSkyWcs(
+        scale = 5.1e-5 * lsst.afw.geom.degrees
```
rearmstr: Since you are change this line, you can remove the white space around ```*```.  Same below.

PR #191, opened 2018-03-23
DM-13854: measureCoaddSources: add command-line PSF cache configuration

PaulPrice:
CoaddPsf is the current tall pole in running measureCoaddSources.
Caching it is simple, inexpensive and realises a hefty gain (25%)
in the runtime.

### DIFF
--- a/python/lsst/pipe/tasks/multiBand.py
+++ b/python/lsst/pipe/tasks/multiBand.py
@@ -887,6 +887,13 @@ class MeasureMergedCoaddSourcesConfig(Config):
 ## \}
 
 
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/multiBand.py
```diff
@@ -1055,7 +1063,7 @@ def __init__(self, butler=None, schema=None, peakSchema=None, refObjLoader=None,
         if self.config.doRunCatalogCalculation:
             self.makeSubtask("catalogCalculation", schema=self.schema)
 
-    def run(self, patchRef):
+    def run(self, patchRef, psfCache=100000):
```
TallJimbo: Switch to 100?

PR #193, opened 2018-04-12
Exclude bad pixel mask from nImage

yalsayyad:
passed to method that filters artifacts.
Note that this nImage is based on the psfMatched mask and will
only approximate the final nImage

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -1681,6 +1681,7 @@ class CompareWarpAssembleCoaddTask(AssembleCoaddTask):
         spanSetArtifactList = []
         spanSetNoDataMaskList = []
         spanSetEdgeList = []
+        badPixelMask = self.getBadPixelMask()
 
         # mask of the warp diffs should = that of only the warp
         templateCoadd.mask.clearAllMaskPlanes()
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1693,7 +1693,8 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
         for warpRef, imageScaler in zip(tempExpRefList, imageScalerList):
             warpDiffExp = self._readAndComputeWarpDiff(warpRef, imageScaler, templateCoadd)
             if warpDiffExp is not None:
-                nImage.array += numpy.where(numpy.isnan(warpDiffExp.image.array),
+                # This nImage only approximates the final nImage because it uses the PSF-matched mask
+                nImage.array += numpy.where(warpDiffExp.mask.array & self.getBadPixelMask(),
                                             0, 1).astype(numpy.uint16)
```
natelust: Consider:
`(~np.bitwise_and(warpDiffExp.mask.array, self.getBadPixelMask()).astype(bool)).astype(np.uint16)`
instead of your where statement if speed is an issue here, as this is about 4 times faster, however where is still in the hundreds of millisecond range, so maybe it does not mater.

likewise above can be,
`(np.isfinite(warpDiffExp.image.array) * ~np.bitwise_and(warpDiffExp.mask.array, self.getBadPixelMask()).astype(bool)).astype(np.uint16)`
if you want to check for nans.

Also, if speed is an issue (again I am not sure how often this is executing, consider taking the call to getBadPixelMask outside the loop and assigning it to a variable.

If speed is not an issue, looks good.

yalsayyad: Alright, going with a slightly more readable version of your second suggestion:
```(numpy.isfinite(warpDiffExp.image.array) *
                                 ((warpDiffExp.mask.array & badPixelMask) == 0)).astype(numpy.uint16)```

PR #194, opened 2018-04-17
Prefilter artifact candidates based on bad pixel mask

yalsayyad:
Overlapping defects can eat up the temporal budget and are excluded
from coadds by their masks. Remove these artifact candidates from the
list before they contribute to the epochCountImage.

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -1394,6 +1394,24 @@ class CompareWarpAssembleCoaddConfig(AssembleCoaddConfig):
         dtype=bool,
         default=True,
     )
+    doPrefilterArtifacts = pexConfig.Field(
+        doc="Ignore artifact candidates that are mostly covered by the bad pixel mask, "
+            "because they will be excluded anyway. This prevents them from contributing "
+            "to the outlier epoch count image and potentially being labeled as persistant."
+            "'Mostly' is defined by the config 'prefilterArtifactsRatio'.",
+        dtype=bool,
+        default=True
+    )
+    prefilterArtifactsMaskPlanes = pexConfig.ListField(
+        doc="Prefilter artifact candidates that are mostly covered by these bad mask planes.",
+        dtype=str,
+        default=('NO_DATA', 'BAD', 'SAT', 'SUSPECT'),
+    )
+    prefilterArtifactsRatio = pexConfig.Field(
+        doc="Prefilter artifact candidates with less than this fraction overlapping good pixels",
+        dtype=float,
+        default=0.05
+    )
 
     def setDefaults(self):
         AssembleCoaddConfig.setDefaults(self)
@@ -1702,6 +1720,10 @@ class CompareWarpAssembleCoaddTask(AssembleCoaddTask):
                 footprints = fpSet.positive
                 slateIm.set(0)
                 spanSetList = [footprint.spans for footprint in footprints.getFootprints()]
+
+                # Remove artifacts due to defects before they contribute to the epochCountImage
+                if self.config.doPrefilterArtifacts:
+                    spanSetList = self._prefilterArtifacts(spanSetList, warpDiffExp)
                 for spans in spanSetList:
                     spans.setImage(slateIm, 1, doClip=True)
                 epochCountImage += slateIm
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1746,6 +1766,32 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
                              'EDGE': edge})
         return altMasks
 
+    def _prefilterArtifacts(self, spanSetList, exp):
+        """!
+        \brief Remove artifact candidates covered by bad mask plane
```
PaulPrice: I suggest using numpydoc, even if it doesn't match what's already in the file --- but maybe that's naughty.
My reasoning is that it looks better, doxygen isn't going to parse python for much longer (if it still is), and someone will thank you when it comes time to update.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1746,6 +1766,32 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
...
+
+        Any future editing of the candidate list that does not depend on
+        temporal information should go in this method.
+
+        @param spanSetList: List of SpanSets representing artifact candidates
+        @param exp: Exposure containing mask planes used to prefilter
+
+        return List of SpanSets with artifacts
+        """
+        badPixelMask = afwImage.Mask.getPlaneBitMask(self.config.prefilterDefectsMaskPlanes)
```
PaulPrice: I think you should prefer `exp.mask.getPlanceBitMask(...)`.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1746,6 +1766,32 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
...
+        goodArr = numpy.where(exp.mask.array & badPixelMask, False, True)
```
PaulPrice: Maybe clearer: `numpy.where((exp.mask.array & badPixelMask) == 0))`

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1746,6 +1766,32 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
...
+        returnSpanSetList = []
+        bbox = exp.getBBox()
+        x0, y0 = exp.getXY0()
+        for i, span in enumerate(spanSetList):
+            y, x = span.clippedTo(bbox).indices()
+            yIdxLocal = [y1 - y0 for y1 in y]
+            xIdxLocal = [x1 - x0 for x1 in x]
```
PaulPrice: Aren't `x` and `y` numpy arrays, so you can just do:

    yIdxLocal = y - y0
    xIdxLocal = x - x0

If they're not numpy arrays, just cast them.

And I think you should spell out `xIndexLocal`.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1746,6 +1766,32 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
...
+            goodRatio = numpy.count_nonzero(goodArr[yIdxLocal, xIdxLocal])/span.getArea()
```
PaulPrice: I would have written this as `goodArr[yIndexLocal, xIndexLocal].sum()/span.getArea()`, but yours works too, perhaps even clearer.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1746,6 +1766,32 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
                              'EDGE': edge})
         return altMasks
 
+    def _prefilterArtifacts(self, spanSetList, exp):
```
PaulPrice: Suggest dropping the leading underscore. I don't think the method is as private as this indicates.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1394,6 +1394,22 @@ class CompareWarpAssembleCoaddConfig(AssembleCoaddConfig):
         dtype=bool,
         default=True,
     )
+    doPrefilterDefects = pexConfig.Field(
+        doc="Ignore artifacts that are mostly covered by the bad pixel mask,"
```
PaulPrice: What does it mean to "ignore" an artifact? It's not that the pixels' contributions to the coadd are ignored, right? That's covered by other configuration parameters. But it's the pixels' contributions to the count image. Why is that important?

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: I promise that future person will be me: DM-13587

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -1702,6 +1718,10 @@ def findArtifacts(self, templateCoadd, tempExpRefList, imageScalerList):
                 footprints = fpSet.positive
                 slateIm.set(0)
                 spanSetList = [footprint.spans for footprint in footprints.getFootprints()]
+
+                # Remove artifacts due to defects before they contribute to the epochCountImage
+                if self.config.doPrefilterDefects:
+                    spanSetList = self._prefilterArtifacts(spanSetList, warpDiffExp)
```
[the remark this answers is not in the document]
yalsayyad: Sure. Renamed *Defects* to *Artifacts* on all three config parameters to match.

python/lsst/pipe/tasks/assembleCoadd.py
yalsayyad: This failed a unit test. Reverting. 

            y, x = span.clippedTo(bbox).indices()
            yIndexLocal = numpy.array(y) - y0
            xIndexLocal = numpy.array(x) - x0
>           goodRatio = numpy.count_nonzero(goodArr[yIndexLocal, xIndexLocal])/span.getArea()
E           TypeError: tuple indices must be integers or slices, not tuple

PR #196, opened 2018-05-08
DM-9832 Ensure contiguous catalogs after detect/deblend

parejkoj:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -511,6 +511,11 @@ def calibrate(self, exposure, exposureIdInfo=None, background=None,
             self.copyIcSourceFields(icSourceCat=icSourceCat,
                                     sourceCat=sourceCat)
 
+        # NOTE: sourceSelectors require contiguous catalogs, so ensure
+        # contiguity now, so views are preserved from here on.
+        if not sourceCat.isContiguous():
+            sourceCat = sourceCat.copy(deep=True)
+
```
TallJimbo: I'd prefer to accomplish this by reserving the necessary space in some combination of `SourceDeblendTask`, `SourceDetectionTask`, or `MergeCoaddDetectionsTask`, which we can calculate as the number of `Footprints` (the number of parents) plus the total number of peaks in all `Footprints` with more than one peak.

But that would be fairly disruptive because it would require callers of these Tasks to use returned catalogs instead of assuming they operate in-place, so it's probably not something I'd recommend doing on this ticket.  But I would appreciate having a ticket created for this and a TODO referencing it dropped in here.

parejkoj: I totally agree, and had meant to file that ticket but forgot. Thanks for the reminder...

Ha! It turns out, you'd already filed it last August! DM-11568. I'll add a TODO here and link that ticket to this epic and related RFCs.

I had explored some options for doing the `reserve` step, but it wasn't trivially obvious, and deserves a good test suite to make sure it's not overshooting too much.

PR #197, opened 2018-05-08
DM-13790: Remove all use of the geom package

r-owen:


python/lsst/pipe/tasks/makeDiscreteSkyMap.py
```diff
@@ -174,9 +174,11 @@ def run(self, butler, dataRefList):
...
             skyMapConfig.radiusList.extend(oldSkyMap.config.radiusList)
         skyMapConfig.update(**self.config.skyMap.toDict())
-        skyMapConfig.raList.append(circle.center[0])
-        skyMapConfig.decList.append(circle.center[1])
-        skyMapConfig.radiusList.append(circle.radius + self.config.borderSize)
+        circleCenter = lsst.sphgeom.LonLat(circle.getCenter())
+        skyMapConfig.raList.append(circleCenter[0].asDegrees())
+        skyMapConfig.decList.append(circleCenter[1].asDegrees())
+        circleRadiusDeg = circle.getOpeningAngle().asDegrees()
+        skyMapConfig.radiusList.append(circleRadiusDeg + self.config.borderSize)
```
PaulPrice: What's wrong with `circle.radius`?

r-owen: `lsst.sphgeom.Circle` does not support `radius` or `getRadius`; the equivalent method is `getOpeningAngle()`. Unfortunately the `sphgeom` pybind11 wrapper doesn't seem to use properties, as `openingAngle` would be nicer than `getOpeningAngle()`. We make small enough use of `sphgeom` that I figured we could live with that, but I am not happy about it.

PR #202, opened 2018-05-21
DM-13950: Convert assembleCoadd.py to numpydoc

mrawls:
The scope of this ticket changed to update only assembleCoadd.py with numpydoc for now.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,0 +1,1 @@
+Copyright (c) 2008-2018 Association of Universities for Research in Astronomy, Inc.
--- a/doc/lsst.pipe.tasks/index.rst
+++ b/doc/lsst.pipe.tasks/index.rst
@@ -11,4 +11,4 @@ lsst.pipe.tasks
 Python API reference
 ====================
 
-.. automodapi:: lsst.pipe.tasks
+.. automodapi:: lsst.pipe.tasks.assembleCoadd
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -1,9 +1,9 @@
+# This file is part of pipe_tasks.
 #
 # LSST Data Management System
-# Copyright 2008-2016 AURA/LSST.
-#
 # This product includes software developed by the
 # LSST Project (http://www.lsst.org/).
+# See COPYRIGHT file at the top of the source tree.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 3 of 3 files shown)


doc/lsst.pipe.tasks/index.rst
```diff
@@ -12,3 +12,6 @@ Python API reference
 ====================
 
 .. automodapi:: lsst.pipe.tasks
+   :skip: AssembleCoaddTask, SafeClipAssembleCoaddTask, CompareWarpAssembleCoaddTask
```
jonathansick: To get the API docs to show up, you'll need to change this to:

```
.. automodapi:: lsst.pipe.tasks.assembleCoadd
```

We'll generalize it to all of `lsst.pipe.tasks` when those APIs are available.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -938,169 +1032,168 @@ def validate(self):
...
+
+       assembleCoadd.py --help
+
+    To demonstrate usage of the `SafeClipAssembleCoaddTask` in the larger
+    context of multi-band processing, we will generate the HSC-I & -R band
+    coadds from HSC engineering test data provided in the ci_hsc package.
+    To begin, assuming that the lsst stack has been already set up, we must
+    set up the obs_subaru and ci_hsc packages. This defines the environment
+    variable $CI_HSC_DIR and points at the location of the package. The raw
+    HSC data live in the ``$CI_HSC_DIR/raw``h v directory. To begin assembling
```
jonathansick: This was generating a build error. Should be:

```
HSC data live in the ``$CI_HSC_DIR/raw`` directory. To begin assembling
```

doc/lsst.pipe.tasks/index.rst
mrawls: Two questions here. (1) I was originally going to have just assembleCoadd listed, but it looks like @r-owen recently added all of lsst.pipe.tasks instead. Perhaps there's some special case I'm missing? (2) Does your comment here affect the `:skip:` line or just the `.. automodapi::` line?

COPYRIGHT
```diff
@@ -0,0 +1 @@
+Copyright (c) 2008-2018 Association of Universities for Research in Astronomy, Inc.
```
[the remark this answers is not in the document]
mrawls: Indeed. Is this an acceptable interim option, or should I add a note saying the copyright file is incomplete or something? Happy to make a ticket for it.

doc/lsst.pipe.tasks/index.rst
jonathansick: The comment is a replacement for both lines.

`lsst.pipe.tasks` doesn't import anything, turns out, so we need to point directly at individual modules

COPYRIGHT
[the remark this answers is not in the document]
jonathansick: I... wouldn't make a ticket just for this. I think it's understood that copyrights need to be systematically audited everywhere given the new system so there's gotta be an overarching ticket for all this anyways. I'll figure out the who and how of this work.

PR #203, opened 2018-05-25
DM-11953: Improve MaxPsfWcsSelectImagesTask

ebellm:
add a best seeing WCS selector.

### DIFF
--- a/python/lsst/pipe/tasks/selectImages.py
+++ b/python/lsst/pipe/tasks/selectImages.py
@@ -25,7 +25,6 @@ import lsst.pex.config as pexConfig
 import lsst.pex.exceptions as pexExceptions
 import lsst.afw.geom as afwGeom
 import lsst.pipe.base as pipeBase
-from lsst.geom import convexHull
 
 __all__ = ["BaseSelectImagesTask", "BaseExposureInfo", "WcsSelectImagesTask", "PsfWcsSelectImagesTask",
            "DatabaseSelectImagesConfig", "BestSeeingWcsSelectImagesTask"]
@@ -380,20 +379,23 @@ class BestSeeingWcsSelectImagesTask(WcsSelectImagesTask):
 
         Returns
         -------
-        pipe.base.Struct with filtered exposureList and dataRefList
-        (if makeDataRefList is True).
+        result : `lsst.pipe.base.Struct`
+            Result struct with components:
+            - ``exposureList``: the selected exposures
+                (`list` of `lsst.pipe.tasks.selectImages.BaseExposureInfo`).
+            - ``dataRefList``: the optional data references corresponding to
+                each element of ``exposureList``
+                (`list` of `lsst.daf.persistence.ButlerDataRef`, or `None`).
 
         Notes
         -----
-        We use the "convexHull" function in the geom package to define
+        We use the "convexHull" method of lsst.sphgeom.ConvexPolygon to define
         polygons on the celestial sphere, and test the polygon of the
         patch for overlap with the polygon of the image.
 
-        We use "convexHull" instead of generating a SphericalConvexPolygon
-        directly because the standard for the inputs to SphericalConvexPolygon
+        We use "convexHull" instead of generating a ConvexPolygon
+        directly because the standard for the inputs to ConvexPolygon
         are pretty high and we don't want to be responsible for reaching them.
-        If "convexHull" is found to be too slow, we can revise this.
-
         """
         if self.config.nImagesMax <= 0:
             raise RuntimeError(f"nImagesMax must be greater than zero: {self.config.nImagesMax}")
@@ -403,7 +405,7 @@ class BestSeeingWcsSelectImagesTask(WcsSelectImagesTask):
         exposureInfoList = []
 
         patchVertices = [coord.getVector() for coord in coordList]
-        patchPoly = convexHull(patchVertices)
+        patchPoly = lsst.sphgeom.ConvexPolygon.convexHull(patchVertices)
 
         for data in selectDataList:
             dataRef = data.dataRef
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/selectImages.py
```diff
@@ -339,3 +339,125 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
+                imageCorners = [imageWcs.pixelToSky(pix) for pix in imageBox.getCorners()]
+            except (pexExceptions.DomainError, pexExceptions.RuntimeError) as e:
+                # Protecting ourselves from awful Wcs solutions in input images
+                self.log.debug("WCS error in testing calexp %s (%s): deselecting", dataRef.dataId, e)
+                continue
+
+            imagePoly = lsst.sphgeom.ConvexPolygon.convexHull([coord.getVector() for coord in imageCorners])
+            if imagePoly is None:
+                self.log.debug("Unable to create polygon from image %s: deselecting", dataRef.dataId)
+                continue
```
kfindeisen: This amounts to silently dropping all invalid polygons. If you're not willing to abort, at least print a warning or higher.

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -339,3 +339,125 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
+            cal = dataRef.get("calexp", immediate=True)
+            psfSize = cal.getPsf().computeShape().getDeterminantRadius()
+            nx, ny = cal.getDimensions()
+
+            imageBox = afwGeom.Box2D(afwGeom.Point2D(0, 0), afwGeom.Extent2D(nx, ny))
+            try:
+                imageCorners = [imageWcs.pixelToSky(pix) for pix in imageBox.getCorners()]
+            except (pexExceptions.DomainError, pexExceptions.RuntimeError) as e:
+                # Protecting ourselves from awful Wcs solutions in input images
+                self.log.debug("WCS error in testing calexp %s (%s): deselecting", dataRef.dataId, e)
```
kfindeisen: This amounts to silently dropping problematic WCS (and probably other things, given how broad the catch clause is). If you're not willing to abort, at least print a warning or higher.

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -339,3 +339,125 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
+            if imagePoly is None:
+                self.log.debug("Unable to create polygon from image %s: deselecting", dataRef.dataId)
+                continue
+
+            if patchPoly.intersects(imagePoly):
+                # "intersects" also covers "contains" or "is contained by",
+                # so there is no guarantee the whole area is covered
+
+                # if min/max PSF values are defined, remove images out of bounds
+                sizeFwhm = psfSize * np.sqrt(8.*np.log(2.))
```
kfindeisen: Please move the definition of `psfSize` to here, since there's no need for it way up at the start of the method and it's confusing to have it so separated from its usage.

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -339,3 +339,125 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
+                if self.config.maxPsfFwhm and sizeFwhm > self.config.maxPsfFwhm:
+                    continue
+                if self.config.minPsfFwhm and sizeFwhm > self.config.minPsfFwhm:
```
kfindeisen: Should this be `<`?

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -339,3 +339,125 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
+        if len(psfSizes) > self.config.nImagesMax:
+            sortedIndices = np.argsort(psfSizes)[:self.config.nImagesMax]
+            filteredDataRefList = [dataRefList[i] for i in sortedIndices]
+            filteredExposureInfoList = [exposureInfoList[i] for i in sortedIndices]
+            self.log.info(f"{len(sortedIndices)} images selected with FWHM "
+                          f"range of {psfSizes[sortedIndices[0]]}--{psfSizes[sortedIndices[-1]]} pixels")
+
+        else:
+            if len(psfSizes) == 0:
+                self.log.info(f"0 images selected.")
```
kfindeisen: Should this be info, or warn?

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -339,3 +339,125 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
+            else:
+                self.log.info(f"{len(psfSizes)} images selected with FWHM range "
+                              f"of {psfSizes[0]}--{psfSizes[-1]} pixels")
```
kfindeisen: This should almost certainly be `debug` -- it's not information of use to somebody running the pipeline normally.

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -339,3 +339,125 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True, selectDataList=[]
...
+    """
+    nImagesMax = pexConfig.Field(
+        dtype=int,
+        doc="Maximum number of images to select",
+        default=5)
+    maxPsfFwhm = pexConfig.Field(
+        dtype=float,
+        doc="Maximum PSF FWHM (in pixels) to select",
+        default=5.,
+        optional=True)
```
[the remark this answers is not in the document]
ebellm: Yes--the code checks if maxPsfFwhm is not None before using the value.

python/lsst/pipe/tasks/selectImages.py
ebellm: For both of these we want to be permissive--the goal is to get a usable coadd, not to guarantee that all images can be processed.

PaulPrice:
There's a lot of code copied here. Can you refactor out the common code?

ebellm:
@PaulPrice Probably only if I nest a call to WcsSelectImagesTask.  Is that a pattern we're comfortable with?

PaulPrice:
Absolutely! I'm thinking you should first call `WcsSelectImagesTask.runDataRef` and then iterate over the results from that.

python/lsst/pipe/tasks/selectImages.py
kfindeisen: It's not ignoring images itself that bothers me, it's the fact that it happens almost silently. Unless you mean that dropping images in one of these two cases is a normal occurrence, not just an acceptable one?

ebellm: With the refactor this is now upstream in `WcsSelectImagesTask`...

PR #206, opened 2018-06-20
DM-14842: Fix deprecation warnings from PropertyList/Set.get

r-owen:


bin.src/reportTaskTiming.py
```diff
@@ -168,7 +168,7 @@ def run(self, dataRefList):
                     resInfo = ResourceInfo(taskName, self.log)
                     resourceDict[taskName] = resInfo
                 if resInfo.isValidName(itemName):
-                    resInfo.setItem(itemName, taskMetadata.get(name, True))
+                    resInfo.setItem(itemName, taskMetadata.getArray(name))
```
yalsayyad: Out of curiosity, do you know what the "True" did on the old API?

r-owen: True was the value for argument `asArray` which forced the return value to be an array (if numeric or scalar). Unfortunately there was no way to force a scalar or the old API might have sufficed; instead the default behavior was to return a scalar if there was only one value, else an array. This occasionally backfired horribly.

This was the only bit of code using the `asArray` argument (at least properly; I recently found and fixed an instance of misuse on an earlier ticket -- code that assumed the argument was a default to return if the item was not present).

PR #211, opened 2018-08-08
tickets/DM-15104

fred3m:


python/lsst/pipe/tasks/multiBand.py
```diff
@@ -818,14 +818,271 @@ def getSkySourceFootprints(self, mergedList, skyInfo, seed):
...
+            Raised if ``parsedCmd`` and ``args`` are both `None`.
+        """
+        if parsedCmd is not None:
+            butler = parsedCmd.butler
+        elif args is not None:
+            dataRefList, kwargs = args
+            butler = dataRefList[0].getButler()
+        else:
+            raise RuntimeError("Neither parsedCmd or args specified")
+        return self.TaskClass(config=self.config, log=self.log, butler=butler)
```
TallJimbo: Seems like you could just use `MergeSourcesRunner` instead of writing your own, *if* you made `psfCache` a config parameter (which might be a good idea anyway, rather than making it special - though I can see how you might want that, given that it's not an option that should change the outputs).

Even if you do want to keep `psfCache` as a special command-line argument, I'm hoping you could inherit from `MergeSourcesRunner` and eliminate most of this code.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -1123,7 +1368,7 @@ def readSources(self, dataRef):
         We also need to add columns to hold the measurements we're about to make
         so we can measure in-place.
         """
-        merged = dataRef.get(self.config.coaddName + "Coadd_mergeDet", immediate=True)
+        merged = dataRef.get(self.config.coaddName + self.inputCatalog, immediate=True)
```
natelust: This will produce different dataset names, is that ok? There is no Coadd_ part of the string in your new version.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -973,16 +973,18 @@ def run(self, patchRefList, psfCache=100):
...
-            fluxCatalogs, templateCatalogs = self.multibandBeblend.run(exposures, sources)
-            for n in len(patchRefList):
-                self.write(patchRefList[n], fluxCatalogs[n], templateCatalogs[n])
+            filters = []
+            exposures = []
+            for patchRef in patchRefList:
+                exposure = patchRef.get(self.config.coaddName + "Coadd_calexp", immediate=True)
+                filters.append(patchRef.dataId["filter"])
+                exposures.append(exposure)
+            # The input sources are the same for all bands, since it is a merged catalog
```
natelust: Are you sure this will always be a merged catalog? what happens when it is not one

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -973,16 +973,18 @@ def run(self, patchRefList, psfCache=100):
...
-            # implement deblender here
-            # ...
-
-            # this is the proper code to write
-            fluxCatalogs, templateCatalogs = self.multibandBeblend.run(exposures, sources)
-            for n in len(patchRefList):
-                self.write(patchRefList[n], fluxCatalogs[n], templateCatalogs[n])
+            filters = []
+            exposures = []
+            for patchRef in patchRefList:
```
natelust: see comment on previous commit about splitting out reading

python/lsst/pipe/tasks/multiBand.py
fred3m: What is `psfCache`? I just added it because it's something that `MeasureMergedCoaddSources` had, but I don't necessarily know what it does, and if it's worth keeping.

TallJimbo: Ah, I bet I know, but my knowledge of the details is sufficiently limited that we should ask @PaulPrice.

PaulPrice: It's the `lsst::utils::Cache` embedded in `lsst::afw::detection::Psf` that caches the most recent PSF image computations. Changing the capacity of the cache influences the job runtime and memory usage.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -818,14 +818,271 @@ def getSkySourceFootprints(self, mergedList, skyInfo, seed):
...
+            Flux conserved sources to write to file.
+            If using the single band deblender, this is the catalog
+            generated.
+        template_sources: `SourceCatalog`
+            Source catalog using the multiband template models
+            as footprints.
+        """
+        # The multiband deblender does not have to conserve flux,
+        # so only write the flux conserved catalog if it exists
+        if flux_sources is not None:
```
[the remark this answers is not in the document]
fred3m: True, but I don't think there is any harm in explicitly checking for `None`, since `flux_sources` should either be a `SourceCatalog` or `None` and this is slightly more robust.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -1033,18 +1288,18 @@ def __init__(self, butler=None, schema=None, peakSchema=None, refObjLoader=None,
         measurements.
         """
         CmdLineTask.__init__(self, **kwargs)
+        if self.config.inputCatalog.startswith("deblended"):
+            self.deblended = True
+        else:
+            self.deblended = False
```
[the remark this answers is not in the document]
fred3m: indeed.

python/lsst/pipe/tasks/multiBand.py
fred3m: This is expected. Measurements can now be run on one of three different catalogs: the flux conserved deblender output `Coadd_deblendedFlux`, the scarlet model output `Coadd_deblendedModel`  or the merged catalog (if no deblender is run) `Coadd_mergeDet`.

python/lsst/pipe/tasks/multiBand.py
fred3m: This is the same way `MeasureMergedCoaddSourcesTask` was executed, so the input is always a merged catalog (see https://github.com/lsst/pipe_tasks/blob/tickets/DM-15104/python/lsst/pipe/tasks/multiBand.py#L1019).

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -818,14 +818,271 @@ def getSkySourceFootprints(self, mergedList, skyInfo, seed):
...
+            for n in len(patchRefList):
+                self.write(patchRefList[n], fluxCatalogs[n], templateCatalogs[n])
+        else:
+            # Use the singeband deblender to deblend each band separately
+            for patchRef in patchRefList:
+                exposure = patchRef.get(self.config.coaddName + "Coadd_calexp", immediate=True)
+                exposure.getPsf().setCacheCapacity(psfCache)
+                sources = self.readSources(patchRef)
+                self.singleBandDeblend.run(exposure, sources)
+                self.write(patchRef, sources)
```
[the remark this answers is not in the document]
fred3m: I think that this is a non-trivial change and someone who better understands `PipelineTask` is better equipped to decided how to split this up, if it can be done in a way that is consistent with other tasks.

PR #213, opened 2018-08-23
DM-10302 Change flux->instFlux

parejkoj:


examples/photoCalTask.py
```diff
@@ -102,7 +102,7 @@ def run():
         print("Adding columns to the source catalogue")
         cat = afwTable.SourceCatalog(schema)
         cat.table.defineCentroid(srcCat.table.getCentroidDefinition())
-        cat.table.definePsfFlux(srcCat.table.getPsfFluxDefinition())
+        cat.table.definePsfFlux(srcCat.table.schema.getAliasMap().get("slot_psfFlux"))
```
r-owen: Are you sure this will be in the alias map? Did you try using `schema.get("slot_psfFlux")`?

parejkoj: I'd been working off the recommendation from the original deprecation warning, but yours is better, I think.

PR #215, opened 2018-08-24
DM-15532: Convert getTargetList to a classmethod

isullivan:


isullivan:
This ticket was closed as a "Won't fix", since it created more problems than it solved. It should also become obsolete with the new `pipelineTask` framework.

PR #218, opened 2018-09-05
DM-9358: Fix setting of calib_psf_candidate flag to match docstring description

laurenam:
The docstring for this flag reads:

"Flag set if the source was a candidate for PSF determination as
determined by the star selector."

However, if the reserve fraction was set to a non-zero value, an
object that was selected as a PSF candidate by the star selector
but then flagged as reserved would have the calib_psfCandidate flag
set to False. This is inconsistent with the docstring (and erroneously
implies the object was not deemed a suitable candidate). This updates
the setting of the flag to reflect its docstring description.

### DIFF
--- a/python/lsst/pipe/tasks/measurePsf.py
+++ b/python/lsst/pipe/tasks/measurePsf.py
@@ -273,16 +273,18 @@ into your debug.py file and run measurePsfTask.py with the @c --debug flag.
         #
         stars = self.starSelector.run(sourceCat=sources, matches=matches, exposure=exposure)
         selectionResult = self.makePsfCandidates.run(stars.sourceCat, exposure=exposure)
+        self.log.info("PSF star selector found %d candidates" % len(selectionResult.psfCandidates))
         reserveResult = self.reserve.run(selectionResult.goodStarCat, expId=expId)
-        psfCandidateList = [cand for cand, use
-                            in zip(selectionResult.psfCandidates, reserveResult.use) if use]
+        # Make list of psf candidates to send to the determiner (omitting those marked as reserved)
+        psfDeterminerList = [cand for cand, use
+                             in zip(selectionResult.psfCandidates, reserveResult.use) if use]
 
-        if psfCandidateList and self.candidateKey is not None:
-            for cand in psfCandidateList:
+        if selectionResult.psfCandidates and self.candidateKey is not None:
+            for cand in selectionResult.psfCandidates:
                 source = cand.getSource()
                 source.set(self.candidateKey, True)
 
-        self.log.info("PSF star selector found %d candidates" % len(psfCandidateList))
+        self.log.info("Sending %d candidates to PSF determiner" % len(psfDeterminerList))
 
         if display:
             frame = display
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/measurePsf.py
```diff
@@ -273,16 +273,17 @@ def run(self, exposure, sources, expId=0, matches=None):
         #
         stars = self.starSelector.run(sourceCat=sources, matches=matches, exposure=exposure)
         selectionResult = self.makePsfCandidates.run(stars.sourceCat, exposure=exposure)
+        self.log.info("PSF star selector found %d candidates" % len(selectionResult.psfCandidates))
         reserveResult = self.reserve.run(selectionResult.goodStarCat, expId=expId)
         psfCandidateList = [cand for cand, use
                             in zip(selectionResult.psfCandidates, reserveResult.use) if use]
```
rearmstr: To be consistent you may want to change the name of the list that gets passed to the determiner.

laurenam: I'm not sure what you mean here...what are you wanting me to be consistent with?  **psfCandidateList** is (and has always been) the name of the list that gets sent to the determiner.  It is a sub-selection of the full potential candidate list from the **makePsfCandidates** Task that gets stored in **selectionResult.psfCandidates**.  All I did was make sure all entries in the latter got their **calib_psf_candidate** flag set to **True** (whereas before it was only the former list getting the flag set, so some bonafide candidates were not marked as such if they ended up on the reserve list).

Perhaps you want a more descriptive name for that list? (e.g. **determinerPsfCandidateList**, **psfCandidateListForDeterminer**, **finalPsfCandidateList**,... they all seem undesirable/ambiguous in their own way...)

rearmstr: My point was that we have defined the `candidate` list as all objects that could potentially be a PSF.  We then make a selection for the determiner.  What about `psfDeterminerList`?  I'm not sure that is any better.  I guess it seems strange that we have changed the code to correctly label candidates, but then we still call the wrong list candidates.

laurenam: I'm still not entirely sure we're on the same page.  As things stand (and trying to use your wording above in my definitions):
**selectionResult.psfCandidates** == "the candidate list as all objects that could potentially be a PSF"
**psfCandidateList** == "selection for the determiner" (i.e. this is NOT the full potential candidate list from above, it is a sub-selection of the above list omitting the reserved objects)

So, if we agree on the above, are you asking simply that I change the name of the latter list from
**psfCandidateList** ==> **psfDeterminerList** ?
I think removing the word "candidate" is confusing, but there doesn't seem to be a permutation without the name actually turning into a sentence that is fully descriptive, so if you're keen on **psfDeterminerList**, I'm happy to go with it (we're really only talking about 3 occurrences of this name anyway)!

rearmstr: Yeah, I'm happy with that.  I think this slightly mislabeled name is why I originally put the bug there in the first place.

PR #221, opened 2018-09-18
Tickets/dm-15636: Improve DcrModel regularization

isullivan:
These changes provide greater flexibility in applying regularization to DcrModels between iterations and between frequencies, and reduce artifacts in the final images.

### DIFF
--- a/python/lsst/pipe/tasks/dcrAssembleCoadd.py
+++ b/python/lsst/pipe/tasks/dcrAssembleCoadd.py
@@ -72,33 +72,33 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
     )
     modelClampFactor = pexConfig.Field(
         dtype=float,
-        doc="Maximum relative change of the model allowed between iterations.",
+        doc="Maximum relative change of the model allowed between iterations. Set to zero to disable.",
         default=2.,
     )
-    regularizeSigma = pexConfig.Field(
+    modelClampSigma = pexConfig.Field(
         dtype=float,
-        doc="Threshold to exclude noise-like pixels from regularization.",
-        default=3.,
-    )
-    clampFrequency = pexConfig.Field(
-        dtype=float,
-        doc="Maximum relative change of the model allowed between subfilters.",
+        doc="Threshold to exclude noise-like pixels from regularization between iterations.",
         default=2.,
     )
-    maxGain = pexConfig.Field(
+    frequencyClampSigma = pexConfig.Field(
         dtype=float,
-        doc="Maximum convergence-weighted gain to apply between forward modeling iterations.",
-        default=2.,
+        doc="Threshold to exclude noise-like pixels from regularization across subfilters.",
+        default=1.,
     )
-    minGain = pexConfig.Field(
+    frequencyClampFactor = pexConfig.Field(
         dtype=float,
-        doc="Minimum convergence-weighted gain to apply between forward modeling iterations.",
-        default=0.5,
+        doc="Maximum relative change of the model allowed between subfilters. Set to zero to disable.",
+        default=2.,
     )
     convergenceMaskPlanes = pexConfig.ListField(
         dtype=str,
         default=["DETECTED"],
         doc="Mask planes to use to calculate convergence."
+    )
+    regularizationWidth = pexConfig.Field(
+        dtype=int,
+        default=2,
+        doc="Minimum radius of a region to include in regularization, in pixels."
     )
     imageWarpMethod = pexConfig.Field(
         dtype=str,
@@ -157,7 +157,7 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
     typically appears in the form of holes next to sources in one subfilter,
     and corresponding extended wings in another. Because each subfilter has
     a narrow bandwidth we assume that physical sources that are above the noise
-    level will not vary in flux by more than a factor of `clampFrequency`
+    level will not vary in flux by more than a factor of `frequencyClampFactor`
     between subfilters, and pixels that have flux deviations larger than that
     factor will have the excess flux distributed evenly among all subfilters.
     """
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -60,41 +60,46 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
             "If not set, skips calculating convergence and runs for ``maxNumIter`` iterations",
         default=True,
     )
+    useProgressiveGain = pexConfig.Field(
+        dtype=bool,
+        doc="Allow the relative weight of the new model solution to increase with each iteration?",
```
natelust: This sentence fragment does not make sense to end with a question mark. Either make it a proper statement or reword it to a proper question. I think our standard says the docs should be statements.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -60,41 +60,46 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
...
-    regularizeSigma = pexConfig.Field(
+    modelClampSigma = pexConfig.Field(
         dtype=float,
-        doc="Threshold to exclude noise-like pixels from regularization.",
-        default=3.,
-    )
-    clampFrequency = pexConfig.Field(
-        dtype=float,
-        doc="Maximum relative change of the model allowed between subfilters.",
+        doc="Threshold to exclude noise-like pixels from regularization between iterations.",
```
natelust: Is the threshold positive or negative, ie are pixels above or below the threshold excluded.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -60,41 +60,46 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
...
         default=2.,
     )
-    maxGain = pexConfig.Field(
+    frequencyClampSigma = pexConfig.Field(
         dtype=float,
-        doc="Maximum convergence-weighted gain to apply between forward modeling iterations.",
-        default=2.,
+        doc="Threshold to exclude noise-like pixels from regularization across subfilters.",
```
natelust: same as above

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -60,41 +60,46 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
...
+        default=1.,
     )
-    minGain = pexConfig.Field(
+    frequencyClampFactor = pexConfig.Field(
         dtype=float,
-        doc="Minimum convergence-weighted gain to apply between forward modeling iterations.",
-        default=0.5,
+        doc="Maximum relative change of the model allowed between subfilters. Set to zero to disable.",
```
natelust: relative change in what units

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -537,14 +548,18 @@ def newModelFromResidual(self, dcrModels, residualGeneratorList, bbox, statsFlag
             Statistics control object for coadd
         weightList : `list` of `float`
             The weight to give each input exposure in the coadd
+        mask : `lsst.afw.image.Mask`
+            Mask of the initial template coadd.
```
natelust: This doc comment does not explain much, consider being more descriptive

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -554,10 +569,25 @@ def newModelFromResidual(self, dcrModels, residualGeneratorList, bbox, statsFlag
             # `MaskedImage`s only support in-place addition, so rename for readability
             residual += model[bbox]
             newModel = residual
-            dcrModels.clampModel(subfilter, newModel, bbox, statsCtrl, self.config.regularizeSigma,
-                                 self.config.modelClampFactor, self.config.convergenceMaskPlanes)
-            dcrModels.conditionDcrModel(subfilter, newModel, bbox, gain=1.)
+            # Catch any invalid values
+            badPixels = ~np.isfinite(newModel.image.array)
+            newModel.image.array[badPixels] = model[bbox].image.array[badPixels]
```
natelust: Is it better to replace these values, or carry around a mask and mask them out of further calculations

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -702,5 +737,29 @@ def fillCoadd(self, dcrModels, skyInfo, tempExpRefList, weightList, calibration=
...
+        ----------
+        modelIter : `int`
+            The current iteration of forward modeling.
+
+        Returns
+        -------
+        gain : `float`
+            Relative weight to give the new solution when updating the model.
+        """
+        if self.config.useProgressiveGain:
```
natelust: the particular values of these numbers do not seem well motivated, can you add a comment explaining why these magic numbers and not others? Or include a link to a paper or something where they come from

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -702,5 +737,29 @@ def fillCoadd(self, dcrModels, skyInfo, tempExpRefList, weightList, calibration=
...
+            baseGain = 0.5
+            if modelIter < 2:
+                return baseGain
```
natelust: baseGain is only used here, you could just return 0.5 or only define it inside this block if you want it named, but naming it does not seem to explain why the value is 0.5 any more so maybe err to less lines of code

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -702,5 +737,29 @@ def fillCoadd(self, dcrModels, skyInfo, tempExpRefList, weightList, calibration=
...
+            else:
+                return np.log(modelIter)
+        else:
+            return 1.
```
natelust: This whole function could probably be:
```
if self.config.useProgressiveGain:
    return 0.5 if modelIter <2 else np.log(modelIter)
return 1
```

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: This parameter has actually become unnecessary with the changes on this ticket, so I've simply removed it.

PR #223, opened 2018-10-02
DM-14580: Create tests for BestSeeingWcsSelectImagesTask

ebellm:


tests/test_BestSeeingWcsSelectImages.py
```diff
@@ -0,0 +1,161 @@
...
+        result = task.runDataRef(self.dataRef, self.coordList, selectDataList=self.selectList)
+        self.assertEqual(len(result.exposureInfoList), 3)
+
+    def testSelectDefaultMakeDatarRefListIsFalse(self):
+        self.localSetUp(mockFWHMs=[self.avgFwhm for i in range(self.config.nImagesMax)])
+        task = selectImages.BestSeeingWcsSelectImagesTask(config=self.config)
+        result = task.runDataRef(self.dataRef, self.coordList,
+                                 selectDataList=self.selectList, makeDataRefList=False)
+        self.assertIsNone(result.dataRefList)
+
```
mrawls: These test cases are marginally interesting, but what I'd really like to see is a slightly more realistic case where more than Nmax images meet the good seeing criterion but have different seeings. For example, if there are 10 images with acceptable FWHM values, but they aren't all the same, and Nmax = 3, I think the desired behavior would be for the selector to choose the three with the lowest FWHM values. This requires testing more than just the length of the resulting exposure list, but rather what FWHM values and/or dataIds made it into the list.

tests/test_BestSeeingWcsSelectImages.py
```diff
@@ -0,0 +1,161 @@
...
+        self.config.minPsfFwhm = 1.0
+        self.config.maxPsfFwhm = 1.5
+        self.config.nImagesMax = 10
+        self.localSetUp(mockFWHMs=[0.5, 1.0, 1.25, 1.5, 2.0])
+        task = selectImages.BestSeeingWcsSelectImagesTask(config=self.config)
+        result = task.runDataRef(self.dataRef, self.coordList, selectDataList=self.selectList)
+        self.assertEqual(len(result.exposureInfoList), 3)
+
+    def testSelectDefaultMakeDatarRefListIsFalse(self):
```
mrawls: Typo in the name (your data is (are?) pirate flavored... Datarrrrrr...)

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -400,7 +400,7 @@ def runDataRef(self, dataRef, coordList, makeDataRefList=True,
         if selectDataList is None:
             selectDataList = []
 
-        result = super().runDataRef(dataRef, coordList, makeDataRefList, selectDataList)
+        result = super().runDataRef(dataRef, coordList, makeDataRefList=True, selectDataList=selectDataList)
```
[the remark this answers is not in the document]
ebellm: We need to allow `BestSeeingWcsSelectImagesTask` to have `makeDataRefList = False`.  But the `BestSeeingWcsSelectImagesTask` relies on having the `dataRefList` returned by the subclassed `WcsImagesTask`, so we have to set `makeDataRefList = True` in the call to `super`.

PR #227, opened 2018-10-19
DM-15635: Approximate filter throughput for DcrCoadds

isullivan:


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -76,6 +77,17 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
         doc="Weight exposures by airmass? Useful if there are relatively few high-airmass observations.",
         default=True,
     )
+    modelWidth = pexConfig.Field(
+        dtype=int,
+        doc="Width of the region around detected sources to include in the DcrModel."
+            "Set to a negative number to disable, which will include all pixels.",
```
mrawls: Unless you care to convince me otherwise, I'd prefer an error to be raised if a nonsensical modelWidth < 0 is chosen. Later it uses a negative value to set all the weights to 1, which seems dangerous; wouldn't throwing an exception be safer? It's also not clear to me why 3 is a good default.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -76,6 +77,17 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
...
+        default=3,
+    )
+    useSubfilterMidpoint = pexConfig.Field(
```
mrawls: Could this be called `useMidpoint` like in ip_diffim? (Or, alternately, could the one in ip_diffim be called `useSubfilterMidpoint`? Mostly I'm seeking consistency!)

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -762,6 +779,30 @@ def calculateGain(self, modelIter, baseGain=1.):
...
+            The input masked image to calculate weights for.
+
+        Returns
+        -------
+        weights : `numpy.ndarray` or `float`
+            A 2D array of weight values that tapers smoothly to zero away from detected sources.
+            Or a placeholder value of 1.0 if ``self.config.modelWidth`` is set to a negative number.
+        """
+        if self.config.modelWidth < 0:
+            return 1.
```
mrawls: Again, I don't like this sneaky behavior. If you want a fallback option to have all the weights be 1, that should be an explicit option, not something that happens if the width is set negative.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: I've switched the name to `splitSubfilters`, and used it consistently throughout.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: That's a good point. I'll make it raise an exception on negative values, and will add a separate config parameter to turn it on or off.

PR #234, opened 2018-11-08
DM-16347: Fix modelWeights bugs.

isullivan:
Some cases errored with mis-matched bounding boxes, and assembling a coadd returned early if the new option `useModelWeights` was turned off.

### DIFF
--- a/python/lsst/pipe/tasks/dcrAssembleCoadd.py
+++ b/python/lsst/pipe/tasks/dcrAssembleCoadd.py
@@ -138,6 +138,13 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
 class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
     """Assemble DCR coadded images from a set of warps.
 
+    Attributes
+    ----------
+    bufferSize : `int`
+        The number of pixels to grow each subregion by to allow for DCR.
+    warpCtrl : `lsst.afw.math.WarpingControl`
+        Configuration settings for warping an image
+
     Notes
     -----
     As with AssembleCoaddTask, we want to assemble a coadded image from a set of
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -812,9 +812,12 @@ def calculateModelWeights(self, maskedImage):
         """
         if self.config.modelWeightsWidth < 0:
             raise ValueError("modelWeightsWidth must not be negative if useModelWeights is set")
-        convergeMask = maskedImage.mask.getPlaneBitMask(self.config.convergenceMaskPlanes)
-        convergeMaskPixels = maskedImage.mask.array & convergeMask > 0
-        weights = np.zeros_like(maskedImage.image.array)
+        bboxGrow = afwGeom.Box2I(bbox)
+        bboxGrow.grow(self.bufferSize)
+        bboxGrow.clip(dcrModels.bbox)
```
laurenam: It took some back-and-forthing to figure out the need for and significance of `bboxGrow`.  For the latter, some documentation or comment would be nice.  For the former, I see that this weight image is used in the context of a similarly `bboxGrow`n image in the `dcrAssembleSubregion()` function at L521.  Given that these are required to be the same, it is a bit uncomfortable that they are computed independently in each function.  Might it be useful to set `bboxGrow` in run itself and pass this directly into both functions (`calculateModelWeights()` and `dcrAssembleSubregion()`, and I do realize the latter will also need the subs for the `dcrModels.assign` call at L528, but it still may be worthwhile for clarity)?  That way the place to document it would naturally be part of the **Parameters** lists.

Regardless, you also need to update the docs for this function.  It would also be nice to have a bit more info on what the weights here actually represent (just calling them [model] weight values that tapers smoothly to zero away from detected sources" is not entirely illuminating...but that is out of scope for this ticket, so is up to you!)

isullivan: Good suggestions! I have renamed `bboxGrow` to `dcrBBox` and moved the definition up to the `run` method, so that it can be passed in to `dcrAssembleSubregion` and `calculateModelWeights`. Then I can remove the duplicated from both those places.

I also updated the documentation.

PR #235, opened 2018-11-09
DM-16468: Use measured convergence to predict gain

isullivan:
The iterative forward modeling in `DcrAssembleCoadd` requires a gain term to weight the new solution and prevent oscillations. The correct gain can be difficult to guess for a user, but can be calculated from the measured values of convergence of the model using a simple Kalman filter, which is added in this ticket.

### DIFF
--- a/python/lsst/pipe/tasks/dcrAssembleCoadd.py
+++ b/python/lsst/pipe/tasks/dcrAssembleCoadd.py
@@ -63,13 +63,19 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
     )
     baseGain = pexConfig.Field(
         dtype=float,
-        doc="Relative weight to give the new solution when updating the model."
-            "A value of 1.0 gives equal weight to both solutions.",
-        default=1.,
+        doc="Relative weight to give the new solution vs. the last solution when updating the model."
+            "A value of 1.0 gives equal weight to both solutions."
+            "Small values imply slower convergence of the solution, but can "
+            "help prevent overshooting and failures in the fit."
+            "If ``baseGain`` is None, a conservative gain "
+            "will be calculated from the number of subfilters. ",
+        default=None,
     )
     useProgressiveGain = pexConfig.Field(
         dtype=bool,
-        doc="Use a gain that slowly increases above ``baseGain`` to accelerate convergence?",
+        doc="Use a gain that slowly increases above ``baseGain`` to accelerate convergence? "
+        "When calculating the next gain, we use up to 5 previous gains and convergence values."
+        "Can be set to False to force the model to change at the rate of ``baseGain``. ",
         default=True,
     )
     doAirmassWeight = pexConfig.Field(
@@ -377,11 +383,12 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
                                                           stats.ctrl)
             self.log.info("Initial convergence : %s", convergenceMetric)
             convergenceList = [convergenceMetric]
+            gainList = []
             convergenceCheck = 1.
             subfilterVariance = None
             while (convergenceCheck > self.config.convergenceThreshold or
                    modelIter < self.config.minNumIter):
-                gain = self.calculateGain(modelIter)
+                gain = self.calculateGain(convergenceList, gainList)
                 self.dcrAssembleSubregion(dcrModels, subBBox, dcrBBox, tempExpRefList, imageScalerList,
                                           weightList, spanSetMaskList, stats.flags, stats.ctrl,
                                           convergenceMetric, baseMask, subfilterVariance, gain,
@@ -779,7 +786,7 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
             dcrCoadds.append(coaddExposure)
         return dcrCoadds
 
-    def calculateGain(self, modelIter):
+    def calculateGain(self, convergenceList, gainList):
         """Calculate the gain to use for the current iteration.
 
         After calculating a new DcrModel, each value is averaged with the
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -63,13 +63,19 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
...
         dtype=float,
-        doc="Relative weight to give the new solution when updating the model."
-            "A value of 1.0 gives equal weight to both solutions.",
-        default=1.,
+        doc="Relative weight to give the new solution vs. the last solution when updating the model."
+            "A value of 1.0 gives equal weight to both solutions."
+            "If ``baseGain`` is None, a conservative gain "
+            "will be calculated from the number of subfilters. "
+            "Small values imply slower convergence of the solution, but can "
+            "help prevent overshooting and failures in the fit.",
```
parejkoj: Swap this and the previous sentence.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -392,6 +399,12 @@ def run(self, skyInfo, tempExpRefList, imageScalerList, weightList,
                                                                   spanSetMaskList,
                                                                   stats.ctrl)
                     convergenceCheck = (convergenceList[-1] - convergenceMetric)/convergenceMetric
+                    if convergenceCheck < 0:
+                        self.log.warn("Coadd %s diverged before reaching maximum iterations or"
+                                      " desired convergence improvement of %s."
+                                      " Divergence: %s",
+                                      subBBox, self.config.convergenceThreshold, convergenceCheck)
```
parejkoj: Is this a "warning", or "error" state? If it diverges, should the code raise an exception, or can things further on continue?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -785,19 +798,71 @@ def calculateGain(self, modelIter):
 
         Parameters
         ----------
-        modelIter : `int`
-            The current iteration of forward modeling.
+        convergenceList : `list` of `float`
+            The quality of fit metric from each previous iteration.
+        gainList : `list` of `float`
+            The gains used in each previous iteration: appended with new gain
+            value. Gains are numbers between ``self.config.baseGain`` and 1.
```
parejkoj: "appended with the new gain value"

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -785,19 +798,71 @@ def calculateGain(self, modelIter):
...
+        ValueError
+            If ``len(convergenceList) != len(gainList)+1``.
         """
-        if self.config.useProgressiveGain:
-            iterGain = np.log(modelIter)*self.config.baseGain if modelIter > 0 else self.config.baseGain
-            return max(self.config.baseGain, iterGain)
-        return self.config.baseGain
+        nIter = len(convergenceList)
+        if nIter != len(gainList) + 1:
+            raise ValueError("convergenceList must be one element longer than gainList.")
```
parejkoj: Oh, we should have put the values in this error, something like: `"(convergenceList (%d) ... gainList (%d)"%(len, len)`

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -785,19 +798,71 @@ def calculateGain(self, modelIter):
...
+            # To calculate the best gain to use, compare the past gains that have been used
+            # with the resulting convergences to estimate the best gain to use.
+            # Algorithmically, this is a Kalman filter.
+            # If forward modeling proceeds perfectly, the convergence metric should
+            # asymptotically approach a final value.
+            # We can estimate that value from the measured changes in convergence
+            # weighted by the gains used in each previous iteration.
+            estFinalConv = [((1 + gainList[i])*convergenceList[i + 1] - convergenceList[i])/gainList[i]
+                            for i in range(nIter - 1)]
+            # WORDS about why negative is bad
```
parejkoj: Oh, we need WORDS here!

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -785,19 +798,71 @@ def calculateGain(self, modelIter):
...
+            estFinalConv = np.array(estFinalConv)
+            estFinalConv[estFinalConv < 0] = 0
+            # Because the estimate may slowly change over time, only use the most recent measurements.
+            estFinalConv = np.median(estFinalConv[max(nIter - 5, 0):])
+            lastGain = gainList[nIter - 2]
```
parejkoj: Why not `gainList[-1]` to make it obvious it's the last value of the list?

Similarly for the other two `convergenceList[-2]` and `convergenceList[-1]`?

tests/test_dcrAssembleCoadd.py
```diff
@@ -0,0 +1,134 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import unittest
+import unittest.mock
+
+import lsst.utils.tests
+
+import lsst.afw.image
+import lsst.daf.persistence
+from lsst.pipe.tasks.dcrAssembleCoadd import DcrAssembleCoaddTask, DcrAssembleCoaddConfig
```
parejkoj: We don't need `unittest.mock`, `lsst.afw.image`, or `lsst.daf.persistence` here. I'm not sure why flake8 didn't complain.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: I believe it should be just a warning, because the model will typically have converged well overall and should be fine to use for making templates for image differencing.

PR #236, opened 2018-11-16
DM-16082: Fix compression-option warnings in coadds test.

TallJimbo:


python/lsst/pipe/tasks/mocks/simpleMapper.py
```diff
@@ -95,6 +95,28 @@ def makeButlerLocation(cls, path, dataId, mapper, suffix=None, storage=None):
         if suffix is None:
             loc = super(ExposurePersistenceType, cls).makeButlerLocation(path, dataId, mapper, suffix=None,
                                                                          storage=storage)
+            # Write options are never applicable for _sub, since that's only
+            # for read None of the values aside from the "NONE"s matter, but
```
parejkoj: Is there missing punctuation between `read` and `None` here?

python/lsst/pipe/tasks/mocks/simpleMapper.py
```diff
@@ -95,6 +95,28 @@ def makeButlerLocation(cls, path, dataId, mapper, suffix=None, storage=None):
...
+            # writing explicit meaningless values for all of them to appease
+            # afw is the price we pay for trying to write a non-CameraMapper
+            # Mapper. It'll all bet better with Gen3 (TM).
```
parejkoj: Is this worth a `TODO:` with a ticket number to remind us to clean it up after Gen3?

TallJimbo: I don't think so; the whole concept of a mapper will be going away, so I don't think there's any way we could end up with the same problem there.

PR #237, opened 2018-11-21
DM-16537: Rework exception handling and add tests.

parejkoj:
We can continue processing if individual exposures are missing, but we should raise other exceptions up the chain.


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -319,7 +319,7 @@ def runDataRef(self, patchRef, selectDataList=[]):
             if any(exps.values()):
                 dataRefList.append(tempExpRef)
             else:
-                self.log.warn("Warp %s could not be created", tempExpRef.dataId)
+                self.log.error("Warp %s could not be created", tempExpRef.dataId)
```
hsinfang: FWIW, this warning has been seen pretty frequently in the HSC-RC2 reprocessing. One case is despite the ccd area overlaps with the sky patch, there are no good pixels in that patch (e.g. pixels masked out as bad).   In a way it's not really an "error" because we could expect it and we chose not to write out a warp file in that case.  In fact I'd argue for writing out a (~empty) file; that would make the execution workflow easier in the future middleware and production system.

hsinfang: An old example with illustration plots can be found [on this page](https://confluence.lsstcorp.org/display/~hchiang2/Collection+of+edge+cases)   (sorry -- rotate the images by 180 deg in your brain and that's the lower-left vertical ccd in the sky)

parejkoj: Thanks for the details. I'll make this back into a `warn`, but this suggests that we should do a better job managing this output, so that we either don't warn (if there's a good reason the warp wasn't made) or error/fail (if the warp should have been made but wasn't).

I've made a ticket for this: https://jira.lsstcorp.org/browse/DM-16591

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -363,19 +363,16 @@ def run(self, calexpRefList, skyInfo, visitId=0):
...
-                # like calexps that don't need the tract, and necessary for meas_mosaic outputs,
+                # like calexps that don't need the tract, and necessary for jointcal outputs,
                 # which do.
                 calExpRef = calExpRef.butlerSubset.butler.dataRef("calexp", dataId=calExpRef.dataId,
                                                                   tract=skyInfo.tractInfo.getId())
                 calExp = self.getCalibratedExposure(calExpRef, bgSubtracted=self.config.bgSubtracted)
-            except Exception as e:
-                self.log.warn("Calexp %s not found; skipping it: %s", calExpRef.dataId, e)
+            except MissingExposureError as e:
+                self.log.warn("Skipping missing data: %s", e)
```
hsinfang: Why no longer include dataId? I would imagine the dataId of the missing calexp useful for debugging, though such information may be obtained through other means.

python/lsst/pipe/tasks/makeCoaddTempExp.py
hsinfang: Thanks for filing the ticket!

python/lsst/pipe/tasks/makeCoaddTempExp.py
parejkoj: The assumption (and implementation) here is that the raised exception contains all the necessary information. Previously, the dataId was being duplicated.

yalsayyad: The only error from `getCalibratedExposure` that will include the `dataId` are failed butler calls.  Anything else goes wrong and we won't have the dataId logged. 

 Note: I'm moving all the Gen2 dataRef's out of `run` in the next month anyway.

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -363,19 +363,16 @@ def run(self, calexpRefList, skyInfo, visitId=0):
         for calExpInd, calExpRef in enumerate(calexpRefList):
             self.log.info("Processing calexp %d of %d for this Warp: id=%s",
                           calExpInd+1, len(calexpRefList), calExpRef.dataId)
-            try:
-                ccdId = calExpRef.get("ccdExposureId", immediate=True)
-            except Exception:
-                ccdId = calExpInd
+            ccdId = calExpRef.get("ccdExposureId", immediate=True)
```
yalsayyad: `obs_comCam` doesn't have a `bypass_ccdExposureId`, but that's not likely to survive.   Your change is now requiring any new obs_package to write a `bypass_ccdExposureId` in order to be able to make a warp. I thought there MIGHT be some notebooks out there that make mock dataRefs.

The conversion to Gen 3 is going to move this to runDataRef anyway.

parejkoj:
The ongoing conversation here is relevant to this ticket: https://community.lsst.org/t/warnings-or-other-not-errors-in-the-stack/3452/4

PR #247, opened 2018-12-17
tickets/DM-16845

natelust:
Fix schema name in butler lookup

The name of the data product for the input schema was wrong. Fixed
the data product name.


parejkoj:
I guess this is fine, though I'm surprised that it worked before without that!

PR #249, opened 2018-12-19
tickets/DM-15845

natelust:
Convert MergeDetectionsTask to a PipelineTask

### DIFF
--- a/python/lsst/pipe/tasks/mergeDetections.py
+++ b/python/lsst/pipe/tasks/mergeDetections.py
@@ -31,11 +31,12 @@ import lsst.afw.table as afwTable
 
 from lsst.meas.algorithms import SkyObjectsTask
 from lsst.pex.config import Config, Field, ListField, ConfigurableField, ConfigField
-from lsst.pipe.base import CmdLineTask
+from lsst.pipe.base import (CmdLineTask, PipelineTask, PipelineTaskConfig, InitOutputDatasetField,
+                            InputDatasetField, InitInputDatasetField, OutputDatasetField, Struct)
 from lsst.pipe.tasks.coaddBase import getSkyInfo
 
 
-class MergeDetectionsConfig(Config):
+class MergeDetectionsConfig(PipelineTaskConfig):
     """!
     @anchor MergeDetectionsConfig_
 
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/mergeDetections.py
```diff
@@ -169,24 +212,57 @@ def __init__(self, butler=None, schema=None, **kwargs):
 
         @param[in] schema     the schema of the detection catalogs used as input to this one
         @param[in] butler     a butler used to read the input schema from disk, if schema is None
+        @param[in] initInputs This a PipelineTask only argument that holds all inputs passed in
```
TallJimbo: `This a PipelineTask only` -> `A PipelineTask-only`

Long-term (i.e. when we remove `CmdLineTask`) we'll want to make this documentation less transitional (including making it numpydoc), but this makes sense for now.

python/lsst/pipe/tasks/mergeDetections.py
```diff
@@ -169,24 +212,57 @@ def __init__(self, butler=None, schema=None, **kwargs):
...
+            skyMap=skyMap,
+            tractInfo=tractInfo,
+            patchInfo=patchInfo,
+            wcs=tractInfo.getWcs(),
+            bbox=patchInfo.getOuterBBox()
+        )
+        inputData['skyInfo'] = skyInfo
+        return self.run(**inputData)
+
+    def run(self, catalogs, skyInfo, idFactory, skySeed):
```
TallJimbo: This looks like a change that will break `pipe_drivers`.  Do you have a branch for that, too?

python/lsst/pipe/tasks/mergeDetections.py
```diff
@@ -169,24 +212,57 @@ def __init__(self, butler=None, schema=None, **kwargs):
...
+        tractNumber = inputDataIds['catalogs'][0]['tract']
+        tractInfo = skyMap[tractNumber]
+        patchInfo = tractInfo.getPatchInfo(inputDataIds['catalogs'][0]['patch'])
+        skyInfo = Struct(
+            skyMap=skyMap,
+            tractInfo=tractInfo,
+            patchInfo=patchInfo,
+            wcs=tractInfo.getWcs(),
+            bbox=patchInfo.getOuterBBox()
+        )
```
TallJimbo: FYI I've just created a ticket that might eventually help with this: https://jira.lsstcorp.org/browse/DM-16919, in that I'd like to just let you declare something like `PatchInfo` as an input Dataset, and you'd get the `PatchInfo` you want directly instead of having to read the SkyMap and unpack the data ID.  That won't help immediately if you need more than a `PatchInfo` here, but I think a `PatchInfo` should probably hold a reference to its own `TractInfo` and `SkyMap` if it doesn't already, and if we did that we wouldn't need the `SkyInfo` struct to be more than a `PatchInfo`.

python/lsst/pipe/tasks/mergeDetections.py
```diff
@@ -57,17 +58,59 @@ class MergeDetectionsConfig(Config):
...
+    )
+
     def setDefaults(self):
         Config.setDefaults(self)
+        self.formatTemplateNames(self.coaddName)
         self.skyObjects.avoidMask = ["DETECTED"]  # Nothing else is available in our custom mask
+        self.quantum.dimensions = ("Tract", "Patch", "SkyMap")
 
     def validate(self):
         Config.validate(self)
```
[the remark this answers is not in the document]
natelust: Done

python/lsst/pipe/tasks/mergeDetections.py
```diff
@@ -57,17 +58,59 @@ class MergeDetectionsConfig(Config):
                              doc="Priority-ordered list of bands for the merge.")
     coaddName = Field(dtype=str, default="deep", doc="Name of coadd")
 
+    schema = InitInputDatasetField(
+        doc="Schema of the input detection catalog",
+        name="",
```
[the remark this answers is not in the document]
natelust: Fixed in 16797 and updated here

python/lsst/pipe/tasks/mergeDetections.py
natelust: I put in a code comment above the docstring to change at a later time

python/lsst/pipe/tasks/mergeDetections.py
natelust: Pipe_drivers calls runDataRef, so it should be fine

PR #251, opened 2018-12-21
Add config parameter includeCalibVar

yalsayyad:
to set whether photometric calibration variance is added to the warp variance.
False by default


tests/test_makeCoaddTempExp.py
```diff
@@ -96,28 +96,34 @@ def test_getCalibratedExposureGetCalexpRaises(self):
         self.assertIn('Exposure not found', str(cm.exception))
 
     def test_getCalibratedExposure(self):
-        task = MakeCoaddTempExpTask(config=self.config)
+        for includeCalibVar in [True, False]:
```
parejkoj: I'm not keen on test for loops like this: they make it harder to debug failing tests. Especially when there are only two states, it's much nicer to take what you've made the internals of the loop and turn it into a method (`_checkCalibratedExposure`?), and then just call that twice with the two states.

yalsayyad: Agreed. How about 4 states? (Check out the latest unit test. This what you had in mind?)

parejkoj:
Nice, yeah, that looks good.

PR #255, opened 2019-01-16
tickets/DM-16873

natelust:


python/lsst/pipe/tasks/multiBand.py
```diff
@@ -819,6 +898,83 @@ def __init__(self, butler=None, schema=None, peakSchema=None, refObjLoader=None,
...
+            inputCatalogsToKeep = []
+            inputCatalogWcsUpdate = []
+            for (i, dataId) in enumerate(inputDataIds['visitCatalogs']):
+                key = (dataId['visit'], dataId['detector'])
+                if (key) in inputVisitIds:
+                    inputCatalogsToKeep.append(inputData['visitCatalogs'][i])
+                    inputCatalogWcsUpdate.append(ccdRecordsWcs[key])
+            inputData['visitCatalogs'] = inputCatalogsToKeep
+            inputData['wcsUpdates'] = inputCatalogWcsUpdate
+            inputData['ccdInputs'] = ccdInputs
```
TallJimbo: Hopefully we can clean this up more fully when we don't have to support the `CmdLineTask` versions of things anymore, but do you think it would be an improvement now to bundle these three arguments into (e.g.) a `namedtuple` or something.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -834,8 +990,20 @@ def runDataRef(self, patchRef, psfCache=100):
...
+                           skyInfo=skyInfo, butler=patchRef.getButler(),
+                           exposureId=self.getExposureId(patchRef))
+
+        if self.config.doMatchSources:
+            self.writeMatches(patchRef, results)
+        self.write(patchRef, results.outputSources)
 
-        self.measurement.run(sources, exposure, exposureId=self.getExposureId(patchRef))
+    def run(self, exposure, sources, skyInfo, exposureId, ccdInputs=None, visitCatalogs=None, wcsUpdates=None,
+            butler=None):
```
TallJimbo: I know we're just trying to get things done quickly right now and plan to clean up in the future, but adding parameter docs for this method might be worth doing on this ticket, especially for anyone who might need to make changes to the Gen2 version before we retire it.

python/lsst/pipe/tasks/propagateVisitFlags.py
```diff
@@ -165,28 +166,23 @@ def run(self, butler, coaddSources, ccdInputs, coaddWcs):
         @param[in,out] coaddSources  Source catalog from the coadd
         @param[in] ccdInputs  Table of CCDs that contribute to the coadd
         @param[in] coaddWcs  Wcs for coadd
+        @param[in] visitCatalogs List of loaded source catalogs for each input ccd in
+                                 the coadd, this is used instead of this method loading
+                                 in the catalogs itself
```
TallJimbo: -> "the coadd. *If provided*, this is used instead of this method loading in the catalogs itself."

python/lsst/pipe/tasks/propagateVisitFlags.py
```diff
@@ -165,28 +166,23 @@ def run(self, butler, coaddSources, ccdInputs, coaddWcs):
...
         indices = numpy.array([s.getId() for s in coaddSources])  # Allowing for non-contiguous data
+        radius = self.config.matchRadius*afwGeom.arcseconds
 
-        # Accumulate counts of flags being set
-        for ccdRecord in ccdInputs:
-            v = ccdRecord.get(visitKey)
-            c = ccdRecord.get(ccdKey)
-            dataId = {"visit": int(v), self.config.ccdName: int(c)}
-            ccdSources = butler.get("src", dataId=dataId, immediate=True)
+        def process_ccd(ccdSources, wcsUpdate):
```
TallJimbo: -> camelCase

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -819,6 +898,83 @@ def __init__(self, butler=None, schema=None, peakSchema=None, refObjLoader=None,
...
+    def getInitOutputDatasets(self):
+        return {"outputSchema": afwTable.SourceCatalog(self.schema)}
+
+    def adaptArgsAndRun(self, inputData, inputDataIds, outputDataIds, butler):
+        refObjLoader = ReferenceObjectLoader(inputDataIds['refCat'], butler,
+                                             config=self.config.refObjLoader, log=self.log)
+        self.match.setRefObjLoader(refObjLoader)
+
+        # Set psfcache
+        inputData['exposure'].getPsf().setCacheCapacity(self.config.psfCache)
```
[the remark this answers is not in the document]
natelust: In general yes this sort of thing could go in run. However in this case, runDataRef has a specific interface for this that I didn't want to change, so it is here so I don't step on gen2 toes. I will add a comment to move it to run after gen2 deprecation.

python/lsst/pipe/tasks/multiBand.py
natelust: Since ccdInputs is in the interface for the existing propagate visit flags, this is something I dont want to bundle as it would just need to be unpacked later. That leaves 2 things which does not seem worth bundling. I think it is better to have an explicit function signature instead of packing things just to unpack them inside the function.

PR #260, opened 2019-01-23
Converted CharacterizeImageTask to pipelineTask.

czwa:
The updated version produces identical image output products, and
catalogs with only minor differences (id values differ; flux values
have 1e-5 level fractional differences).

### DIFF
--- a/python/lsst/pipe/tasks/characterizeImage.py
+++ b/python/lsst/pipe/tasks/characterizeImage.py
@@ -40,7 +40,37 @@ from .repair import RepairTask
 __all__ = ["CharacterizeImageConfig", "CharacterizeImageTask"]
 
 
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -40,7 +40,42 @@
...
 
 
-class CharacterizeImageConfig(pexConfig.Config):
+class CharacterizeImageConfig(pipeBase.PipelineTaskConfig):
+    # Gen3 IO options
+    taskName = pexConfig.Field(
+        dtype=str,
+        doc="Name of task",
+        default="charImage",
+    )
```
TallJimbo: I don't see this config option used below, and I didn't think something like this was necessary for PipelineTasks in general.  Does it need to exist?

PR #264, opened 2019-01-25
tickets/DM-17390 Update CalibrateTask to be a PipelineTask

natelust:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -444,6 +547,39 @@ def runDataRef(self, dataRef, exposure=None, background=None, icSourceCat=None,
...
+            photoRefObjLoader = ReferenceObjectLoader(inputDataIds['photoRefCat'],
+                                                      butler,
+                                                      self.config.photoRefObjLoader,
+                                                      self.log)
+            self.pixelMargin = photoRefObjLoader.config.pixelMargin
+            self.photoCal.match.setRefObjLoader(photoRefObjLoader)
+
+        results = self.run(**inputData)
+
+        if self.config.doWriteMatches and results.astromMatches is not None:
```
TallJimbo: Is it possible for the first clause here to be true while the second is false?  If so, that could be a problem in the future (because we're lying to the activator about what it should expect to see in the outputs).  If we can easily convince ourselves that `if self.config.doWriteMatches` alone is sufficient here now, that'd be ideal; if not, I'd like to at least add a warning if `results.astromMatrches` is unexpectedly `None` and open a ticket to follow up later.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -598,6 +734,10 @@ def run(self, exposure, exposureIdInfo=None, background=None,
             sourceCat=sourceCat,
             astromMatches=astromMatches,
             matchMeta=matchMeta,
+
```
TallJimbo: Just seems like a weird place for a newline.  I'd either drop it or replace it with comment headings if you want to define groups of output struct fields.

PR #265, opened 2019-01-25
DM-17412: Converted MergeMeasurementTask to PipelineTask.

czwa:
Runs with similar, but not exact, results to the Gen2 run.


TallJimbo:
I think it's more common (and maybe in the style guide?) for us to use present tense in git commit messages.

PR #270, opened 2019-02-05
DM-17689: Avoid magic numbers in IdFactory declarations.

TallJimbo:


python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
```diff
@@ -146,7 +147,10 @@ def __init__(self, initInputs, **kwargs):
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
         packedId, maxBits = butlerQC.quantum.dataId.pack("tract_patch", returnMaxBits=True)
```
timj, line 149: Do we still need this line?

timj (approved):

python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
TallJimbo: Good catch.

PR #272, opened 2019-02-07
DM-17564: Convert DcrModel to image type from maskedImage.

isullivan:
This simpler data model reduces execution time by a factor of ~3.

### DIFF
--- a/python/lsst/pipe/tasks/dcrAssembleCoadd.py
+++ b/python/lsst/pipe/tasks/dcrAssembleCoadd.py
@@ -25,7 +25,6 @@ import numpy as np
 from scipy import ndimage
 import lsst.afw.geom as afwGeom
 import lsst.afw.image as afwImage
-import lsst.afw.math as afwMath
 import lsst.coadd.utils as coaddUtils
 from lsst.ip.diffim.dcrModel import applyDcr, calculateDcr, DcrModel
 import lsst.pex.config as pexConfig
@@ -125,15 +124,10 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
         default=2,
         doc="Minimum radius of a region to include in regularization, in pixels."
     )
-    imageWarpMethod = pexConfig.Field(
-        dtype=str,
-        doc="Name of the warping kernel to use for shifting the image and variance planes.",
-        default="lanczos3",
-    )
-    maskWarpMethod = pexConfig.Field(
-        dtype=str,
-        doc="Name of the warping kernel to use for shifting the mask plane.",
-        default="bilinear",
+    imageInterpOrder = pexConfig.Field(
+        dtype=int,
+        doc="The order of the spline interpolation used to shift the image plane.",
+        default=3,
     )
 
     def setDefaults(self):
@@ -145,8 +139,6 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
         # The deepCoadd and nImage files will be overwritten by this Task, so don't write them the first time
         self.assembleStaticSkyModel.doNImage = False
         self.assembleStaticSkyModel.doWrite = False
-        self.sigmaClip = 10
-        self.statistic = 'MEAN'
 
 
 class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
@@ -156,8 +148,6 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
     ----------
     bufferSize : `int`
         The number of pixels to grow each subregion by to allow for DCR.
-    warpCtrl : `lsst.afw.math.WarpingControl`
-        Configuration settings for warping an image
 
     Notes
     -----
@@ -260,7 +250,7 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
     def prepareDcrInputs(self, templateCoadd, tempExpRefList, weightList):
         """Prepare the DCR coadd by iterating through the visitInfo of the input warps.
 
-        Sets the properties ``warpCtrl`` and ``bufferSize``.
+        Sets the property ``bufferSize``.
 
         Parameters
         ----------
@@ -304,13 +294,6 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
         self.log.info("Selected airmasses:\n%s", airmassList)
         self.log.info("Selected parallactic angles:\n%s", angleList)
         self.bufferSize = int(np.ceil(np.max(dcrShifts)) + 1)
-        # Turn off the warping cache, since we set the linear interpolation length to the entire subregion
-        # This warper is only used for applying DCR shifts, which are assumed to be uniform across a patch
-        warpCache = 0
-        warpInterpLength = max(self.config.subregionSize)
-        self.warpCtrl = afwMath.WarpingControl(self.config.imageWarpMethod,
-                                               self.config.maskWarpMethod,
-                                               cacheSize=warpCache, interpLength=warpInterpLength)
         dcrModels = DcrModel.fromImage(templateCoadd.maskedImage,
                                        self.config.dcrNumSubfilters,
                                        filterInfo=filterInfo,
@@ -388,8 +371,8 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
         stats = self.prepareStats(mask=badPixelMask)
         dcrModels = self.prepareDcrInputs(templateCoadd, tempExpRefList, weightList)
         if self.config.doNImage:
-            dcrNImages = self.calculateNImage(dcrModels, skyInfo.bbox,
-                                              tempExpRefList, spanSetMaskList, stats.ctrl)
+            dcrNImages, dcrWeights = self.calculateNImage(dcrModels, skyInfo.bbox, tempExpRefList,
+                                                          spanSetMaskList, stats.ctrl)
             nImage = afwImage.ImageU(skyInfo.bbox)
             # Note that this nImage will be a factor of dcrNumSubfilters higher than
             # the nImage returned by assembleCoadd for most pixels. This is because each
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -561,49 +555,43 @@ def dcrAssembleSubregion(self, dcrModels, bbox, dcrBBox, tempExpRefList, imageSc
...
+            templateImage = dcrModels.buildMatchedTemplate(exposure=exposure,
+                                                           order=self.config.imageInterpOrder,
                                                            splitSubfilters=self.config.splitSubfilters)
-            imageScaler.scaleMaskedImage(maskedImage)
-            if altMaskSpans is not None:
-                self.applyAltMaskPlanes(maskedImage.mask, altMaskSpans)
+            residual = exposure.image.array - templateImage.array
+            # Note that the variance plane here is used to store weights, not the actual variance
+            residual *= exposure.variance.array
+            residualGeneratorList.append(self.dcrResiduals(residual, visitInfo, wcs, dcrModels.filter))
```
parejkoj: `residualGeneratorList` sounds like a `generator` that produces `list`s or something?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -716,24 +694,19 @@ def calculateConvergence(self, dcrModels, bbox, tempExpRefList, imageScalerList,
...
-            singleMetric = self.calculateSingleConvergence(dcrModels, exposure, significanceImage, statsCtrl,
-                                                           altMaskSpans=altMaskSpans)
-            metric += singleMetric*expWeight
+        for tempExpRef, expWeight in zip(tempExpRefList, weightList):
+            exposure = subExposures[tempExpRef.dataId["visit"]][bbox]
+            singleMetric = self.calculateSingleConvergence(dcrModels, exposure, significanceImage, statsCtrl)
+            metric += singleMetric
             metricList[tempExpRef.dataId["visit"]] = singleMetric
-            weight += expWeight
+            weight += 1.
```
parejkoj: I'm confused: you're adding both `singleMetric` and `1` to `weight` here. Why both?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -974,5 +944,41 @@ def applyModelWeights(self, modelImages, refImage, modelWeights):
...
+        bbox : `lsst.afw.geom.box.Box2I`
+            Sub-region to coadd
+        statsCtrl : `lsst.afw.math.StatisticsControl`
+            Statistics control object for coadd
+        tempExpRefList : `list` of `lsst.daf.persistence.ButlerDataRef`
+            The data references to the input warped exposures.
+        imageScalerList : `list` of `lsst.pipe.task.ImageScaler`
+            The image scalars correct for the zero point of the exposures.
+        spanSetMaskList : `list` of `dict` containing spanSet lists, or None
+            Each element is dict with keys = mask plane name to add the spans to
```
parejkoj: Please expand the docstring for `spanSetMaskList` a bit: I'm having trouble parsing this.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -974,5 +944,41 @@ def applyModelWeights(self, modelImages, refImage, modelWeights):
...
+
+        Returns
+        -------
+        subExposures : `dict` of `lsst.afw.image.ExposureF`
+            The pre-loaded exposures for the current subregion.
```
parejkoj: `dict` of what to what?

parejkoj: And also explain here that you're using the variance plane as a weight plane instead.

I'm not sure I'm happy about it, but it does simplify the API so I guess it's ok?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -974,5 +944,41 @@ def applyModelWeights(self, modelImages, refImage, modelWeights):
...
+        subExposures = {}
+        for tempExpRef, imageScaler, altMaskSpans in zipIterables:
+            exposure = tempExpRef.get(tempExpName + "_sub", bbox=bbox)
+            if altMaskSpans is not None:
+                self.applyAltMaskPlanes(exposure.mask, altMaskSpans)
+            imageScaler.scaleMaskedImage(exposure.maskedImage)
+            # Note that the variance plane here is used to store weights, not the actual variance
+            exposure.variance.array[:, :] = 0.
+            exposure.variance.array[exposure.mask.array & statsCtrl.getAndMask() == 0] = 1.
+            exposure.image.array[exposure.mask.array & statsCtrl.getAndMask() > 0] = 0.
```
parejkoj: Why are these treated differently? You use `exposure.mask` to set both, but the opposite statsCtrl masks.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: It's a list of generators, and I agree that it is confusing but this seemed to be the most memory-efficient way to pass nSubfilter X nImage residual arrays. I can add an additional comment to explain it.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: `singleMetric` is being added to `metric`, not `weight`.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -499,25 +485,39 @@ def calculateNImage(self, dcrModels, bbox, tempExpRefList, spanSetMaskList, stat
...
-                dcrNImage.array[shiftedImage.mask.array & statsCtrl.getAndMask() == 0] += 1
-        return dcrNImages
-
-    def dcrAssembleSubregion(self, dcrModels, bbox, dcrBBox, tempExpRefList, imageScalerList, weightList,
-                             spanSetMaskList, statsFlags, statsCtrl, convergenceMetric,
-                             baseMask, gain, modelWeights, refImage):
+            for dcr, dcrNImage, dcrWeight in zip(dcrShift, dcrNImages, dcrWeights):
+                shiftedWeights = applyDcr(weightImage, dcr, useInverse=True,
+                                          order=self.config.maskInterpOrder)
+                dcrNImage.array += np.rint(shiftedWeights).astype(dcrNImage.array.dtype)
```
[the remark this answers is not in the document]
isullivan: Sadly it casts to the wrong type of `int` and throws an error. I'm not very happy with how this is implemented, so I'd welcome any suggestions you might have.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -499,25 +485,39 @@ def calculateNImage(self, dcrModels, bbox, tempExpRefList, spanSetMaskList, stat
...
         for tempExpRef, altMaskSpans in zip(tempExpRefList, spanSetMaskList):
             exposure = tempExpRef.get(tempExpName + "_sub", bbox=bbox)
+            # padValue = afwImage.pixel.SinglePixelF(0., exposure.mask.getPlaneBitMask("NO_DATA"), 0)
             visitInfo = exposure.getInfo().getVisitInfo()
             wcs = exposure.getInfo().getWcs()
             mask = exposure.mask
             if altMaskSpans is not None:
                 self.applyAltMaskPlanes(mask, altMaskSpans)
+            weightImage = np.zeros_like(exposure.image.array)
+            weightImage[mask.array & statsCtrl.getAndMask() == 0] = 1.
```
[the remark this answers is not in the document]
isullivan: I've added parenthesis to make the order of operations more clear: `(mask.array & statsCtrl.getAndMask()) == 0`

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -499,25 +485,39 @@ def calculateNImage(self, dcrModels, bbox, tempExpRefList, spanSetMaskList, stat
...
+        for subfilter in range(self.config.dcrNumSubfilters):
+            goodPix = dcrWeights[subfilter].array > 0
+            dcrNImages[subfilter].array[~goodPix] = 0
+            dcrWeights[subfilter].array[~goodPix] = 0.
+            dcrWeights[subfilter].array[goodPix] = 1./dcrWeights[subfilter].array[goodPix]
+        return (dcrNImages, dcrWeights)
+
+    def dcrAssembleSubregion(self, dcrModels, subExposures, bbox, dcrBBox, tempExpRefList,
+                             statsCtrl, convergenceMetric,
+                             baseMask, gain, modelWeights, refImage, dcrWeights):
```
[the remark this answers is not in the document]
isullivan: I've created a new ticket, DM-17882 to simplify the call signature.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: It's setting the weight of unmasked pixels to 1., and the value of masked pixels to 0. I've added parenthesis to make the order of operations clear, and a comment to explain what I'm doing to the two arrays.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -125,15 +124,15 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
...
+        default=3,
     )
-    maskWarpMethod = pexConfig.Field(
-        dtype=str,
-        doc="Name of the warping kernel to use for shifting the mask plane.",
-        default="bilinear",
+    maskInterpOrder = pexConfig.Field(
+        dtype=int,
+        doc="The order of the spline interpolation used to shift the mask plane",
+        default=1,
```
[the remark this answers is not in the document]
isullivan: I've removed `maskInterpOrder` entirely. The name and description were misleading, since it was actually used for the weights, not the mask plane. And, the weights really need to use the same interpolation as the image, so it shouldn't be configurable by itself.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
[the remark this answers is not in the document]
parejkoj: "wrong type of int"? What is the type of `dcrNImage`? And what is the error?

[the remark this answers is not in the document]
isullivan: The error if `.astype` is omitted is `TypeError: Cannot cast ufunc add output from dtype('float32') to dtype('uint16'), since `rint` returns the values rounded to the nearest integer as `float`s, but my comment about the "wrong type or int" came from setting `np.rint(...).astype(int)`, which raises `TypeError: Cannot cast ufunc add output from dtype('int64') to dtype('uint16') with casting rule 'same_kind'`.
So, I think I need to either leave this as it is, or add a completely different calculation of the nImage.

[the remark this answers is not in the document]
parejkoj: Huh. Yeah, lets leave it as is and we can revisit when we look at performance.

PR #280, opened 2019-03-01
DM-17427: objectMasks: update parser to handle Goulding mask format

PaulPrice:
Andy Goulding has provided new masks based on Gaia DR2, and with a
better tuned size; but the format is subtly different:
* The rotation angle of a 'box' is not specified (it’s supposed to be
  zero anyway).
* The 'ID:' string is replaced with 'Gaia DR2'.
* Floating point values use scientific notation in places.

We've decided that (apart from the lack of "ID:", which we want in
order to be explicit) it's best to update the parser rather than fix
the mask files: the differences from what we have been expecting are
reasonable, and fixing the parser now means we may not have to fix
it much for the next version.

Expanded the regex to support the new format, defaulting the
rotation angle of boxes to 0. Added comments to the regex so it's
easier to understand.

### DIFF
--- a/python/lsst/pipe/tasks/objectMasks.py
+++ b/python/lsst/pipe/tasks/objectMasks.py
@@ -158,16 +158,22 @@ class ObjectMaskCatalog:
                 # with the format as specified in the above docstring.
                 mat = re.search(r"^\s*(box|circle)"
                                 r"(?:\s+|\s*\(\s*)"   # open paren or space
-                                r"(\d+(?:\.\d*)?([d]*))" r"(?:\s+|\s*,\s*)"
-                                r"([+-]?\d+(?:\.\d*)?)([d]*)" r"(?:\s+|\s*,\s*)"
-                                r"([+-]?\d+(?:\.\d*)?)([d]*)" r"(?:\s+|\s*,\s*)?"
-                                r"(?:([+-]?\d+(?:\.\d*)?)([d]*)"
-                                r"\s*,\s*"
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/objectMasks.py
```diff
@@ -186,12 +192,13 @@ def read(fileName):
...
                     height = NaN
-                    angle = NaN
+                    angle = 0.0*afwGeom.degrees
 
                     if _type == "box":
                         width = convertToAngle(param1, param1Unit, "width", fileName, lineNo)
                         height = convertToAngle(param2, param2Unit, "height", fileName, lineNo)
-                        angle = convertToAngle(param3, param3Unit, "angle", fileName, lineNo)
+                        if param3 is not None:
+                            angle = convertToAngle(param3, param3Unit, "angle", fileName, lineNo)
```
natelust: You are basically doing work to optionally read in an angle if supplied but below there can only be an angle of zero, and it is only output as a warning. Should that be stronger than a warning as it is not going to do as the user requests? Or should we just take rotation out all the way?

PaulPrice: I went to change this to an error, but then noticed that it actually does end up throwing an error because of this. It's putting out warnings as it reads each line, and then if there were any problems it throws an error when it's done reading. I think failing late here is the right thing to do, as it allows the user to fix all the bad lines rather than one by one (nicely done, @RobertLuptonTheGood!).

PR #281, opened 2019-03-08
DM-17029: Update LoadReferenceObjectsTask to output fluxes in nanojansky

parejkoj:


python/lsst/pipe/tasks/photoCal.py
```diff
@@ -360,18 +360,20 @@ def extractMagArrays(self, matches, filterName, sourceKeys):
                               fluxField)
                 refFluxErrArr = np.sqrt(refFluxArr)
 
-            refMagArr = u.Quantity(refFluxArr, u.Jy).to_value(u.ABmag)
-            refMagErrArr = abMagErrFromFluxErr(refFluxErrArr, refFluxArr)
+            refMagArr = u.Quantity(refFluxArr, u.nJy).to_value(u.ABmag)
+            # HACK convert to Jy until we have a replacement for this (DM-16903)
+            refMagErrArr = abMagErrFromFluxErr(refFluxErrArr*1e-9, refFluxArr*1e-9)
```
SimonKrughoff: I'd prefer this be internal to `abMagErrFromFluxErr`.  That would make implementing DM-16903 easier since you wouldn't have to look for every place `abMagErrFromFluxErr` is used.

parejkoj: I'm torn on this one. There are fortunately very few uses of these methods, but would be an API change. I don't know if we can include it under RFC-575? I suppose with that RFC, all `flux` units dealt with in the stack will be `nJy`, so maybe it is appropriate to change these functions now?

It looks like these two magErr<->fluxErr functions are only used in a handful of places in the stack, so maybe it wouldn't affect all that much?

SimonKrughoff: I trust your judgement on this one.  I didn't appreciate the API change argument until you voiced it.  I agree there are few enough instances of those functions and it is easy enough to grep that it shouldn't be a big deal.

PR #287, opened 2019-04-30
Tickets/DM-18742: speed up DcrModel convergence

isullivan:
Add an option that amplifies the differences between DCR model planes to accelerate convergence. With default settings this is only applied within the footprints of detected sources, and only during the iterative forward modeling step of constructing the DcrModel.

### DIFF
--- a/python/lsst/pipe/tasks/dcrAssembleCoadd.py
+++ b/python/lsst/pipe/tasks/dcrAssembleCoadd.py
@@ -129,6 +129,11 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
         doc="The order of the spline interpolation used to shift the image plane.",
         default=3,
     )
+    accelerateModel = pexConfig.Field(
+        dtype=float,
+        doc="Factor to amplify the differences between model planes by to speed convergence.",
+        default=3,
+    )
 
     def setDefaults(self):
         CompareWarpAssembleCoaddConfig.setDefaults(self)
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -112,7 +112,7 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
         dtype=float,
         doc="Maximum relative change of the model allowed between subfilters."
             "Set to zero to disable.",
-        default=4.,
+        default=10.,
```
kfindeisen: Is this to compensate for the rescaling? It seems a bit arbitrary.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -796,7 +805,9 @@ def fillCoadd(self, dcrModels, skyInfo, tempExpRefList, weightList, calibration=
             the model for one subfilter.
         """
         dcrCoadds = []
+        refModel = dcrModels.getReferenceImage()
         for model in dcrModels:
+            model.array = (model.array - refModel)*self.config.accelerateModel + refModel
```
kfindeisen: Should this be conditional on `accelerateModel > 1`, like the similar code in `buildMatchedTemplate`?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: No, it's unrelated. I found the results improved very slightly with this change while I was testing frequency regularization. I can revert it if you think that would be cleaner.

PR #288, opened 2019-04-30
tickets/DM-19272 Add script and task to create skymap in gen3

natelust:


python/lsst/pipe/tasks/makeGen3SkyMap.py
```diff
@@ -0,0 +1,119 @@
...
+                                                       hash=skyMapHash)
+                raise RuntimeError(
+                    (f"SkyMap with name {existing.name} and hash {skyMapHash} already exist in "
+                     f"the butler collection {self.collection}, SkyMaps must be unique within "
+                     "a collection")
+                )
+            except ValueError:
+                self.log.info(f"Inserting SkyMap {self.config.skyMapName} with hash={skyMapHash}")
+                with self.butler.registry.transaction():
+                    skyMap.register(self.config.skyMapName, self.butler.registry)
```
hsinfang: It seems I'm not able to create one deepCoadd_skyMap  skymap with config1 and then create another deepCoadd_skyMap  skymap with config2, even with a different butler collection name. I would get `lsst.daf.butler.core.registry.ConflictingDefinitionError: Existing definition for skymap entry with {skymap: deepCoadd_skyMap}.` 

Is this intentional? Is the idea to change the skyMapName when a different config is used?

hsinfang: Hmm... it seems changing `skyMapName` doesn't change the hash; the hash is determined by other config parameters. 

Is it not allowed to make two skymaps to two collections using the same configs, for example, 
`makeGen3Skymap.py REPO collection1`  and then `makeGen3Skymap.py REPO collection2` ?

hsinfang:
Attempted to run this with oracle gave me ```sqlalchemy.exc.CompileError: The 'oracle' dialect with current database version settings does not support in-place multirow inserts.```   
It's very possible that I'm not setting up the Oracle registry correctly yet.  But were you able to do this with Oracle?

hsinfang:
The above `sqlalchemy.exc.CompileError` was with today's master `dat_butler` and `skymap`. If I instead used `w_2019_17` + this branch, it seemed to run through.  Sounds like something with DM-19271 optimization?

python/lsst/pipe/tasks/makeGen3SkyMap.py
hsinfang: In today's standup @TallJimbo clarified that a duplicate skymap of the same set of configs (determining the hash; `skyMapName` not included) cannot be ingested no matter the collections. That explains the `ConflictingDefinitionError` I saw. It seems okay to me for now, when we each have our own sqlite3 registry. I'm not sure how it will be when we share things on one Oracle database, but probably fine for now.

python/lsst/pipe/tasks/makeGen3SkyMap.py
```diff
@@ -0,0 +1,131 @@
...
+    skyMap = skyMapRegistry.makeField(
+        doc="type of skyMap",
+        default="dodeca",
+    )
+    doWrite = pexConfig.Field(
+        doc="persist the skyMap? If False then run generates the sky map and returns it, "
+            "but does not save it to the data repository",
+        dtype=bool,
+        default=True,
+    )
```
TallJimbo: I don't think we need this option - if someone just wanted to create a SkyMap from a config, they wouldn't really need to use this task.

python/lsst/pipe/tasks/makeGen3SkyMap.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Parameters
+    ----------
+    config : `MakeGen3SkyMapConfig` or None
+        Instance of a configuration class specifying task options, a default
+        config is created if value is None
+    butler : `str`
+        Path to a gen3 butler
+    collection : `str`
+        The name of the collection which the created skymap will be inserted
```
TallJimbo: I think it's probably better to accept a `Butler` instance in `run`, and actually construct it from a path and collection string in the `bin` script - I envision the shared logic for constructing a butler from command-line arguments to eventually move to common place, and when it does, that will reduce the need for changes to this task.

python/lsst/pipe/tasks/makeGen3SkyMap.py
```diff
@@ -0,0 +1,131 @@
...
+                    self.butler.put(skyMap, self.config.datasetTypeName, {"skymap": self.config.name})
+
+        return pipeBase.Struct(
+            skyMap=skyMap
+        )
+
+    def logSkyMapInfo(self, skyMap):
+        """!Log information about a sky map
+        @param[in] skyMap  sky map (an lsst.skyMap.SkyMap)
+        """
```
TallJimbo: Should be a numpydoc docstring, and note that the actual class is `lsst.skymap.BaseSkyMap`.

Also, given the duplication with the Gen2 `makeSkyMapTask`, maybe we should move this code into `BaseSkyMap`, as a method that takes a `Log` instance?

PR #289, opened 2019-05-04
Tickets/DM-19517: recalculate the PSF for DcrCoadds

isullivan:


parejkoj:
Note: the prefix on the PR name should just be `DM-NNNN:` per the dev guide: https://developer.lsst.io/work/flow.html#make-a-pull-request

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -252,6 +288,35 @@ def runDataRef(self, dataRef, selectDataList=None, warpRefList=None):
 
         return results
 
+    def processResults(self, coaddExposure, dataRef):
+        """Interpolate over missing data and mask bright stars.
+
+        Also detect sources on the coadd exposure and measure the final PSF,
+        if ``doCalculatePsf`` is set.
```
parejkoj: Do you mean `config.doCalculatePsf` here?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -252,6 +288,35 @@ def runDataRef(self, dataRef, selectDataList=None, warpRefList=None):
...
+
+        Parameters
+        ----------
+        coaddExposure : `lsst.afw.image.Exposure`
+            The final coadded exposure.
+        dataRef : `lsst.daf.persistence.ButlerDataRef`
+            Data reference defining the patch for coaddition and the
+            reference Warp
+        """
+        AssembleCoaddTask.processResults(self, coaddExposure, dataRef)
```
parejkoj: Shouldn't this be `super().processResults`?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -252,6 +288,35 @@ def runDataRef(self, dataRef, selectDataList=None, warpRefList=None):
...
+
+        if self.config.doCalculatePsf:
+            idFactory = self.makeIdFactory(dataRef)
+            expId = int(dataRef.get("dcrCoaddId"))
```
parejkoj: Why do you need this `int` cast?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -284,25 +349,27 @@ def prepareDcrInputs(self, templateCoadd, tempExpRefList, weightList):
...
+            airmassDict[visit] = airmass
+            angleDict[visit] = parallacticAngle
             if self.config.doAirmassWeight:
                 weightList[visitNum] *= airmass
             dcrShifts.append(np.max(np.abs(calculateDcr(visitInfo, templateCoadd.getWcs(),
                                                         filterInfo, self.config.dcrNumSubfilters))))
-        self.log.info("Selected airmasses:\n%s", airmassList)
-        self.log.info("Selected parallactic angles:\n%s", angleList)
+        self.log.info("Selected airmasses:\n%s", airmassDict)
+        self.log.info("Selected parallactic angles:\n%s", angleDict)
```
parejkoj: How many things end up in these dicts? I fear it might be a lot of output. Maybe min/max/mean instead in `log.info`, and the full dicts in `log.debug`?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -1004,3 +1073,33 @@ def loadSubExposures(self, bbox, statsCtrl, tempExpRefList, imageScalerList, spa
...
         return subExposures
+
+    def selectCoaddPsf(self, templateCoadd, tempExpRefList):
+        """Compute the PSF of the coadd from the exposures with the best seeing.
+
+        Parameters
+        ----------
+        templateCoadd : `lsst.afw.image.ExposureF`
+            The initial coadd exposure before accounting for DCR.
+        tempExpRefList : `list` of `lsst.daf.persistence.ButlerDataRef`
```
parejkoj: Can you just call these `warpRefList` instead?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -1004,3 +1073,33 @@ def loadSubExposures(self, bbox, statsCtrl, tempExpRefList, imageScalerList, spa
...
+        tempExpName = self.getTempExpDatasetName(self.warpType)
+        ccds = templateCoadd.getInfo().getCoaddInputs().ccds
+        psfRefSize = templateCoadd.getPsf().computeShape().getDeterminantRadius()*sigma2fwhm
+        psfSizeList = []
+        for visitNum, tempExpRef in enumerate(tempExpRefList):
+            psf = tempExpRef.get(tempExpName).getPsf()
+            psfSize = psf.computeShape().getDeterminantRadius()*sigma2fwhm
+            psfSizeList.append(psfSize)
+        sizeThreshold = min(np.median(psfSizeList), psfRefSize)
+        goodVisits = np.array(psfSizeList) <= sizeThreshold
```
parejkoj: Please add a comment about why this particular choice of size threshold.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: That's a relatively recent name change I hadn't seen. I'll change it.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: I don't think I do need it. I'll remove it.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: There will be one entry for each input visit. I think that's okay, since we already log the weight and number of positive/negative sources detected in each visit.

parejkoj: We talked about this on slack: what we really want would be a database with all of this information for the run. The logs are not a good place for it, but we don't have an obviously better one right now. :crying_cat_face:

PR #295, opened 2019-05-30
DM-18739: allow ingesting defects with a slim wrapper on ingestCalibs

SimonKrughoff:
I believe this also addresses DM-16817, so am asking for review of that with this same ticket.


python/lsst/pipe/tasks/ingest.py
```diff
@@ -307,20 +304,32 @@ def createTable(self, conn, table=None):
         @param conn    Database connection
         @param table   Name of table to create in database
         """
+        cursor = conn.cursor()
         if table is None:
             table = self.config.table
+        cmd = "SELECT name FROM sqlite_master WHERE type='table' AND name='%s'" % table
```
r-owen: I strongly suggest you leave `%s` in the string and pass the argument separately to `cursor.execute` -- both here and in the other simple sql commands below (it may be too much work for the more complicated command). This allows `execute` to sanitize the inputs for you, to prevent SQL injection. For example:
```
cmd = "SELECT name FROM sqlite_master WHERE type='table' AND name='%s'"
cursor.execute(cmd, table)
```
In cases where you don't think it's worth the effort please consider using f strings, as they are easier to read.

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -1,6 +1,7 @@
 import collections
 import datetime
 import sqlite3
+from dateutil import parser
```
r-owen: Consider `import dateutil`... `dateutil.parser.parse(...)`

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -38,6 +39,8 @@ def getCalibType(self, filename):
             obstype = "fringe"
         elif "sky" in obstype:
             obstype = "sky"
+        elif "defects" in obstype:
+            obstype = "defects"
```
r-owen: This is getting a bit long and the stutter worries me (a typo could easily creep in). Consider a loop, such as:
```
remap_dict = {"zero": "bias"}
for name in ("flat", "zero", "bias", "dark", "fringe", "sky", "defects"):
    if name in obstype:
        return remap_dict.get(name, name)
```

python/lsst/pipe/tasks/ingestDefects.py
```diff
@@ -0,0 +1,51 @@
...
+        self.add_argument("-n", "--dry-run", dest="dryrun", action="store_true",
+                          default=False, help="Don't perform any action?")
+        self.add_argument("--create", action="store_true", help="Create new registry?")
+        self.add_argument("--ignore-ingested", dest="ignoreIngested", action="store_true",
+                          help="Don't register files that have already been registered")
+        self.add_argument("root", help="Root directory to scan for defects.")
+
+
+class IngestDefectsTask(IngestCalibsTask):
+    """Task that generates registry for calibration images"""
```
r-owen: This doc string is wrong because this only handles defects. Consider copying the "run" doc string, since "Task that..." is not very helpful (it is clearly a task) and doc strings are supposed to be in active voice.

At one time task doc strings were supposed to be extensive, with sections for examples, debugging (which debug flags are supported)...is that still the case?

python/lsst/pipe/tasks/ingestDefects.py
```diff
@@ -0,0 +1,51 @@
...
+                    self.log.info('%i defects written for sensor: %s and calibDate: %s' %
+                                  (len(defects[d][s]), d, s.isoformat()))
+                    defects[d][s].writeFits(full_file_name)
+                    file_names.append(full_file_name)
+            args.files = file_names
+            args.mode = 'move'
+            args.validity = None  # Validity range is determined from the files
+            IngestCalibsTask.run(self, args)
+        except Exception:
+            raise(Exception)
```
r-owen: This except/raise does nothing useful. Please delete it.

tests/test_processCcd.py
```diff
@@ -132,20 +132,20 @@ def testProcessCcd(self):
...
+                        ("bgMean", bgMean, 191.48635852060525),
+                        ("bgStdDev", bgStdDev, 0.2399466881603354),
+                        ("numGoodPix", numGoodPix, 1966820),
+                        ("imMean", imMean, 1.1237668985230562),
+                        ("imStdDev", imStdDev, 85.81296241298496),
+                        ("varMean", varMean, 131.24003624152013),
+                        ("varStdDev", varStdDev, 55.98012493452948),
+                        ("psfIxx", psfIxx, 2.769679536557131),
+                        ("psfIyy", psfIyy, 2.2013649766299324),
+                        ("psfIxy", psfIxy, 0.14797939531970852)
```
r-owen: You are only testing to 7 places; please consider shortening these values so they only have a few extra digits. It would make the test a bit easier to read (the interesting digits are lost in the middle). I realizing rounding by hand is no fun, but you could easily print the data to the desired precision and copy/paste or use Python to generate this whole block.

python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -186,11 +192,6 @@ def __init__(self, *args, **kwargs):
                           help="Mode of delivering the files to their destination")
         self.add_argument("--create", action="store_true", help="Create new registry?")
         self.add_argument("--validity", type=int, required=True, help="Calibration validity period (days)")
-        self.add_argument("--calibType", type=str, default=None,
-                          choices=[None, "bias", "dark", "flat", "fringe", "sky", "defect"],
-                          help="Type of the calibration data to be ingested;"
-                               " if omitted, the type is determined from"
-                               " the file header information")
```
mrawls: If calibType can no longer be specified on the command line, I am pretty sure decam fringe ingestion will break, although it is already pretty broken (it must presently be run with mode=skip and the files must be moved/copied/linked manually to the calib repo).

PR #299, opened 2019-06-14
DM-19839: Use base sourceDetectionTask for measuring DcrCoadd PSF

isullivan:


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -314,8 +314,12 @@ def processResults(self, coaddExposure, dataRef):
                 exposure=coaddExposure,
                 exposureId=expId
             )
-            psfResults = self.measurePsf.run(coaddExposure, coaddSources, expId=expId)
-            coaddExposure.setPsf(psfResults.psf)
+            try:
+                psfResults = self.measurePsf.run(coaddExposure, coaddSources, expId=expId)
+            except RuntimeError as e:
+                self.log.warn("Unable to calculate PSF, using default coadd PSF: %s" % e)
```
mrawls: Just to double check, will things be able to proceed if `coaddExposure.setPsf` is never called? In other words, please confirm that is it *not* possible for `coaddExposure` to be missing a PSF entirely.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -168,7 +166,18 @@ def setDefaults(self):
...
+        # Only use bright sources for PSF measurement
+        self.detectSourcePsf.thresholdValue = 50
+        self.detectSourcePsf.nSigmaToGrow = 2
+        # A valid star for PSF measurement should at least fill 5x5 pixels
+        self.detectSourcePsf.minPixels = 25
+        # Use the variance plane to calculate signal to noise
+        self.detectSourcePsf.thresholdType = "pixel_stdev"
+        # The signal to noise limit is good enough, while the flux limit is set
+        # in dimensionless units and may not be appropriate for all data sets.
+        self.measurePsf.starSelector['objectSize'].doFluxLimit = False
```
mrawls: Where do the majority of these configs come from? 💯 for comments! I just want to check that it makes sense to set these defaults here and not somewhere where they would apply more widely.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -220,15 +229,14 @@ class DcrAssembleCoaddTask(CompareWarpAssembleCoaddTask):
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.config.doCalculatePsf:
-            schema = afwTable.SourceTable.makeMinimalSchema()
-            self.makeSubtask("detection", schema=schema)
-            self.makeSubtask('measurement', schema=schema)
-            self.makeSubtask("measurePsf", schema=schema)
+            self.schema = afwTable.SourceTable.makeMinimalSchema()
+            self.makeSubtask("detectSourcePsf", schema=self.schema)
+            self.makeSubtask('measurement', schema=self.schema)
```
mrawls: Can we get some consistent double vs. single quote action while you're at it, please? It's also not immediately clear to me why there is a `measurement` and a `measurePsf` subtasks as separate things. I really like how you've updated the name from `detection` --> `detectSourcePsf`. Would it be possible to do the same for `measurement` which is pretty vague?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: Yes, the PSF gets set earlier. So, if `self.config.doCalculatePsf=True` it over-writes the existing PSF.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: We definitely don't want these settings for the detection Task in general. I could imagine creating thin wrappers around the base detection Task with sensible names and default settings geared towards different use cases, and putting these settings in one geared towards detecting sources for PSF measurement. That sounds like an RFC, though...

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: How is `measurePsfSources`? I would probably also change the detection task to `detectPsfSources` for consistency.

mrawls: I think that's good! It helps clarify that you're detecting and measuring the PSF sources first, and then (presumably) measuring the PSF itself later.

PR #309, opened 2019-09-03
DM-21153: Fix bug in DCR coadds introduced in DM-20692

isullivan:
The merge of DM-20692 changed the call signature of `AssembleCoaddTask.processResults()`, which is used in `DcrAssembleCoaddTask`.  The breaking change was not caught by existing unit tests.
The fix is to simplify the code by moving the code to re-measure the coadd PSF to its own function.

### DIFF
--- a/python/lsst/pipe/tasks/dcrAssembleCoadd.py
+++ b/python/lsst/pipe/tasks/dcrAssembleCoadd.py
@@ -148,7 +148,7 @@ class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
         dtype=bool,
         doc="Set to detect stars and recalculate the PSF from the final coadd."
         "Otherwise the PSF is estimated from a selection of the best input exposures",
-        default=True,
+        default=False,
     )
     detectPsfSources = pexConfig.ConfigurableField(
         target=measAlg.SourceDetectionTask,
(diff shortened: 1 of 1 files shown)


RobertLuptonTheGood:
I don't understand, "The fix is to simplify the code by moving the code to re-measure the coadd PSF to its own function".  We don't measure PSFs on coadds, we calculate them based on the inputs.  Is this what you meant?

isullivan:
No, the coadd PSF that is calculated from the inputs is potentially contaminated by DCR, which we have hopefully corrected for while constructing the DCR model. So, there is the option to measure the PSF of the DCR model during the final steps of processing the results.

RobertLuptonTheGood:
How do you deal with discontinuities of the imaging coverage, which is precisely what the coadd PSF is designed to handle?  The PSF model needs to be chromatic (including DCR), so that should be handled automatically.

isullivan:
You are correct that this option does not handle discontinuities of the imaging coverage. However, even letting the PSF model be chromatic does not solve the problem of DCR in the input images: the PSF of a moderately-high-airmass observation will be elongated in one direction due to DCR, while if we are able to construct the DCR model correctly we will undo that elongation in the coadd. We could try to correct for that elongation at the start when calculating the coadd PSF, but the current code doesn't do anything like that.
So, right now you have to pick the less bad PSF for your data: either calculate the PSF from the inputs and ignore DCR, or measure the PSF from the final DCR model and ignore imaging discontinuities. In any case, the DCR code is currently broken on `master` regardless of which option you choose, and this PR fixes it.

RobertLuptonTheGood:
You're right that just having a chromatic PSF doesn't solve the problem, but I think that the information needed to correctly derive a coadd DCR-corrected PSF is present.  Please open a ticket to address this, and another to investigate whether CCD boundaries significantly affect your measured PSFs (for photometry, but if we want to use ZOGY the requirements are probably much stronger)

isullivan:
Yes, we are going to need a better DCR coadd PSF to use for ZOGY, and also to account for variable seeing in the DCR model. The ticket is DM-21202, and if you have particular suggestions or cautions before I start working on that, please add them as a comment on that ticket.
I'm not planning to write a ticket to investigate the measured PSF around CCD boundaries, however, because that option will become obsolete once the PSF is calculated correctly, and will be deleted.

RobertLuptonTheGood:
Great.  I agree that you don't need a new ticket providing the problem is clearly visible in the work being planned (and I think DM-21202 is probably good enough)

PR #312, opened 2019-09-19
DM-17023: update to new daf_butler APIs.

TallJimbo:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -273,6 +273,31 @@ class CalibrateConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Calibrate
...
+            if self.connections.astromRefCat != self.astromRefObjLoader.ref_dataset_name:
+                raise ValueError(
+                    f"Gen2 ({self.astromRefObjLoader.ref_dataset_name}) and Gen3 "
+                    f"({self.connections.astromRefCat}) astrometry reference catalogs are different.  "
+                    f"These options must be kept in sync until Gen2 is retired."
+                )
+        except AttributeError:
+            # Retargeting can mean that astromRefObjLoader doesn't have a
+            # ref_dataset_name, which is fine.
+            pass
```
andy-slac: This could quietly hide other errors than one described in comment, e.g. misspelled attribute name. It may be safer to do `getattr(self.astromRefObjLoader, "ref_dataset_name", None)` instead of catching an error. Same for `try ... except` below.

PR #313, opened 2019-09-19
DM-21361: Gen3 API conversion was not quite right for Gen2

yalsayyad:
Fixed a couple of other things on this ticket that weren't obvious with only subaru gen3 data to test with. In separate commits.

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -431,6 +431,7 @@ class AssembleCoaddTask(CoaddBaseTask, pipeBase.PipelineTask):
         retStruct = self.run(inputData['skyInfo'], inputs.tempExpRefList, inputs.imageScalerList,
                              inputs.weightList, supplementaryData=supplementaryData)
 
+        inputData.setdefault('brightObjectMask', None)
         self.processResults(retStruct.coaddExposure, inputData['brightObjectMask'], outputDataId)
 
         if self.config.doWrite:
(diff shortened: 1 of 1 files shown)


TallJimbo:
My only comments about pre-existing code (and stuff I think I should have caught on my last review), so feel free to ignore for this ticket:

- I took a quick look at `reorderAndPadList`, and I don't understand the docs for `inputKey` and `outputKey` - were those maybe for a previous version in which those weren't lists?  And the docs for the return value are a bit circular.  I can tell what the function is doing from reading the code, and it doesn't seem consistent with those docs.

- It's a shame the deferred handle object exposes `datasetRefOrType` as as public attribute - that's really an implementation detail (in that it's just saving all of the arguments to `butler.put` directly).  It looks like we should be doing some normalization of the those arguments up-front so it can just have a `ref` attribute directly.

yalsayyad:
* I'll update the docs for reorderAndPadList now since I'm not in a rush. 
* I thought so too. I'll open a ticket and make an inline comment, where it's used

yalsayyad:
Also, since you opened the door to complain about gen3 2.0,  I would prefer we call them `DeferredDataRefs` instead of `DeferredDatasetHandles`.   It quacks like the gen2 DataRefs.

TallJimbo:
> Also, since you opened the door to complain about gen3 2.0, I would prefer we call them DeferredDataRefs instead of DeferredDatasetHandles. It quacks like the gen2 DataRefs.

In isolation (or if I had to do it all over again), I'd agree.  Right now it's a bit trickier, because Gen3 has a `DatasetRef` concept that I'd also like to avoid confusion with.  But maybe it's not too late to rename `DatasetRef` -> `DatasetHandle` so we can rename `DeferredDatasetHandle` -> `[Deferred]DataRef`, especially because I think the current `DatasetRef` needs to be split into two or three similar-but-not-identical classes, and that will involve some renaming anyway.

PR #315, opened 2019-09-20
DM-21229: Select images based on PSF size in arcsec not pixels

isullivan:


isullivan:
My original commit failed on Jenkins, because I hadn't updated the unit tests. Feel free to take another look now that I have fixed the tests.

PR #325, opened 2019-11-06
DM-21421 Add pipeline definitions

natelust:


pipelines/Forced.yaml
```diff
@@ -0,0 +1,4 @@
+name: Forced
+tasks:
+  forcedPhotCcd: lsst.meas.base.forcedPhotCcd.ForcedPhotCcdTask
+  forcedPhotCoadd: lsst.meas.base.forcedPhotCoadd.ForcedPhotCoaddTask
```
andy-slac: Newline is missing in the last line, same for two other files below.

PR #326, opened 2019-11-14
DM-20039: add fakeSources from a generator

sr525:


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -373,6 +374,7 @@ def trim(row):
 
         return fakeCat[fakeCat.apply(trim, axis=1)]
 
+
```
natelust: I think this new blank line should not be here

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -531,8 +522,8 @@ def addFakeSources(self, image, fakeImages, sourceType):
         ----------
         image : `lsst.afw.image.exposure.exposure.ExposureF`
                     The image into which the fake sources should be added
-        fakeImages : `list`
-                    A list of tuples of `lsst.afw.image.ImageF` and `lsst.geom.Point2D,
+        fakeImages : `typing.Iterator`
```
natelust: After verifying with jonathan I think this should read
fakeImages : `typing.Iterator` [`tuple` ['lsst.afw.image.ImageF`, `lsst.geom.Point2d`]]

with the description line telling you what the image and point should be. i.e. "An iterator of tuples that contain images of fake sources, and the point at which they are to be inserted"

PR #327, opened 2019-11-14
DM-22222: Set default configs for ProcessCcd pipeline

arunkannawadi:


python/lsst/pipe/tasks/processCcd.py
```diff
@@ -63,11 +64,13 @@ class ProcessCcdConfig(pexConfig.Config):
     )
 
     def setDefaults(self):
-        self.isr.doWrite = False
-        self.charImage.doWriteExposure = False
-        self.charImage.detection.doTempLocalBackground = False
-        self.calibrate.detection.doTempLocalBackground = False
-        self.calibrate.deblend.maxFootprintSize = 2000
```
natelust: go to calibrate.CalibrateConfig and add detection.doTempLocalBackground = False and deblend.maxFootprintSize = 2000 into the setDefaults method.

python/lsst/pipe/tasks/processCcd.py
```diff
@@ -63,11 +64,13 @@ class ProcessCcdConfig(pexConfig.Config):
     )
 
     def setDefaults(self):
-        self.isr.doWrite = False
```
natelust: This does not need to be moved anywhere just set this back to the way it was on master

arunkannawadi:
@natelust In short, your review appears to be that I should find where the class exists and go set the configs in the corresponding class, instead of setting everything in one setDefaults method. Does it mean that I should read the yaml file in all those places?

natelust:
No, you should not be reading the yaml file at all. Reset it back to what it was wit git. Once you see the setDefault method in the classes I pointed you to you will see similar lines of code to what you were looking at. Just copy those lines into the setDefaults method of the corresponding class.

arunkannawadi:
Barring comments for the lines I moved, this should be good to go now.

PR #328, opened 2019-11-15
DM-20558 Rescale the coadd variances for difference imaging templates

kgabor:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -167,6 +175,7 @@ class ImageDifferenceConfig(pexConfig.Config):
         target=RegisterTask,
         doc="Task to enable image-to-image image registration (warping)",
     )
+
```
yalsayyad: Accidental whitespace added between configs

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -191,6 +200,7 @@ def setDefaults(self):
         self.subtract['al'].kernel.active.spatialKernelOrder = 1
         self.subtract['al'].kernel.active.spatialBgOrder = 2
         self.doPreConvolve = False
+        self.doScaleTemplateVariance = False
```
yalsayyad: Remove this. It's redundant, with the original default you specified above. 

In fact, every default in this set defaults method that is NOT a subtask config should be moved above.

python/lsst/pipe/tasks/scaleVariance.py
```diff
@@ -26,6 +26,8 @@
 from lsst.pipe.base import Task
 from lsst.meas.algorithms import SubtractBackgroundTask
 
+__all__ = ["ScaleVarianceConfig", "ScaleVarianceTask"]
+
```
yalsayyad: This is part of the "Add docs" commit, right.

kgabor: Automodapi may work without this, but I think it's better to have it explicitly written for it.

python/lsst/pipe/tasks/imageDifference.py
kgabor: These actually override the default values, I can't see why.

PR #329, opened 2019-11-19
Enable multiple ingest calls (e.g. by OODS).

ktlim:


srp3rd:
I looked at this, tested it using the new API, and it works great

PR #330, opened 2019-12-17
DM-22505 Create run method in ImageDifferenceTask and move all functionality from runDataRef except butler usage

kgabor:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -367,23 +368,175 @@ def runDataRef(self, sensorRef, templateIdList=None):
...
+        else:
+            self.log.info("Source selection via src product")
+            # Sources already exist; for data release processing
+            selectSources = sensorRef.get("src")
+
+        # Prepare kernelSources for AL either based on the template or on
+        # external reference catalog
+        if self.config.doSubtract:
+            if self.config.subtract.name == 'al':
+                if self.config.doSelectSources:
```
isullivan: This may be beyond the scope of this ticket, but I would really like to see the code in the `if self.config.doSelectSources:` block moved to its own method. This would include the code in the `if self.config.doSelectSources:` block at line 578 (I'll make a similar comment there).

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -423,55 +573,12 @@ def runDataRef(self, sensorRef, templateIdList=None):
                         preConvPsf = srcPsf
                     afwMath.convolve(destMI, srcMI, preConvPsf.getLocalKernel(), convControl)
                     exposure.setMaskedImage(destMI)
-                    scienceSigmaPost = scienceSigmaOrig*math.sqrt(2)
-                else:
-                    scienceSigmaPost = scienceSigmaOrig
 
                 # If requested, find sources in the image
                 if self.config.doSelectSources:
```
isullivan: Ideally this `if self.config.doSelectSources:` code block would get moved into its own method. This would include the code from line 408 above where I made a similar comment. This might be beyond the scope of this ticket, though.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -499,13 +606,19 @@ def runDataRef(self, sensorRef, templateIdList=None):
 
                     self.log.info("Selected %d / %d sources for Psf matching (%d for control sample)"
                                   % (len(kernelSources), len(selectSources), len(controlSources)))
+                else:
+                    # AL subtraction will do its own source detection
+                    kernelSources = None
+
                 allresids = {}
                 if self.config.doUseRegister:
```
isullivan: This may also be beyond the scope of this ticket, but I would like to see everything in this `if self.config.doUseRegister:` code block moved to its own method.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -367,23 +368,175 @@ def runDataRef(self, sensorRef, templateIdList=None):
...
+                idFactory=idFactory,
+            )
+        else:
+            self.log.info("Source selection via src product")
+            # Sources already exist; for data release processing
+            selectSources = sensorRef.get("src")
+
+        # Prepare kernelSources for AL either based on the template or on
+        # external reference catalog
+        if self.config.doSubtract:
```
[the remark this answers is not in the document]
kgabor: Let's try that... moved back.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -367,23 +368,175 @@ def runDataRef(self, sensorRef, templateIdList=None):
...
+        if self.config.doWriteMatchedExp:
+            sensorRef.put(results.matchedExposure, self.config.coaddName + "Diff_matchedExp")
+        if self.config.doAddMetrics and self.config.doSelectSources:
+            sensorRef.put(results.selectSources, self.config.coaddName + "Diff_kernelSrc")
+        if self.config.doWriteSubtractedExp:
+            sensorRef.put(results.subtractedExposure, subtractedExposureName)
+        return results
+
+    def run(self, exposure=None, selectSources=None, templateExposure=None, templateSources=None,
+            matchingSources=None, idFactory=None, calexpBackgroundExposure=None, subtractedExposure=None):
```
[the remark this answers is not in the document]
kgabor: As loadAndMatch is now within run(), this is input is no longer necessary.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -708,12 +815,12 @@ def runDataRef(self, sensorRef, templateIdList=None):
 
             # Match with the calexp sources if possible
             if self.config.doMatchSources:
-                if sensorRef.datasetExists("src"):
+                if selectSources is not None:
```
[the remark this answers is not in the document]
kgabor: If the calexp does not have a catalog and doSelectSources==False, selectSources won't be assigned in the AL branch. Also, if the calexp does not have a catalog, and the zogy branch is run, selectSources may remain None. To preserve the functionality, I'd rather leave it as it is.

python/lsst/pipe/tasks/imageDifference.py
kgabor: Added  todo comments.

PR #332, opened 2020-01-06
DM-22796: Rearrange test so that temp data files are removed if tests pass

timj:
This saves 200MB in the installed footprint.  The test data are
not deleted if a test fails.

### DIFF
--- a/tests/nopytest_test_coadds.py
+++ b/tests/nopytest_test_coadds.py
@@ -80,71 +80,28 @@ from lsst.pipe.tasks.multiBand import (DetectCoaddSourcesTask, MergeDetectionsTa
 DATAREPO_ROOT = os.path.join(os.path.dirname(__file__), ".tests", "testCoadds-data")
 
 
+def assertWrapper(func):
+    """Decorator to intercept any test failures and reraise whilst recording
+    a failure.
+
+    This allows tearDownClass to behave differently depending on
+    whether any of the tests failed.  In this case we clean up the test
+    data if all the tests passed but leave it behind if any of the tests
+    failed for any reason."""
+    def wrapped(self):
+        try:
+            func(self)
+        except Exception:
+            # Set the CLASS property since the data are per-test class
+            # so any failure is important in any of the tests.
+            type(self).failed = True
+            raise
+
+    return wrapped
+
+
 def setup_module(module):
     lsst.utils.tests.init()
-
-    if os.path.exists(DATAREPO_ROOT):
-        print("Deleting existing repo: %r" % (DATAREPO_ROOT,))
-        shutil.rmtree(DATAREPO_ROOT)
-
-    if not haveMeasBase:
-        raise unittest.SkipTest("meas_base could not be imported")
-
-    # Create a task that creates simulated images and builds a coadd from them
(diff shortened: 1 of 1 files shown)


tests/nopytest_test_coadds.py
```diff
@@ -80,72 +80,24 @@
 DATAREPO_ROOT = os.path.join(os.path.dirname(__file__), ".tests", "testCoadds-data")
 
 
+def assertWrapper(func):
+    """Decorator to intercept AssertionError and reraise whilst recording
+    a failure."""
+    def wrapped(self):
+        try:
+            func(self)
+        except Exception:
```
jdswinbank: Should we be catching `Exception` here, or `AssertionError` (per the docstring, above)?

tests/nopytest_test_coadds.py
```diff
@@ -218,6 +170,85 @@ def runForcedPhotCcdTask(butler):
...
+        runForcedPhotCoaddTask(butler, mocksTask)
+        runForcedPhotCcdTask(butler)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Removes test data if tests ran okay"""
+        if os.path.exists(DATAREPO_ROOT):
+            if not cls.failed:
+                print(f"Deleting temporary data repository {DATAREPO_ROOT}")
+                # Errors can be ignored at this point
```
jdswinbank: I'm not sure the comment really adds much... presumably, if the errors couldn't be ignored, the code wouldn't say `ignore_errors=True`. :-) I'd suggest either removing the comment, or using it to explain _why_ it's appropriate to ignore errors.

tests/nopytest_test_coadds.py
```diff
@@ -218,6 +170,85 @@ def runForcedPhotCcdTask(butler):
...
+        mergeMeasTask = MergeMeasurementsTask(config=mergeMeasConfig, butler=butler)
+        mergeMeasTask.writeSchemas(butler)
+        runTaskOnPatchList(butler, mergeMeasTask, mocksTask)
+
+        runForcedPhotCoaddTask(butler, mocksTask)
+        runForcedPhotCcdTask(butler)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Removes test data if tests ran okay"""
```
jdswinbank: Perhaps “if all tests passed” would be clearer?

tests/nopytest_test_coadds.py
```diff
@@ -80,72 +80,24 @@
 DATAREPO_ROOT = os.path.join(os.path.dirname(__file__), ".tests", "testCoadds-data")
 
 
+def assertWrapper(func):
+    """Decorator to intercept AssertionError and reraise whilst recording
+    a failure."""
```
jdswinbank: I think it'd be worth writing just another sentence here describing why this is a useful thing to do, just to help the reader orient themselves. Something along the lines of “this enables us to track whether any tests failed, and if so to preserve the test data for future reference”.

tests/nopytest_test_coadds.py
timj: I did have AssertionError here in the lsst_ci variant but then I got to pondering that any failures whilst running a test would also be a good reason not to delete. I can fix the docstring.

PR #333, opened 2020-01-09
DM-22541 Add initial gen3 support to ImageDifferenceTask.

kgabor:
* Add connection and minimal definition of inputs and outputs.
 * Add skymap as a task dimension, we do not support multiple skymaps in one run.

### DIFF
--- a/python/lsst/pipe/tasks/imageDifference.py
+++ b/python/lsst/pipe/tasks/imageDifference.py
@@ -23,6 +23,7 @@ import math
 import random
 import numpy
 
+import lsst.utils
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 import lsst.daf.base as dafBase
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -348,6 +401,39 @@ def makeIdFactory(expId, expBits):
...
+        Returns
+        -------
+        None
+        """
+        if self.getTemplate.config.coaddName == 'dcr':
+            # TODO DM-22952
+            raise NotImplementedError("TODO DM-22952: Dcr coadd templates are not yet supported in gen3.")
+
+        inputs = butlerQC.get(inputRefs)
+        expId, expBits = butlerQC.quantum.dataId.pack("visit_detector",
```
isullivan: This is missing an important log message that is in the Gen 2 version: `self.log.info("Processing %s" % (sensorRef.dataId))`. I think you can get this here with `self.log.info("Processing %s" % (inputs["exposure"].dataId))`

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -550,7 +636,7 @@ def run(self, exposure=None, selectSources=None, templateExposure=None, template
                         selectSources = self.subtract.getSelectSources(
                             exposure,
                             sigma=scienceSigmaPost,
-                            doSmooth=not self.doPreConvolve,
+                            doSmooth=not self.config.doPreConvolve,
```
isullivan: Has this option just been broken for a long time?

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -48,7 +48,60 @@
...
 
 
-class ImageDifferenceConfig(pexConfig.Config):
+class ImageDifferenceTaskConnections(
+    pipeBase.PipelineTaskConnections,
+    dimensions=("instrument", "visit", "detector", "skymap"),
+    defaultTemplates={"coaddName": "deep",
+                      "warpTypeSuffix": "",
+                      "fakesType": ""}
+):
```
isullivan: The formatting would be more consistent with the rest of the package as:
```
class ImageDifferenceTaskConnections(pipeBase.PipelineTaskConnections,
                                     dimensions=("instrument", "visit", "detector", "skymap"),
                                     defaultTemplates={"coaddName": "deep",
                                                       "warpTypeSuffix": "",
                                                       "fakesType": ""}
                                     ):
```

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -348,6 +401,39 @@ def makeIdFactory(expId, expBits):
...
+        expId : `int`
+            Exposure id.
+
+        expBits: `int`
+            Number of used bits in ``expId``.
+
+        Returns
+        -------
+        None
+        """
```
isullivan: This docstring looks copy-pasted from above. I think you actually want to delete it, and add a `@utils.inheritDoc(pipeBase.PipelineTask)` decorator above since these are just the standard parameters for a pipelineTask.

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -348,6 +401,39 @@ def makeIdFactory(expId, expBits):
...
+        inputs = butlerQC.get(inputRefs)
+        expId, expBits = butlerQC.quantum.dataId.pack("visit_detector",
+                                                      returnMaxBits=True)
+        inputs['idFactory'] = self.makeIdFactory(expId=expId, expBits=expBits)
+        inputs['templateExposure'] = self.getTemplate.assembleTemplateExposure(
+            butlerQC, inputRefs.skyMap, inputRefs.coaddExposures, inputs['exposure']
+        )
+        del inputs['coaddExposures']
+        del inputs['skyMap']
+        outputs = self.run(**inputs)
```
isullivan: I would suggest defining new variables `idFactory` and `templateExposure` above, and passing the arguments to `self.run()` explicitly, instead manipulating `inputs` so that it unpacks correctly.

python/lsst/pipe/tasks/imageDifference.py
kgabor: This was spotted when I did a test run with doSelectSources == True.

python/lsst/pipe/tasks/imageDifference.py
kgabor: Sure. Evolved around the `self.run(**inputs)` tutorial example.

PR #334, opened 2020-01-10
DM-23026: Make sure ic source fields are being propagated

sr525:


python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -309,6 +326,111 @@ def run(self, fakeCat, exposure, wcs=None, photoCalib=None, exposureIdInfo=None)
...
+        icSourceSchemaMapper = afwTable.SchemaMapper(icSourceCat.schema, sourceCat.schema)
+
+        # Add the desired columns from the config option icSourceFieldsToCopy
+        missingFieldNames = []
+        for fieldName in self.config.icSourceFieldsToCopy:
+            try:
+                schemaItem = icSourceCat.schema.find(fieldName)
+            except KeyError:
+                missingFieldNames.append(fieldName)
+            else:
```
PaulPrice: This is prolly a matter of personal preference, but since everything you're doing is embedded in a `for` loop, I'd put a `continue` in the `except` block, get rid of the `else` and move its contents to the left --- I'd find that easier to read. Totally up to you.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -309,6 +326,111 @@ def run(self, fakeCat, exposure, wcs=None, photoCalib=None, exposureIdInfo=None)
...
+        sourceSchemaMapper = afwTable.SchemaMapper(sourceCat.schema)
+        sourceSchemaMapper.addMinimalSchema(sourceCat.schema, True)
+
+        icSourceSchemaMapper = afwTable.SchemaMapper(icSourceCat.schema, sourceCat.schema)
+
+        # Add the desired columns from the config option icSourceFieldsToCopy
+        missingFieldNames = []
+        for fieldName in self.config.icSourceFieldsToCopy:
+            try:
+                schemaItem = icSourceCat.schema.find(fieldName)
```
PaulPrice: Can't you just do an `if fieldName in icSourceCat.schema`?

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -309,6 +326,111 @@ def run(self, fakeCat, exposure, wcs=None, photoCalib=None, exposureIdInfo=None)
...
+        if missingFieldNames:
+            raise RuntimeError("isSourceCat is missing fields {} specified in "
+                               "icSourceFieldsToCopy".format(missingFieldNames))
+
+        self.calibSourceKey = icSourceSchemaMapper.addOutputField(afwTable.Field["Flag"]("calib_detected",
+                                                                  "Source was detected as an icSource"))
+
+        schema = icSourceSchemaMapper.getOutputSchema()
+        newCat = afwTable.SourceCatalog(schema)
+        newCat.reserve(len(sourceCat))
```
PaulPrice: It's prolly not necessary (I think `extend` does it for you), but glad to see you're thinking about memory allocation issues.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -309,6 +326,111 @@ def run(self, fakeCat, exposure, wcs=None, photoCalib=None, exposureIdInfo=None)
...
+
+        mc = afwTable.MatchControl()
+        mc.findOnlyClosest = False  # return all matched objects
+        matches = afwTable.matchXy(newCat, icSourceCat, self.config.matchRadiusPix, mc)
+        deblendKey = newCat.schema["deblend_nChild"].asKey()
+        # if deblended, keep children
+        matches = [m for m in matches if m[1].get(deblendKey) == 0]
+
+        # Because we had to allow multiple matches to handle parents, we now
+        # need to prune to the best matches
```
PaulPrice: Why not prune the catalog before matching?

    select = newCat["deblend_nChild"] == 0
    matches = afwTable.matchXy(newCat[select], icSourceCat, self.config.matchRadiusPix)

That's a more efficient operation (fewer sources in the catalog) and less code.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -309,6 +326,111 @@ def run(self, fakeCat, exposure, wcs=None, photoCalib=None, exposureIdInfo=None)
...
+            src.setFlag(self.calibSourceKey, True)
+            # src.assign copies the footprint from icSrc, which we don't want
+            # (DM-407)
+            # so set icSrc's footprint to src's footprint before src.assign,
+            # then restore it
+            icSrcFootprint = icSrc.getFootprint()
+            try:
+                icSrc.setFootprint(src.getFootprint())
+                src.assign(icSrc, icSourceSchemaMapper)
+            finally:
```
PaulPrice: Nice!

python/lsst/pipe/tasks/processCcdWithFakes.py
sr525: I changed this to an if else, which also affects the above comment.

python/lsst/pipe/tasks/processCcdWithFakes.py
sr525: Because I'm silly and didn't really think about what I had copied. Thank you.

PR #335, opened 2020-01-14
DM-22797: Various FITS header fixes

PaulPrice:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -835,17 +835,19 @@ def setMetadata(self, exposure, photoRes=None):
...
             magZero = photoRes.zp - 2.5*math.log10(exposureTime)
-            self.metadata.set('MAGZERO', magZero)
         except Exception:
             self.log.warn("Could not set normalized MAGZERO in header: no "
                           "exposure time")
+            magZero = math.nan
 
         try:
-            metadata = exposure.getMetadata()
+            metadata.set('MAGZERO', magZero)
```
parejkoj: If we're going to write something to the headers instead of having people use our objects, we should at least write something that looks like what our objects produce: a multiplicative factor that takes instFlux to nanojansky. This is a lot easier for users to work with and understand, I think.

PaulPrice: All I'm doing here is fixing a bug. You're welcome to RFC an alternative.

parejkoj: When I was doing the PhotoCalib work, I didn't even know about the existence of these values. I'd have probably removed this whole block if I did. Why are we writing output in `calibrate` instead of in the task that actually generates the values `PhotoCalTask`?

PaulPrice: `PhotoCalTask` returns the value it measures, and the caller, `CalibrateTask` writes it. I don't know why it was decided to do it like that, but it dates from caf337031aea02b21c572b628cc0bb509a17fa72, over 7 years ago.

PR #339, opened 2020-01-17
DM-21917: Convert DcrAssembleCoadd to a pipelineTask

isullivan:
These changes can only be merged after DM-23008 and DM-22648

This has been verified to work using HSC r-band test data. There should be no DCR in that dataset, but I was able to verify that the Gen 3 butler and pipelinetask functionality works.

### DIFF
--- a/python/lsst/pipe/tasks/dcrAssembleCoadd.py
+++ b/python/lsst/pipe/tasks/dcrAssembleCoadd.py
@@ -27,18 +27,80 @@ import lsst.geom as geom
 import lsst.afw.image as afwImage
 import lsst.afw.table as afwTable
 import lsst.coadd.utils as coaddUtils
+from lsst.daf.butler import DeferredDatasetHandle
 from lsst.ip.diffim.dcrModel import applyDcr, calculateDcr, DcrModel
 import lsst.meas.algorithms as measAlg
 from lsst.meas.base import SingleFrameMeasurementTask
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
-from .assembleCoadd import AssembleCoaddTask, CompareWarpAssembleCoaddTask, CompareWarpAssembleCoaddConfig
+import lsst.utils as utils
+from .assembleCoadd import (AssembleCoaddTask,
+                            CompareWarpAssembleCoaddConfig,
+                            CompareWarpAssembleCoaddTask)
+from .coaddBase import makeSkyInfo
 from .measurePsf import MeasurePsfTask
 
-__all__ = ["DcrAssembleCoaddTask", "DcrAssembleCoaddConfig"]
+__all__ = ["DcrAssembleCoaddConnections", "DcrAssembleCoaddTask", "DcrAssembleCoaddConfig"]
 
 
-class DcrAssembleCoaddConfig(CompareWarpAssembleCoaddConfig):
+class DcrAssembleCoaddConnections(pipeBase.PipelineTaskConnections,
+                                  dimensions=("tract", "patch", "abstract_filter", "skymap"),
+                                  defaultTemplates={"inputCoaddName": "deep",
+                                                    "outputCoaddName": "dcr",
+                                                    "warpType": "direct",
+                                                    "warpTypeSuffix": "",
+                                                    "fakesType": ""}):
+    inputWarps = pipeBase.connectionTypes.Input(
+        doc=("Input list of warps to be assembled i.e. stacked."
+             "WarpType (e.g. direct, psfMatched) is controlled by the warpType config parameter"),
+        name="{inputCoaddName}Coadd_{warpType}Warp",
+        storageClass="ExposureF",
+        dimensions=("tract", "patch", "skymap", "visit", "instrument"),
+        deferLoad=True,
+        multiple=True
+    )
+    skyMap = pipeBase.connectionTypes.Input(
+        doc="Input definition of geometry/bbox and projection/wcs for coadded exposures",
+        name="{inputCoaddName}Coadd_skyMap",
+        storageClass="SkyMap",
+        dimensions=("skymap", ),
+    )
+    brightObjectMask = pipeBase.connectionTypes.PrerequisiteInput(
+        doc=("Input Bright Object Mask mask produced with external catalogs to be applied to the mask plane"
+             " BRIGHT_OBJECT."),
+        name="brightObjectMask",
+        storageClass="ObjectMaskCatalog",
+        dimensions=("tract", "patch", "skymap", "abstract_filter"),
+    )
+    templateExposure = pipeBase.connectionTypes.Input(
+        doc="Input coadded exposure, produced by previous call to AssembleCoadd",
+        name="{fakesType}{inputCoaddName}Coadd{warpTypeSuffix}",
+        storageClass="ExposureF",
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -378,9 +511,18 @@ def prepareDcrInputs(self, templateCoadd, warpRefList, weightList):
         angleDict = {}
         psfSizeDict = {}
         for visitNum, warpExpRef in enumerate(warpRefList):
-            visitInfo = warpExpRef.get(tempExpName + "_visitInfo")
-            visit = warpExpRef.dataId["visit"]
-            psf = warpExpRef.get(tempExpName).getPsf()
+            if isinstance(warpExpRef, DeferredDatasetHandle):
+                # Gen 3 API
+                bbox = geom.Box2I(templateCoadd.getBBox().getMin(), geom.Extent2I(1, 1))
```
natelust: watch DM-22363, which I hope to merge today. It will allow you to use a deferredDatasetHandle to get a component of a dataset directly. i.e. warpExpRef.get(component="visitInfo").

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -395,7 +537,8 @@ def prepareDcrInputs(self, templateCoadd, warpRefList, weightList):
         self.log.info("Selected parallactic angles:\n%s", angleDict)
         self.log.info("Selected PSF sizes:\n%s", psfSizeDict)
         self.bufferSize = int(np.ceil(np.max(dcrShifts)) + 1)
-        psf = self.selectCoaddPsf(templateCoadd, warpRefList)
+#        psf = self.selectCoaddPsf(templateCoadd, warpRefList)
```
natelust: remove comment

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -664,7 +815,11 @@ def dcrAssembleSubregion(self, dcrModels, subExposures, bbox, dcrBBox, warpRefLi
         residualGeneratorList = []
 
         for warpExpRef in warpRefList:
-            exposure = subExposures[warpExpRef.dataId["visit"]]
+            if isinstance(warpExpRef, DeferredDatasetHandle):
+                visit = warpExpRef.datasetRefOrType.dataId["visit"]
```
natelust: datasetRefOrType is not guaranteed to have a dataId attribute, or it might be None. If you want an error message other than "str has no attribute dataId" or something like that, you should either try catch, or check dataId here. (in practice it will almost certainly do what you want, but the check will guard against some change somewhere else)

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -807,10 +963,14 @@ def calculateConvergence(self, dcrModels, subExposures, bbox, warpRefList, weigh
         metric = 0.
         metricList = {}
         for warpExpRef, expWeight in zip(warpRefList, weightList):
-            exposure = subExposures[warpExpRef.dataId["visit"]][bbox]
+            if isinstance(warpExpRef, DeferredDatasetHandle):
+                visit = warpExpRef.datasetRefOrType.dataId["visit"]
```
natelust: same as above

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -1093,7 +1255,12 @@ def loadSubExposures(self, bbox, statsCtrl, warpRefList, imageScalerList, spanSe
         zipIterables = zip(warpRefList, imageScalerList, spanSetMaskList)
         subExposures = {}
         for warpExpRef, imageScaler, altMaskSpans in zipIterables:
-            exposure = warpExpRef.get(tempExpName + "_sub", bbox=bbox)
+            if isinstance(warpExpRef, DeferredDatasetHandle):
+                visit = warpExpRef.datasetRefOrType.dataId["visit"]
```
natelust: same as above

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -1131,9 +1299,17 @@ def selectCoaddPsf(self, templateCoadd, warpRefList):
         psfSizes = np.zeros(len(ccds))
         ccdVisits = np.array(ccds["visit"])
         for warpExpRef in warpRefList:
-            psf = warpExpRef.get(tempExpName).getPsf()
+            if isinstance(warpExpRef, DeferredDatasetHandle):
+                # Gen 3 API
+                bbox = geom.Box2I(templateCoadd.getBBox().getMin(), geom.Extent2I(1, 1))
+                tempExp = warpExpRef.get(parameters={'bbox': bbox})
+                psf = tempExp.getPsf()
+                visit = warpExpRef.datasetRefOrType.dataId["visit"]
```
natelust: same as above

timj:
A caching datastore is on the todo list but we haven't implemented one yet.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -253,6 +310,63 @@ def __init__(self, *args, **kwargs):
...
+        if len(inputs.tempExpRefList) == 0:
+            self.log.warn("No coadd temporary exposures found")
+            return
+
+        supplementaryData = self.makeSupplementaryDataGen3(butlerQC, inputRefs, outputRefs)
+        retStruct = self.run(inputData['skyInfo'], inputs.tempExpRefList, inputs.imageScalerList,
+                             inputs.weightList, supplementaryData=supplementaryData)
+
+        inputData.setdefault('brightObjectMask', None)
+        for subfilter in range(self.config.dcrNumSubfilters):
```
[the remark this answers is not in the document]
isullivan: Setting the PSF can certainly be moved inside the `run` method, but changing the behavior of `processResults` would require additional changes to `assembleCoaddTask`. That's not hard, but it's beyond the scope of this ticket. Would you like me to file a ticket to change it in both places?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: This was an accidental change that made it in from an unrelated test. I'll revert this line.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: That's good to know, and we should fix this since that error message would be would be hard for a user to debug. We use dataId in the same way all throughout `assembleCoaddTask` which this inherits from, so I think the solution would be to define a simple helper function there and use it in all of the assemble* tasks. Do you mind if I do that in a separate (very simple) ticket?

python/lsst/pipe/tasks/dcrAssembleCoadd.py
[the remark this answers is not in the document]
natelust: no, that is ok. I think there is plans to refactor it at some point in the future post gen2 anyway

python/lsst/pipe/tasks/dcrAssembleCoadd.py
natelust: no problem

PR #340, opened 2020-01-17
DM-21308: Replace doApplyUberCal with doApplyExternal configs

erykoff:


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -501,26 +505,15 @@ def getCalibratedExposure(self, dataRef, bgSubtracted):
...
-                skyWcs = dataRef.get("jointcal_wcs")
-                exposure.setWcs(skyWcs)
+        if self.config.doApplyExternalPhotoCalib:
+            photoCalib = dataRef.get(f"{self.config.externalPhotoCalibName}_photoCalib")
         else:
             photoCalib = exposure.getPhotoCalib()
 
+        if self.config.doApplyExternalSkyWcs:
+            skyWcs = dataRef.get(f"{self.config.externalSkyWcsName}_wcs")
+            exposure.setWcs(skyWcs)
```
mrawls: I am belatedly seeing the comment in the **obs_subaru HSC** makeCoaddTempExp.py which states "This was copy-pasted from MakeCoaddTempExpTask, and extra step added to check whether an additional mask is needed." BOOO. The docstring further admits it is a hack.

This explains why I was confusing myself thinking I had already reviewed this code 😠 

Anyway, assuming this code duplication is an independent problem I cannot control, and will be fixed in DM-20074 per the comment in obs_subaru, please see my review comments in obs_subaru, where I suggest changes in literally identical code.

tests/test_makeCoaddTempExp.py
```diff
@@ -42,16 +42,16 @@ def setUp(self):
         meanCalibration = 1e-4
         calibrationErr = 1e-5
         self.exposurePhotoCalib = lsst.afw.image.PhotoCalib(meanCalibration, calibrationErr)
-        # a "jointcal_photoCalib" calibration to return
-        self.jointcalPhotoCalib = lsst.afw.image.PhotoCalib(1e-6, 1e-8)
+        # An ubercalibrated photoCalib calibration to return
+        self.uberPhotoCalib = lsst.afw.image.PhotoCalib(1e-6, 1e-8)
```
mrawls: Please call this external rather than uber throughout the tests

mrawls:
Please also consider squashing some of the commits here.

PR #342, opened 2020-01-17
DM-23056: Suppress FutureWarnings from LSST code

jdswinbank:


jdswinbank:
I don't believe there should be any problems with that — these changes are all just reorganizing the API internally, and don't do anything to the on disk format.

PR #346, opened 2020-02-05
DM-23342: Fix ingestDriver

PaulPrice:


python/lsst/pipe/tasks/ingest.py
```diff
@@ -544,37 +551,38 @@ def runFile(self, infile, registry, args, pos):
...
-            except Exception as exc:
-                raise IngestError(f"Failed to register file {infile}", infile, pos) from exc
+            return None
+        if registry is not None and hduInfoList is not None:
+            for info in hduInfoList:
+                try:
+                    self.register.addRow(registry, info, dryrun=args.dryrun, create=args.create)
+                except Exception as exc:
+                    raise IngestError(f"Failed to register file {infile}", infile, pos) from exc
+        return hduInfoList
```
ktlim: I would recommend only returning this if `registry is None` and otherwise returning `None` if everything has been registered.

PR #347, opened 2020-02-05
DM-23103: Update functor unittests to no longer rely on test_multilevel_parq.csv.gz 

morriscb:


tests/test_functors.py
```diff
@@ -42,119 +42,181 @@
...
 
             psfColor_GR = self._funcVal(Color(fluxName, 'HSC-G', 'HSC-R',
-                                              dataset=dataset))
+                                              dataset=dataset),
+                                        parq)
 
             self.assertTrue(np.allclose((psfMag_G - psfMag_R).dropna(), psfColor_GR, rtol=0, atol=1e-13))
 
         # Check that behavior as expected when dataset not provided;
         #  that is, that the color comes from forced and default Mag is meas
```
timothydmorton: The following is not testing the same thing it used to be; when the test data was a partial real dataset, 'meas' and 'forced' were different, and then since the default 'dataset' when not explicitly provided were different for `Mag` and `Color`, the `np.allclose()` below returned `False` (as it used to be an `assertFalse`).  So either the behavior of `Mag` and `Color` have changed or the generated test dataset has identical 'meas' and 'forced' values.

tests/test_functors.py
```diff
@@ -42,119 +42,181 @@
...
+        self.columns.extend(["base_PsfFlux_instFlux", "base_PsfFlux_instFluxErr",
+                             "modelfit_CModel_instFlux", "modelfit_CModel_instFluxErr"])
+        self.dataDict["base_PsfFlux_instFlux"] = np.full(self.nRecords, 1000)
+        self.dataDict["base_PsfFlux_instFluxErr"] = np.full(self.nRecords, 10)
+        self.dataDict["modelfit_CModel_instFlux"] = np.full(self.nRecords, 1000)
+        self.dataDict["modelfit_CModel_instFluxErr"] = np.full(self.nRecords, 10)
+        parq = self.simulateMultiParquet(self.dataDict)
+
         for filt in self.filters:
             filt = 'HSC-G'
```
timothydmorton: I see this isn't your change, but while we're here, seems `filt = 'HSC-G'` should be removed?

tests/test_functors.py
```diff
@@ -174,12 +236,12 @@ def testComposite(self):
...
         func2 = CompositeFunctor(funcDict2, filt=filt)
-        df2 = self._compositeFuncVal(func2)
+        df2 = self._compositeFuncVal(func2, parq)
         self.assertTrue(df.equals(df2))
 
         func2.filt = 'HSC-R'
-        df3 = self._compositeFuncVal(func2)
-        self.assertFalse(df2.equals(df3))
+        df3 = self._compositeFuncVal(func2, parq)
+        self.assertTrue(df2.equals(df3))
```
timothydmorton: This also changed from `assertFalse` to `assertTrue`...why? I don't remember why it was so, but should make sure that this isn't evaluating to true because of identical simulated values.  This probably means that generally it would be better to make sure that simulated values in each column are different, this way equality/inequality tests will be more useful.

tests/test_functors.py
```diff
@@ -42,119 +42,181 @@
...
+        for ds in self.datasets:
+            for filterName in self.filters:
+                df = copy.copy(simpleDF)
+                df.reindex(sorted(df.columns), axis=1)
+                df['dataset'] = ds
+                df['filter'] = filterName
+                df.columns = pd.MultiIndex.from_tuples(
+                    [(ds, filterName, c) for c in df.columns],
+                    names=('dataset', 'filter', 'column'))
+                dfFilterDSCombos.append(df)
```
timothydmorton: Relating to several comments below, generating this fake data should make sure that different filters/datasets have different values in their columns, not the same, otherwise the equality/inequality tests later on lose their meaning.

tests/test_functors.py
morriscb: This seems like the test should be more of a general test of a generic functor loading/calculating on the correct dataset/filter. I've reverted it to fail again (thanks for catching it, didn't quite appreciate why it was supposed to fail). I can write the generic functor test in a future ticket if you would like me to make it. I have more tickets to write specific tests of values so I'll likely just leave things as simple values for now.

tests/test_functors.py
morriscb: There's a switch of the filter value just above and since the values are the same the test is now True. I've done a similar step of modifying the value for specifically R band this time force the test to fail.

tests/test_functors.py
morriscb: I would argue that such a test should be performed on a more generic functor, making sure that different datasets/filters with different values return said different values. Testing these on functors with a specific use seems to conflate two tests 1) that the internal calculations return the values expected (not being currently tested except in the case that color = mag1 - mag2), 2) that fuctors properly access different dataset/filter columns (tested currently within specific functor unittests rather than a specific test case for the functionality.) Like I said, this won't sit idle if we don't address it in this ticket. I'll be working on the functors and the SDM for the next few months so I can make a ticket to specifically write a test to test the dataset/filter functionality.

tests/test_functors.py
timothydmorton: sure, sounds good

tests/test_functors.py
timothydmorton: Yup, sounds good.  As you see, these tests were originally pretty bare-bones, so I do think it's a good idea to flesh them out to make sure they are computing the right things.

PR #349, opened 2020-02-11
DM-23352: Add jointcal, skyCorr to forcedPhotCcd

PaulPrice:


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -581,13 +544,6 @@ class MakeWarpConnections(pipeBase.PipelineTaskConnections,
         dimensions=("instrument", "visit", "detector"),
         multiple=True,
     )
-    skyCorrList = cT.Input(
```
yalsayyad: This can't be removed from the connections if there's the possibility that a subtask needs it.

PR #351, opened 2020-02-14
DM-23104: Enable per-band flags in transform object

yalsayyad:


python/lsst/pipe/tasks/functors.py
```diff
@@ -445,12 +449,10 @@ class FootprintNPix(Column):
 class CoordColumn(Column):
     """Base class for coordinate column, in degrees
     """
-    _allow_difference = False
     _radians = True
-    _defaultNoDup = True
+    _defaultNoDup = False
```
morriscb: What's the reason for this change in the default behavior?

yalsayyad: :+1: I tried to explain in the commit message, but it could use clarifying. Lots of words here:

We have a `CoordColumn` which is more specific than `Column` (it can convert to radians) and ` RA|DecColumn` which is a special case of  `CoordColumn`  that looks for columns coord_ra and coord_dec.  Before this ticket,  CoordColumn doesn't explode out per band.  Why not?   I didn't want to write separate non-expanding `CoordColumn` because we can have those already with `RAColumn` and `DecColumn`.  The alternative,  (exploding) `Column` <-- `ExplodingCoordColumn` <-- (non-exploding) `CoordColumn` <-- (non-exploding)`[RA|Dec]Column` for functors that are all very slight variations of each other seems like overkill vs. `(exploding) `Column` <-- (exploding) `CoordColumn` <-- (non-exploding)` [RA|Dec]Column` which is what this PR proposes. I think the only person who would notice is @timothydmorton

PR #352, opened 2020-02-25
DM-23638: add filename attribute to ParquetTable

timothydmorton:
The QA dashboard work needs to be able to easily access the filename of a ParquetTable object; this adds that ability.


timothydmorton:
Whoops, just noticed that this also 'blackened' the whole file due to my sublime settings; if I should revert back those formatting changes, let me know.

PR #355, opened 2020-03-06
DM-23030: Modify Photometry SDM Functor to use stored calibration value.

morriscb:


python/lsst/pipe/tasks/functors.py
```diff
@@ -1004,3 +1004,184 @@ def columns(self):
...
+
+    @property
+    def columns(self):
+        return [self.instFluxCol, self.photoCalibCol]
+
+    def _func(self, df):
+        return self.instFluxToMagnitude(df[self.col], self.fluxMag0)
+
+
+class LocalMagnitudeErr(Photometry):
```
kfindeisen: Should this be `LocalPhotometry`?

tests/test_functors.py
```diff
@@ -305,6 +313,65 @@ def testLocalPhotometry(self):
...
+                                    atol=1e-13,
+                                    rtol=0))
+        self.assertTrue(np.allclose(
+            magErr.values,
+            2.5 / np.log(10) * nanoJanskyErr.values / nanoJansky.values,
+            atol=1e-13,
+            rtol=0))
+
+        # Test functors
+        func = LocalNanojansky("base_PsfFlux_instFlux",
```
kfindeisen: Why not have one test case for each class instead of putting everything into a very long method? You'll get more information when you run the tests, it'll be easier to make changes to the test code, and you won't have to worry about weird side effects from shared variables like `func` and `val`.

python/lsst/pipe/tasks/functors.py
```diff
@@ -1004,3 +1004,184 @@ def columns(self):
...
+        ----------
+        instFlux : `numpy.ndarray` or `pd.Series`
+            Array of instrument flux measurements
+        localCalib : `numpy.ndarray` or `pd.Series`
+            Array of local photometric calibration estimates.
+
+        Returns
+        -------
+        calibFlux : `numpy.ndarray` or `pd.Series`
+            Array of calibrated flux measurements.
```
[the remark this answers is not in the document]
morriscb: Comes out as Series.

python/lsst/pipe/tasks/functors.py
morriscb: Don't know why this showed up in your diff? This was something I found and fixed during building the unittest. Github here even says it's outdated which is weird.

tests/test_functors.py
morriscb: Moved the tests into their own function so the duplicated code is reduced. Considering how closely related the functors are and how they dependent they are the previously tested methods, I'm fine with leaving them here for now.

morriscb:
Hey @kfindeisen, think I've gotten through all of your comments. Take a look and let me know if I missed anything. I'll squash the commits once you have a look.

python/lsst/pipe/tasks/functors.py
```diff
@@ -1124,6 +1123,14 @@ def instFluxErrToMagnitudeErr(self, instFlux, instFluxErr, localCalib, localCali
...
 class LocalNanojansky(LocalPhotometry):
     """Compute calibrated fluxes using the local calibration value.
+
+    See also
+    --------
+    LocalPhotometry
+    LocalNanojansky
+    LocalNanojanskyErr
+    LocalMagnitude
+    LocalMagnitudeErr
```
kfindeisen: I think this might be a bit overkill. You can at least remove `LocalPhotometry`, because the docs always have an automatic link to the base class.

PR #356, opened 2020-03-09
DM-23031: Create dipole separation functor

morriscb:
Create pixelScale unittest

Debug unittest

Compare to afw WCS pixel scale.

Adjust tolerence test.

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -897,6 +897,83 @@ class RadiusFromQuadrupole(Functor):
         return (df[self.colXX]*df[self.colYY] - df[self.colXY]**2)**0.25
 
 
+class ComputePixelScale(Functor):
+    """Compute the local pixel scale from the stored CDMatrix.
+    """
+    name = "Pixel Scale"
+
+    def __init__(self,
+                 colCD_1_1,
+                 colCD_1_2,
+                 colCD_2_1,
+                 colCD_2_2,
+                 **kwargs):
+        self.colCD_1_1 = colCD_1_1
+        self.colCD_1_2 = colCD_1_2
+        self.colCD_2_1 = colCD_2_1
+        self.colCD_2_2 = colCD_2_2
+        super().__init__(**kwargs)
+
+    @property
+    def columns(self):
+        return [self.colCD_1_1, self.colCD_1_2,
+                self.colCD_2_1, self.colCD_2_2]
+
+    def pixelScale(self, cd11, cd12, cd21, cd22):
+        """Compute the local pixel scale conversion.
+
+        Parameters
+        ----------
+        cd11 : `pandas.Series`
+            [1, 1] element of the local CDMatricies.
+        cd12 : `pandas.Series`
+            [1, 2] element of the local CDMatricies.
+        cd21 : `pandas.Series`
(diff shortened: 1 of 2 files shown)


TallJimbo:
Oh, I should add that I'm pretty certain `SkyWcs.getPixelScale` *does* get the sin(dec) part right, so if your test against that passes for widely different dec values, you can rest easy about that part.

morriscb:
> It's hard for me to evaluate whether this is correct without knowing what `base_LocalWcs_CDMatrix` is - if that is an affine approximation of the mapping from the pixel coordinates of the measurement image to a local gnomonic projection with angular-unit pixels, centered at the position of the object, then I think it's correct in the limit that the transformation between those two frames is just scale and rotation, not shear (i.e. as long as a distance in x is the same on the sky as a distance in y - I think this is a reasonable assumption anywhere on any of our coadds, but I'm not so sure about single-visit images near the edge of the focal plane).
> 
> To unpack that a bit, I said "local gnomonic projection with angular-unit pixels" instead of just "sky" because that's one a way to get the sin(dec) part right in those matrix elements, and I'm guessing from the "LocalWcs" nomenclature that this is what you actually did - apologies if I once knew and just forgot.

It's straight from the getCDMatrix method after getting a TANWcs at the objects coordinate. You can check out the code here: https://github.com/lsst/meas_base/blob/27e4659cd5023e8322ef7898a1568491f2b23f57/python/lsst/meas/base/plugins.py#L498

morriscb:
I've checked the SkyOrigin on the Wcs and the declination is +53. I then tried other declination values and tests still passed. @TallJimbo if you could have one last look at the math, keeping in mind that I'm using the `SkyWcs.getCdMatrix(Point2D)`. Looking at the SkyWcs code this is the same CdMatrix call that is used when calling `getTanWcs` so hopefully the math is working the way I think it is.

TallJimbo:
Huh. `SkyWcs.getCdMatrix` with a position argument is something I don't really understand the utility of - it apparently evaluates the Jacobian of the transformation from pixels to "intermediate world coordinates", which is a fairly arbitrary FITS WCS concept.  For a simple TAN WCS that would mean you'll get the right answer as long as you aren't too far away from the projection point (tract or visit center), but I don't know what "too far" means.  I'm thinking we should merge this ticket and open a new one to rewrite the LocalWcs logic to avoid calling `getCdMatrix` (and maybe even remove the variant of that that takes a position, as I think it's dangerous).

timj:
Naively, it seems to me that SkyWcs.getCdMatrix can't ever work reliably if you have large distortions in your WCS. The reason we are using complex WCS transformations is precisely because a CD matrix is not good enough.

morriscb:
@TallJimbo okay, I'll merge this. However, for the SDM we still need some sort of local (at the point the object) approximation of the WCS that is store-able in the catalog. `getTanWcs`, which relies on `getCdMatrix`, seemed to be the way to do this. Is there another option currently given the Wcs methods available?

TallJimbo:
I think something based on `SkyWcs.linearizePixelToSky` would be safe.  Probably not on its own; I think we'd at least need to multiply that matrix by either `sin(dec)` or `cos(dec)`, and maybe a bit more than that, but that method would include all of the nontrivial stuff (I just haven't worked through the details of what the trivial stuff would be).

PR #361, opened 2020-03-24
DM-23946: Change localWcs plugins to use a localGnomonicWcs transform.

TallJimbo:
Debug test fuctors

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -901,10 +901,10 @@ class RadiusFromQuadrupole(Functor):
         return (df[self.colXX]*df[self.colYY] - df[self.colXY]**2)**0.25
 
 
-class ComputePixelScale(Functor):
-    """Compute the local pixel scale from the stored CDMatrix.
+class LocalWcs(Functor):
+    """Computations using the stored localWcs.
     """
-    name = "Pixel Scale"
+    name = "LocalWcsOperations"
 
     def __init__(self,
                  colCD_1_1,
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/functors.py
```diff
@@ -918,50 +918,155 @@ def __init__(self,
...
+        cd21 : `pandas.Series`
+            [2, 1] element of the local Wcs affine transform.
+        cd22 : `pandas.Series`
+            [2, 2] element of the local Wcs affine transform.
+
+        Returns
+        -------
+        raDecTuple : tuple
+            RA and dec conversion of x and y given the local Wcs. Returned
+            units are in radians.
```
TallJimbo, line 945: I think the documentation for tuple returns is to document each tuple element separately, i.e.:
```
Returns
------
ra : `pandas.Series`
    RA coordinate from conversion of (x, y)
dec : ...
```

Also, if the arguments could be scalar floats or `np.ndarray` as well as `pandas.Series`, it's probably best to document that somehow.  Maybe @jonathansick has thoughts on how - there's no single interface type AFAIK, but it's got to be a common pattern.

python/lsst/pipe/tasks/functors.py
```diff
@@ -918,50 +918,155 @@ def __init__(self,
...
+
+        """
+        return (x * cd11 + y * cd12, x * cd21 + y * cd22)
+
+    def computeSkySeperation(self, ra1, dec1, ra2, dec2):
```
TallJimbo, line 950: typo: separation (also below)

PR #363, opened 2020-03-26
DM-23975: Change Gen3 subfilter definition from string to int

isullivan:
Also adds a missing collections argument.


TallJimbo:
@isullivan , I've just pushed a mostly-unrelated commit here because I noticed that we can now take advantage of a recent butler change to remove the do-nothing collection command-line argument from `makeGen3DcrSubfilters.py`.  Feel free to leave it here or cherry-pick to a different ticket if you prefer.

PR #364, opened 2020-03-28
DM-23044: PTC task should persist usable linearity models 

plazas:


python/lsst/pipe/tasks/read_curated_calibs.py
```diff
@@ -74,7 +76,7 @@ def check_metadata(obj, valid_start, instrument, chip_id, filepath, data_name):
     """
     md = obj.getMetadata()
     finst = md['INSTRUME']
-    fchip_id = md['DETECTOR']
+    fchip_id = md['DETECTOR_ID']
```
timj: I don't think you can change this because all the other calibration datasets use `DETECTOR`.  This routine only works if we have standardized metadata. (although I don't know how the defects tests passed with this change). @SimonKrughoff what am I missing?

python/lsst/pipe/tasks/read_curated_calibs.py
```diff
@@ -25,8 +26,9 @@ def read_one_chip(root, chip_name, chip_id):
         A dictionary of objects constructed from the appropriate factory class.
         The key is the validity start time as a `datetime` object.
     """
-    factory_map = {'qe_curve': Curve, 'defects': Defects}
-    files = glob.glob(os.path.join(root, chip_name, '*.ecsv'))
+    factory_map = {'qe_curves': Curve, 'defects': Defects, 'linearizer': Linearizer}
+    files = glob.glob(os.path.join(root, chip_name, '*.ecsv')) + glob.glob(os.path.join(root,
```
timj: In the interests of future proofing can we do this as:
```python
files = []
extensions = (".ecsv", ".yaml")
for ext in extensions:
    files.extend(glob.glob(os.path.join(root, chip_name, f"*.{ext}")))
```
?

python/lsst/pipe/tasks/read_curated_calibs.py
plazas: I'll go back to 'DETECTOR'. This is ambiguous and was stopping the code when I was passing the detector name instead of the id, which is what it actually expects. That's why I thought that changing the name would help clarify.

timj: Please remember to squash your new change with the previous commit so that the change to the header name disappears from the history.

python/lsst/pipe/tasks/read_curated_calibs.py
```diff
@@ -25,8 +26,12 @@ def read_one_chip(root, chip_name, chip_id):
         A dictionary of objects constructed from the appropriate factory class.
         The key is the validity start time as a `datetime` object.
     """
-    factory_map = {'qe_curve': Curve, 'defects': Defects}
-    files = glob.glob(os.path.join(root, chip_name, '*.ecsv'))
+    factory_map = {'qe_curves': Curve, 'defects': Defects, 'linearizer': Linearizer}
```
timj: You need to turn this back to qe_curve.  It's a rebase error (it used to be qe_curves but we changed it to qe_curve and on rebase you reverted that change).

python/lsst/pipe/tasks/read_curated_calibs.py
SimonKrughoff: I don't believe it is ambiguous.  The original posting on CLO lays out the three required metadata keys and `DETECTOR` is one of them.

plazas: I was referring to the case when you look only at the code: given that there is detector name and detector ID,  just "DETECTOR" seemed ambiguous (in the case when someone does not know what CLO stands for or has not read that post or does not know of its existence).

SimonKrughoff: Perhaps it deserves a comment in the code, but since I had pointed you at the CLO post in the past, I was surprised you found it ambiguous to the point that you'd change it unilaterally.

plazas: Please forgive me, but I can't find in my records the post that you are talking about. When I was trying to make the code work, I probably did not remember this post that you mention and decided to change the code in order to move forward in a way that I thought would make things clearer. I have changed the keyword back to just DETECTOR, so it's OK.  I was also surprised to find that it was not clear by the name of the keyword that it was expecting an ID and not a name: after a few rounds of crashing and debugging and looking at the source code, then I was able to tell.

SimonKrughoff: Here's an example: https://lsstc.slack.com/archives/C2JPMCF5X/p1575927103181300

You respond by saying
> Thank you for pointing this out again! Sorry that I forgot. I’ll ask if I run into problems.
So I think there are other examples.

plazas: Thank you for posting this out again!  I'm OK with changing the keyword back to "DETECTOR", which I already did,  although I think that perhaps it should be "DETECTORID" or "DETECTORINDEX", IMO.

SimonKrughoff: Noted.

PR #366, opened 2020-03-31
DM-22643: convert visualizeVisit to gen3

czwa:
Add gen3 implementation of visualizeVisit.

### DIFF
--- a/pipelines/VisualizeVisit.yaml
+++ b/pipelines/VisualizeVisit.yaml
@@ -1,14 +1,8 @@
 description: Visualize full visit/focal plane with binned and mosaic images.
 tasks:
-  bin:
-    class: lsst.pipe.tasks.visualizeVisit.VisBinTask
-    config:
-      binning: 8
-      connections.inputExp: 'calexp'
-      connections.outputExp: 'calexpBin'
-  moz:
-    class: lsst.pipe.tasks.visualizeVisit.VisMosaicTask
-    config:
-      binning: 8
-      connections.inputExps: 'calexpBin'
-      connections.outputData: 'calexpFP'
+  binning:
+    class: lsst.pipe.tasks.visualizeVisit.VisualizeBinExpTask
+  mosaic:
+    class: lsst.pipe.tasks.visualizeVisit.VisualizeMosaicExpTask
+contracts:
+  - binning.binning == mosaic.binning
--- a/pipelines/VisualizeVisit.yaml
+++ b/pipelines/VisualizeVisit.yaml
@@ -1,0 +1,14 @@
+description: Visualize full visit/focal plane with binned and mosaic images.
+tasks:
+  bin:
+    class: lsst.pipe.tasks.visualizeVisit.VisBinTask
+    config:
+      binning: 8
+      connections.inputExp: 'calexp'
+      connections.outputExp: 'calexpBin'
+  moz:
+    class: lsst.pipe.tasks.visualizeVisit.VisMosaicTask
+    config:
+      binning: 8
+      connections.inputExps: 'calexpBin'
+      connections.outputData: 'calexpFP'
--- a/python/lsst/pipe/tasks/visualizeVisit.py
+++ b/python/lsst/pipe/tasks/visualizeVisit.py
@@ -28,11 +28,12 @@ import lsst.afw.image as afwImage
 import lsst.afw.cameraGeom.utils as afwUtils
 
 
-__all__ = ['VisBinConfig', 'VisBinTask', 'VisMosaicConfig', 'VisMosaicTask']
+__all__ = ['VisualizeBinExpConfig', 'VisualizeBinExpTask',
+           'VisualizeMosaicExpConfig', 'VisualizeMosaicExpTask']
 
 
-class VisBinConnections(pipeBase.PipelineTaskConnections,
-                        dimensions=("instrument", "detector")):
+class VisualizeBinExpConnections(pipeBase.PipelineTaskConnections,
+                                 dimensions=("instrument", "detector")):
     inputExp = cT.Input(
         name="calexp",
         doc="Input exposure data to mosaic.",
@@ -54,14 +55,14 @@ class VisBinConnections(pipeBase.PipelineTaskConnections,
     )
 
 
-class VisBinConfig(pipeBase.PipelineTaskConfig,
-                   pipelineConnections=VisBinConnections):
+class VisualizeBinExpConfig(pipeBase.PipelineTaskConfig,
+                            pipelineConnections=VisualizeBinExpConnections):
     """Configuration for focal plane visualization.
     """
     binning = pexConfig.Field(
         dtype=int,
         default=8,
-        doc="Binning factor to apply.",
+        doc="Binning factor to apply to each input exposure's image data.",
     )
     detectorKeyword = pexConfig.Field(
         dtype=str,
@@ -70,12 +71,16 @@ class VisBinConfig(pipeBase.PipelineTaskConfig,
     )
 
 
-class VisBinTask(pipeBase.PipelineTask,
-                 pipeBase.CmdLineTask):
-    """Task for focal plane visualization.
+class VisualizeBinExpTask(pipeBase.PipelineTask,
+                          pipeBase.CmdLineTask):
+    """Bin the detectors of an exposure.
+
+    The outputs of this task should be passed to
+    VisualizeMosaicExpTask to be mosaicked into a full focal plane
+    visualization image.
     """
-    ConfigClass = VisBinConfig
-    _DefaultName = 'VisBin'
+    ConfigClass = VisualizeBinExpConfig
+    _DefaultName = 'VisualizeBinExp'
 
     def run(self, inputExp, camera):
         """Bin input image, attach associated detector.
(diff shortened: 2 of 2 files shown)


pipelines/VisualizeVisit.yaml
```diff
@@ -0,0 +1,14 @@
+description: Visualize full visit/focal plane with binned and mosaic images.
+tasks:
+  bin:
+    class: lsst.pipe.tasks.visualizeVisit.VisBinTask
+    config:
+      binning: 8
+      connections.inputExp: 'calexp'
+      connections.outputExp: 'calexpBin'
+  moz:
```
natelust, line 9: maybe use more descriptive labels for these tasks

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+        doc="Binning factor to apply.",
+    )
+    detectorKeyword = pexConfig.Field(
+        dtype=str,
+        default='DET-ID',
+        doc="Metadata keyword to use to find detector if not available from input.",
+    )
+
+
+class VisBinTask(pipeBase.PipelineTask,
```
natelust, line 73: It is more verbose, but I think our coding standards perfer longer descriptive names, so something like VisualizeBinTask. On that note, I am not sure what the "bin" is in this name, maybe be more descriptive there.

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+                 pipeBase.CmdLineTask):
+    """Task for focal plane visualization.
```
natelust, line 75: Please expand this doc string just a bit, maybe some what/and or how

pipelines/VisualizeVisit.yaml
```diff
@@ -0,0 +1,14 @@
+description: Visualize full visit/focal plane with binned and mosaic images.
+tasks:
+  bin:
+    class: lsst.pipe.tasks.visualizeVisit.VisBinTask
+    config:
+      binning: 8
+      connections.inputExp: 'calexp'
+      connections.outputExp: 'calexpBin'
```
natelust, line 8: Do you need to set these since it appears you are using the default names?

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+                Binned version of input image. (`lsst.afw.image.Exposure`)
+        """
+        if inputExp.getDetector() is None:
+            detectorId = inputExp.getMetadata().get(self.config.detectorKeyword)
+            if detectorId is not None:
+                inputExp.setDetector(camera[detectorId])
+
+        binned = inputExp.getMaskedImage()
+        binned = afwMath.binImage(binned, self.config.binning)
+        outputExp = afwImage.makeExposure(binned)
```
natelust, line 105: if this is going to be an exposure, you should copy all the associated info from an exposure, ie the exposure info and such.

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+    )
+    camera = cT.PrerequisiteInput(
+        name="camera",
+        doc="Input camera to use for mosaic geometry.",
+        storageClass="Camera",
+        dimensions=("instrument", "calibration_label"),
+    )
+
+    outputData = cT.Output(
+        name="calexpFP",
```
natelust, line 131: Maybe use a slightly more descriptive name

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+            CCD exposures, binned by `binning`.  The keys are the
+            detectorIDs, with the values the binned image exposure.
+
+        Returns
+        -------
+        image : `lsst.afw.image.Image`
+            Image mosaicked from the individual binned images for each
+            detector.
+        """
+        class ImageSource:
```
natelust, line 171: Dont put a class in a class just to hide the namespace, you have an all at the top of the file, so just put it at file scope. That will make both more readable.

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+            """Source of images for makeImageFromCamera"""
+            def __init__(self, exposures):
+                self.isTrimmed = True
```
natelust, line 174: Why even do this,its not settable and does not seem to be used anywhere

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+                self.exposures = exposures
+                self.background = np.nan
```
natelust, line 176: Document this with a comment, why are you doing this, why is it a class member if its only used in once method and is not alterable with init arguments. Are you expecting anyone to modify this from an instance ie foo.background? If not consider making it a local variable in getCcdImage, or if it is better as a class attribute make it private self._background

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+                else:
+                    image = self.exposures[detector.getId()]
+                if hasattr(image, "getMaskedImage"):
+                    image = image.getMaskedImage()
+                if hasattr(image, "getMask"):
+                    mask = image.getMask()
+                    isBad = mask.getArray() & mask.getPlaneBitMask("NO_DATA") > 0
+                    image = image.clone()
+                    image.getImage().getArray()[isBad] = self.background
+                if hasattr(image, "getImage"):
```
natelust, line 213: Do you actually want to be chaining all these if hasattrs together, or should some of them be elif s?

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+                    image = image.getImage()
+
+                image = afwMath.rotateImageBy90(image, detector.getOrientation().getNQuarter())
```
natelust, line 216: You should check if they actually need rotated or put a comment saying that rotateImageBy90 does the check, It would be a shame to do lots of linear algebra with identity matrices

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+
+        Returns
+        -------
+        output : `lsst.pipe.base.Struct`
+            Results struct with attribute:
+
+            ``outputExp``
+                Binned version of input image. (`lsst.afw.image.Exposure`)
+        """
+        expDict = dict((exp.getDetector().getId(), exp) for exp in inputExps)
```
natelust, line 245: This works, but is doing more work under the hood than is needed, just use dict comprehension:
```
expDict = {exp.getDetector().getId(): exp for exp in inputExps}
```

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+            Input exposure data to bin.
+        camera : `lsst.afw.cameraGeom.Camera`
+            Input camera to use for mosaic geometry.
+
+        Returns
+        -------
+        output : `lsst.pipe.base.Struct`
+            Results struct with attribute:
+
+            ``outputExp``
```
natelust, line 242: Something is weird with the indention here

python/lsst/pipe/tasks/visualizeVisit.py
```diff
@@ -0,0 +1,250 @@
...
+        storageClass="ImageF",
+        dimensions=("instrument", ),
+    )
+
+
+class VisMosaicConfig(pipeBase.PipelineTaskConfig,
+                      pipelineConnections=VisMosaicConnections):
+    """Configuration for focal plane visualization.
+    """
+    binning = pexConfig.Field(
```
natelust, line 142: So this is binning done on the focal plane exposure that is done from binned exposures? So if you bin exposures by 8, and then set this binning to 8 the final is 64 times smaller? This seems a bit unclear. If I am correct (or if I am not) please write more doc string to describe this, and maybe put it in comments. Someone just reading the configs might guess that this value must be set to match the other from the other task for some reason.

python/lsst/pipe/tasks/visualizeVisit.py
czwa: Makes sense to me.  Renamed to Visualize{Action}ExpTask, as they are tasks that Bin/Mosaic Exposure data for Visualization.

pipelines/VisualizeVisit.yaml
czwa: Probably not.  It's likely an artifact from me changing those defaults before making the PR.

python/lsst/pipe/tasks/visualizeVisit.py
czwa: I've actually cut out a lot of that checking, as the original code eventually strips everything but the Image plane, so it doesn't make sense to check things we throw out.

python/lsst/pipe/tasks/visualizeVisit.py
czwa: Unfortunately not.  
https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-struct

python/lsst/pipe/tasks/visualizeVisit.py
czwa: It is multiplicative.  I'll update the docstring to make that clear, and to mention why this is a good thing.

czwa: I've updated the docstring again to fix my incorrect understanding, after some additional testing from Yusra and Lee.  The two numbers must agree between tasks, and so I've added a contract enforcing that to the default pipeline.

PR #368, opened 2020-04-01
DM-24290: Remove if Butler exists checks from bin.src scripts.

DinoBektesevic:
Checks are made against local filesystem only. This stopped being a valid assumption since S3Datastore was implemented as the data repository does not need to be local anymore.

The failure case should be handled internally by Butler itself at instantiation time.

These check currently block executing `makeGen3SkyMap` if the location of the data repository is in S3 (f.e. `s3://bucketname/`). 

The errors raised by the butler are either `RuntimeError` if the butler config file is of unknown format or `FileNotFoundError` if it's not at the given location. I think this is fine, but if required I think it should be possible to reinstate the exact error messages as before by attempting to instantiate a Butler, catching the exceptions and then printing and exiting with error code 1.

### DIFF
--- a/bin.src/makeGen3DcrSubfilters.py
+++ b/bin.src/makeGen3DcrSubfilters.py
@@ -35,11 +35,6 @@ parser.add_argument("-C", "--config-file", dest="configFile",
 
 args = parser.parse_args()
 
-# Verify any supplied paths actually exist on disk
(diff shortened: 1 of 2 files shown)


DinoBektesevic:
@TallJimbo maybe you could share your thoughts on this PR when you find the time?

PR #371, opened 2020-04-07
Use uppercase hierarch keys for cfitsio 3.47

brianv0:
This is a fix for cfitsio 3.47. 

The array are stored by cfitsio 3.47 as upper case (via `HIERARCH`, so to retrieve the array  the key needs to be upper case.


timj:
Closing this because the code has moved to ci_hsc_gen3.

PR #374, opened 2020-04-14
DM-15536: Switch postISRCCD to use exposure, not visit.

TallJimbo:
In Gen3 (at least), visits are on-sky, and possibly combinations of snaps; ISR runs on exposures.

### DIFF
--- a/python/lsst/pipe/tasks/characterizeImage.py
+++ b/python/lsst/pipe/tasks/characterizeImage.py
@@ -74,6 +74,18 @@ class CharacterizeImageConnections(pipeBase.PipelineTaskConnections,
         storageClass="SourceCatalog",
     )
 
+    def adjustQuantum(self, datasetRefMap: pipeBase.InputQuantizedConnection):
+        # Docstring inherited from PipelineTaskConnections
+        try:
+            return super().adjustQuantum(datasetRefMap)
+        except pipeBase.ScalarError as err:
+            raise pipeBase.ScalarError(
+                f"CharacterizeImageTask can at present only be run on visits that are associated with "
+                f"exactly one exposure.  Either this is not a valid exposure for this pipeline, or the "
+                f"snap-combination step you probably want hasn't been configured to run between ISR and "
+                f"this task (as of this writing, that would be because it hasn't been implemented yet)."
+            ) from err
+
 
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -48,7 +48,7 @@ class CharacterizeImageConnections(pipeBase.PipelineTaskConnections,
         doc="Input exposure data",
         name="postISRCCD",
         storageClass="ExposureF",
-        dimensions=["instrument", "visit", "detector"],
+        dimensions=["instrument", "exposure", "detector"],
```
natelust, line 51: Because the dimensions here assume that exposure is 1:1 with visit, as that is eveything currently implemented, could you add an adjustQuantum function that makes this assertion with a comment explaining why. This will mean that this connection will need to be `multiple` and `runQuantum` adjusted but that will put things in place for the future.

timj: Are you saying that characterizeImage by definition requires that visits have been constructed from exposures for it to work? I thought `postISRCCD` is per-exposure and not per-visit though? This would mean you need to have the combineSnaps task inserted between the two phases.

TallJimbo: Yes, we'd need either a separate PipelineTask in between to combine snaps, or to begin CharacterizeImageTask by doing that work (in, say a subtask).  We don't have the algorithmic code for that (if we have anything, it's now quite bitrotted, and IIRC always made bad assumptions), so @natelust and I agreed in 1-1 discussions to just make sure the error message clarifies that the pipelines only support the case where there is one exposure per visit right now.

timj: Is it worth creating a snapCombine task that does nothing other than forward input to output and complain if there is more than one exposure to process? We surely need something to take the place of the exposure->visit transition else every post-ISR task is going to need hacks.

TallJimbo: The only other post-ISR tasks I can think of operate on calibration frames, and don't need visits at all (the rest have to follow `CharacterizeImageTask`).  So I don't think it makes sense to add that yet - it'd just fill up Pipelines with noise and data repositories with redundant intermediate datasets for now.

PR #383, opened 2020-04-30
Add additonal columns and returns needed for processCcdWithFakesDrive…

sr525:
…r and for

making source tables from the fakes.

### DIFF
--- a/python/lsst/pipe/tasks/processCcdWithFakes.py
+++ b/python/lsst/pipe/tasks/processCcdWithFakes.py
@@ -74,6 +74,20 @@ class ProcessCcdWithFakesConnections(PipelineTaskConnections, dimensions=("instr
         dimensions=("Tract", "SkyMap", "Instrument", "Visit", "Detector")
     )
 
+    icSourceCat = cT.Input(
+        doc="Catalog of calibration sources",
+        name="icSrc",
+        storageClass="sourceCatalog",
+        dimensions=("tract", "skymap", "instrument", "visit", "detector")
+    )
+
+    sfdSourceCat = cT.Input(
+        doc="Catalog of calibration sources",
+        name="src",
+        storageClass="sourceCatalog",
+        dimensions=("tract", "skymap", "instrument", "visit", "detector")
+    )
+
     outputExposure = cT.Output(
         doc="Exposure with fake sources added.",
         name="fakes_calexp",
@@ -118,6 +132,14 @@ class ProcessCcdWithFakesConfig(PipelineTaskConfig,
              "RuntimeError exception.")
     )
 
+    srcFieldsToCopy = pexConfig.ListField(
+        dtype=str,
+        default=("calib_photometry_reserved", "calib_photometry_used", "calib_astrometry_used"),
+        doc=("Fields to copy from the `src` catalog to the output catalog "
+             "for matching sources Any missing fields will trigger a "
+             "RuntimeError exception.")
+    )
+
     matchRadiusPix = pexConfig.Field(
         dtype=float,
         default=3,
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -237,12 +247,15 @@ def runDataRef(self, dataRef):
             photoCalib = calexp.getPhotoCalib()
 
         icSourceCat = dataRef.get("icSrc", immediate=True)
+        sfdSourceCat = dataRef.get("src", immediate=True)
 
         resultStruct = self.run(fakeCat, calexp, wcs=wcs, photoCalib=photoCalib,
-                                exposureIdInfo=exposureIdInfo, icSourceCat=icSourceCat)
+                                exposureIdInfo=exposureIdInfo, icSourceCat=icSourceCat,
+                                sfdSourceCat=sfdSourceCat)
```
czwa, line 254: sfdSourceCat is not listed in the Parameters section of the docstring.  Neither is icSourceCat, which isn't technically in scope here, but should be documented here as well (so N_parameters = N_documentedParameters).  If you could state what an "sfd" is as well, so it's clear what the content should be ("single frame driver"?).

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -237,12 +247,15 @@ def runDataRef(self, dataRef):
             photoCalib = calexp.getPhotoCalib()
 
         icSourceCat = dataRef.get("icSrc", immediate=True)
+        sfdSourceCat = dataRef.get("src", immediate=True)
```
czwa, line 250: ProcessCcdWithFakesConnections needs to be updated with cT.Input() entries for sfdSourceCat so this can work in Gen3.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -326,12 +340,13 @@ def run(self, fakeCat, exposure, wcs=None, photoCalib=None, exposureIdInfo=None,
...
         self.catalogCalculation.run(sourceCat)
-        sourceCat = self.copyCalibrationFields(icSourceCat, sourceCat)
+        sourceCat = self.copyCalibrationFields(icSourceCat, sourceCat, self.config.calibrationFieldsToCopy)
+        sourceCat = self.copyCalibrationFields(sfdSourceCat, sourceCat, self.config.srcFieldsToCopy)
 
         resultStruct = pipeBase.Struct(outputExposure=exposure, outputCat=sourceCat)
         return resultStruct
 
-    def copyCalibrationFields(self, calibCat, sourceCat):
+    def copyCalibrationFields(self, calibCat, sourceCat, fieldsToCopy):
```
czwa, line 349: Docstring for fieldsToCopy.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -370,10 +385,13 @@ def copyCalibrationFields(self, calibCat, sourceCat):
...
         if missingFieldNames:
             raise RuntimeError(f"calibCat is missing fields {missingFieldNames} specified in "
-                               "calibrationFieldsToCopy")
+                               "fieldsToCopy")
 
-        self.calibSourceKey = calibSchemaMapper.addOutputField(afwTable.Field["Flag"]("calib_detected",
-                                                               "Source was detected as an icSource"))
+        if "calib_detected" not in calibSchemaMapper.getOutputSchema():
+            self.calibSourceKey = calibSchemaMapper.addOutputField(afwTable.Field["Flag"]("calib_detected",
+                                                                   "Source was detected as an icSource"))
```
czwa, line 392: Is this the correct column doc still?

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -370,10 +385,13 @@ def copyCalibrationFields(self, calibCat, sourceCat):
...
+        else:
+            self.calibSourceKey = None
```
czwa, line 394: Is this None because if 'calib_detected' is already in the calibSchemaMapper.getOutputSchema(), then it will be set as part of the match process?

python/lsst/pipe/tasks/processCcdWithFakes.py
sr525: Thank you for catching both of these.

python/lsst/pipe/tasks/processCcdWithFakes.py
sr525: This should only be added from the icSource cat as it will already exist the second time through.

PR #384, opened 2020-05-05
DM-23078: Add sky objects to the single frame processing step

leeskelvin:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -665,6 +676,13 @@ def run(self, exposure, exposureIdInfo=None, background=None,
         if detRes.fpSets.background:
             for bg in detRes.fpSets.background:
                 background.append(bg)
+        if self.config.doSkySources:
+            skySourceFootprints = self.skySources.run(mask=exposure.mask, seed=exposureIdInfo.expId)
+            if skySourceFootprints:
+                for foot in skySourceFootprints:
+                    s = sourceCat.addNew()
+                    s.setFootprint(foot)
+                    s.set(self.skySources.skySourceKey, True)
```
TallJimbo, line 685: It's a little strange to have the logic to add the column to the schema in `SkyObjectTask` and the logic to fill the column here.  I can think of a couple of solutions:
 - Move the logic to add the column to the schema to `CalibrateTask.__init__` (leaving `SkyObjectTask` unchanged on this ticket).
 - Create a new `SkySourcesTask` that inherits from `SkyObjectTask`, with a `run` method that calls `super().run` and then runs the code here.

laurenam: Good point!  I’m leaning towards the former (i.e. move it to the `if self.config.doSkySources:` block at line 497 above).  What do you think @leeskelvin?

TallJimbo: Upon further reflection, I'm leaning much harder that way as well, and the reason why is quite topical: if you think back to the meeting a few weeks ago about design principles,
 - making a `SkySourcesTask` that inherits from `SkyObjectTask` but has a `run` method with a different signature (as it would have to to modify the catalog) would violate Liskov substitution;
 - putting this functionality into `SkyObjectTask` itself would violate the interface segregation principle - the existing code that uses just `SkyObjectTask` would then depend on code it doesn't need.

There are ways to do this with a new `SkySourcesTask` that don't have these problems (e.g. it could call `SkyObjectsTask` as a subtask), but I don't think it's worth it for ~6 lines of code.

laurenam: Thanks so much for the explanation!  I often find it difficult to truly absorb coding design principles without concrete examples, so this is super helpful!!  Your last point is along the lines of what I tried to ask for advice about one Wednesday meeting long ago when we first started in on this ticket...so thanks for that too 👍

leeskelvin: Thanks Jim, this feels like something @laurenam and I can address later today!

leeskelvin: Thanks again Jim. Lauren and I have adopted the change you recommend, moving the logic to add the column to the schema into CalibrateTask.__init__. We've also removed our PR for meas_algorithms, leaving that repo unchanged. 

We've re-run our test dataset, and all seems to be as before. The new changes have also successfully completed a Jenkins run. Keen to hear your thoughts.

leeskelvin:
Done, thanks Jim. I've updated the commit message. I'll push this to master and delete the branch, cheers!

PR #391, opened 2020-06-10
DM-25115 Implement zogy subtraction with non-spatial PSF

kgabor:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -586,11 +586,13 @@ def run(self, exposure=None, selectSources=None, templateExposure=None, template
...
 
             if self.config.subtract.name == 'zogy':
-                subtractRes = self.subtract.subtractExposures(templateExposure, exposure,
-                                                              doWarping=True,
-                                                              spatiallyVarying=self.config.doSpatiallyVarying,
-                                                              doPreConvolve=self.config.doPreConvolve)
-                subtractedExposure = subtractRes.subtractedExposure
+                subtractRes = self.subtract.run(exposure, templateExposure, doWarping=True)
+                if self.config.doPreConvolve:
+                    subtractedExposure = subtractRes.scoreExp
```
isullivan, line 591: See my comments in `ip_diffim` about score vs scorr

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -586,11 +586,13 @@ def run(self, exposure=None, selectSources=None, templateExposure=None, template
                 self.metadata.add("scaleTemplateVarianceFactor", templateVarFactor)
 
             if self.config.subtract.name == 'zogy':
-                subtractRes = self.subtract.subtractExposures(templateExposure, exposure,
-                                                              doWarping=True,
-                                                              spatiallyVarying=self.config.doSpatiallyVarying,
-                                                              doPreConvolve=self.config.doPreConvolve)
-                subtractedExposure = subtractRes.subtractedExposure
+                subtractRes = self.subtract.run(exposure, templateExposure, doWarping=True)
```
isullivan, line 589: I would also supply the keyword `calculateS=True` in the call to run, though see my comment in `ip_diffim` about renaming that keyword.

PR #397, opened 2020-06-30
DM-24327: Include coadd cutouts in alert packets

morriscb:
Write out warpedExposure.

Debug name.


timj:
Minor comment for the future, but we rebase branches before merging, we do not update them by merging from master to the ticket branch before merging back to master.  GitHub goes out of its way to convince you to update the branch with a big button but that button should not be used.

morriscb:
Sorry about that. Think I saw something about this before. Juggling merging 6 branches just made it too tempting. Won't use in the future. Anything I need to fix?

timj:
Once merged we can't do anything about it. It was a heads up for the future.

PR #398, opened 2020-07-14
Add ability to add fake sources selectively to either the template or…

sr525:
… the calexp.

### DIFF
--- a/python/lsst/pipe/tasks/insertFakes.py
+++ b/python/lsst/pipe/tasks/insertFakes.py
@@ -190,6 +190,20 @@ class InsertFakesConfig(PipelineTaskConfig,
         default="deep",
     )
 
+    doSubSelectSources = pexConfig.Field(
+        doc="Set to True if you wish to sub select sources to be input based on the value in the column"
+            "set in the sourceSelectionColName config option.",
+        dtype=bool,
+        default=False
+    )
+
+    sourceSelectionColName = pexConfig.Field(
+        doc="The name of the column in the input fakes catalogue to be used to determine which sources to"
+            "add, default is none and when this is used all sources are added.",
+        dtype=str,
+        default="templateSource"
+    )
+
 
 class InsertFakesTask(PipelineTask, CmdLineTask):
     """Insert fake objects into images.
@@ -209,7 +223,9 @@ class InsertFakesTask(PipelineTask, CmdLineTask):
         input file.
     `cleanCat`
         Remove rows of the input fake catalog which have half light radius, of either the bulge or the disk,
-        that are 0.
+        that are 0. Also removes rows that have Sersic index outside of galsim's allowed paramters. If
+        the config option sourceSelectionColName is set then this function limits the catalog of input fakes
+        to only those which are True in this column.
     `addFakeSources`
         Add the fake sources to the image.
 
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -190,6 +190,14 @@ class InsertFakesConfig(PipelineTaskConfig,
         default="deep",
     )
 
+    template = pexConfig.Field(
+        doc="If the image is to be used as a template then set to True. This will add only sources with"
```
morriscb, line 194: The doc description is a bit confusing and could use some simplifying. Especially the third sentence. As for a suggestion for the names "template" seems fine, but perhaps "image" could be changed to "visistExposure" or "ccdVisitExposure"?

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -540,6 +548,13 @@ def cleanCat(self, fakeCat):
         -------
         fakeCat : `pandas.core.frame.DataFrame`
                     The input catalog of fake sources but with the bad objects removed
+
+        Notes
+        -----
+        If the cofig option 'template' is set to True then only objects with the 'template' column
+        in the input fakes catalog will be added to the given image. If the 'image' column is set
+        to True then objects will only be added if the config option 'template' is set to False. If
```
morriscb, line 556: Again, the notes here is a bit confusing and could use some edits.

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -555,6 +570,22 @@ def cleanCat(self, fakeCat):
         self.log.info("Removing %d rows with nBulge or nDisk outside of %0.2f <= n <= %0.2f" %
                       (badRows, minN, maxN))
 
+        if self.config.template:
+            try:
+                goodRows = (fakeCat["template"])
```
morriscb, line 575: `goodRows` and `badRows` is a bit unclear what you're actually doing here. I assume this is left over from another test but the rows for these settings aren't bad or good. They are just used or aren't.

python/lsst/pipe/tasks/insertFakes.py
morriscb: The columns themselves might need a bit more descriptive name like `isTemplateSource` or `insertIntoTemplate`. Might help with naming to think of them like this.

python/lsst/pipe/tasks/mkTransientFakesCatalog.py
```diff
@@ -0,0 +1,345 @@
...
+
+    insertFakes = pexConfig.ConfigurableField(target=InsertFakesTask,
+                                              doc="Configuration for the fake sources")
+
+    def setDefaults(self):
+        super().setDefaults()
+
+
+class MkTransientFakesCatalogTask(PipelineTask, CmdLineTask):
+    """Insert fake objects into calexps.
```
morriscb, line 150: Is this doc copied from a previous task? Doesn't line up with the task name.

python/lsst/pipe/tasks/mkTransientFakesCatalog.py
```diff
@@ -0,0 +1,345 @@
...
+                                            doc="The calibration task to use.")
+
+    insertFakes = pexConfig.ConfigurableField(target=InsertFakesTask,
+                                              doc="Configuration for the fake sources")
+
+    def setDefaults(self):
+        super().setDefaults()
+
+
+class MkTransientFakesCatalogTask(PipelineTask, CmdLineTask):
```
morriscb, line 149: Is this task name communicating what the task is doing? Looks to insert into an image and run calibrate to detect sources. The name currently reads like it's creating a catalog of fakes.

python/lsst/pipe/tasks/mkTransientFakesCatalog.py
sr525: mkTransientFakesCatalogTask was supposed to be deleted, I'm not sure why it wasn't deleted remotely.

python/lsst/pipe/tasks/insertFakes.py
sr525: In case anyone wants to put sources with different fluxes into different coadds, I changed the column names to 'isTemplateSource' and 'isNotOnlyTemplateSource'.

sr525: I have tried to make this doc clearer and will add more details to the community post instructions when I update them. (Which I will do once we have finalised the discussion on this.)

python/lsst/pipe/tasks/insertFakes.py
sr525: I've changed the names throughout this function as the point you make here is valid not just for this bit and I would like to keep the naming consistent.

PR #399, opened 2020-07-19
DM-26011: Issue better error message if the detector name is not known to the camera

timj:


python/lsst/pipe/tasks/read_curated_calibs.py
```diff
@@ -127,6 +127,13 @@ def read_all(root, camera):
     calib_types = set()
     for d in dirs:
         chip_name = os.path.basename(d)
+        # Give informative error message if the detector name is not known
+        # rather than a simple KeyError
+        if chip_name not in name_map:
+            detectors = [det for det in camera.getNameIter()]
```
ktlim, line 133: Are you sure you want to do this for LSSTCam?

timj: I did try it out and it was only 3 lines on my terminal. I think I'll list the first 5 detectors as examples or something (I did it first because the name of the camera always say lsstCam in the obs_lsst camera YAML -- something which I later fixed).

PR #401, opened 2020-07-30
DM-26171: calibrations cannot be retrieved on the last day of a validity range

czwa:
Remove single-day gaps in validation ranges.

### DIFF
--- a/python/lsst/pipe/tasks/ingestCalibs.py
+++ b/python/lsst/pipe/tasks/ingestCalibs.py
@@ -161,7 +161,7 @@ class CalibsRegisterTask(RegisterTask):
             # A calib is valid until it is superseded
             for thisDate, nextDate in zip(dates[:-1], dates[1:]):
                 valids[thisDate][0] = thisDate
-                valids[thisDate][1] = nextDate - datetime.timedelta(1)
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -175,7 +175,7 @@ def fixSubsetValidity(self, conn, table, detectorData, validity):
                 if valids[date][1] > midpoint:
                     nextDate = dates[i + 1]
                     valids[nextDate][0] = midpoint + datetime.timedelta(1)
-                    valids[date][1] = midpoint
+                    valids[date][1] = midpoint + datetime.timedelta(1)
```
RobertLuptonTheGood, line 178: Did you find a way to confirm that this part of the fix is correct experimentally?

czwa: My initial change removed the `+ datetime.timedelta(1)` on line 177.  However, this resulted in both `validEnd_i` and `validStart_{i+1}` being one day early.  That was clearly wrong, as the `calibDate` was not contained within the range, leading to this version that increments both fields.

PR #403, opened 2020-07-31
Add ability to add images directly

sr525:


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -342,17 +350,109 @@ def run(self, fakeCat, image, wcs, photoCalib):
...
+        starImages = []
+
+        for (imFile, sourceType, mag, x, y) in zip(fakeCat[band + "imFilename"].array,
+                                                   fakeCat["sourceType"].array,
+                                                   fakeCat[self.config.magVar % band].array,
+                                                   fakeCat["x"].array, fakeCat["y"].array):
+
+            im = afwImage.ImageF.readFits(imFile)
+
+            xy = geom.Point2D(x, y)
```
taranu, line 419: This code block including the try/except appears in slightly modified form three times now, so you may want to consider putting it into a separate function with a note explaining its purpose (since it's only covered in the mkFakeStars docstring).

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -342,17 +350,109 @@ def run(self, fakeCat, image, wcs, photoCalib):
...
+        pixelScale : `float`
+                    The pixel scale of the image the sources are to be added to.
+
+        Returns
+        -------
+        galImages : `list`
+                    A list of tuples of `lsst.afw.image.exposure.exposure.ExposureF` and
+                    `lsst.geom.Point2D` of their locations.
+                    For sources labelled as galaxy.
+        starImages :
```
taranu, line 398: I think this should also end with : `list`.

python/lsst/pipe/tasks/insertFakes.py
sr525: I think each is different enough to leave it, I'll improve the docs though.

PR #409, opened 2020-09-16
DM-26553: Create gen3 unittests and test CreateApFakes pipeline tasks on data.

morriscb:


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -681,8 +680,7 @@ def cleanCat(self, fakeCat, starCheckVal):
         will be added.
         """
 
-        rowsToKeep = (((fakeCat[self.config.bulgeHLR] != 0.0) & (fakeCat[self.config.diskHLR] != 0.0))
-                      | (fakeCat[self.config.sourceType] == starCheckVal))
+        rowsToKeep = ((fakeCat[self.config.bulgeHLR] != 0.0) & (fakeCat[self.config.diskHLR] != 0.0))
```
sr525, line 683: You've lost a line here from the check.

kfindeisen (commented):
Even after looking at `CreateRandomApFakesTask`, I'm a bit unsure about the use of `tract` as one of the dimensions for `ProcessCcdWithFakesTask`. It looks like the Gen 2 version of `ProcessCcdWithFakesTask` operated mostly on the detector level, with tracts only used for auxiliary data.

Is this a deliberate change between the Gen 2 and Gen 3 behavior? If not, it will cause problems with other tasks that assume `fakes_calexp` and `fakes_src` are detector-level.

I don't want to sign off on this until I understand what's going on.

python/lsst/pipe/tasks/insertFakes.py
morriscb: Huh, don't know when that happened. Put it back.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -52,50 +53,50 @@ class ProcessCcdWithFakesConnections(PipelineTaskConnections, dimensions=("instr
...
+        dimensions=("tract", "instrument", "visit", "detector")
     )
 
     outputCat = cT.Output(
         doc="Source catalog produced in calibrate task with fakes also measured.",
         name="fakes_src",
         storageClass="SourceCatalog",
-        dimensions=("instrument", "visit", "detector"),
+        dimensions=("tract", "instrument", "visit", "detector"),
     )
```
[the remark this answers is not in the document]
morriscb: Removed the joincal requirements based on the config.

I think the confusion for the tract here was I read the policy in obs_base and confused some of the coad and calexp fake produces. Removed the tract dimension from the fake src and calexp outputs. As for if there is something wrong with the Gen2 code. I have no idea. I've only run this as a SubTask in ap_pipe/gen2 land and am only really concerned about getting this all hooked up as Gen3 pipeline tasks for the future.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -98,6 +99,13 @@ class ProcessCcdWithFakesConnections(PipelineTaskConnections, dimensions=("instr
         dimensions=("instrument", "visit", "detector"),
     )
 
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+
+        if not config.useUpdatedCalibs:
+            self.prerequisiteInputs.remove("wcs")
+            self.prerequisiteInputs.remove("photoCalib")
```
kfindeisen, line 107: Should be `inputs`, not `prerequisiteInputs`.

python/lsst/pipe/tasks/processCcdWithFakes.py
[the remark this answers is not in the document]
kfindeisen: I _think_ the new dimensions are correct if you only want one of the tract inputs per image, but can you rerun your tests before merging, just to make sure?

kfindeisen (approved):

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -52,36 +53,36 @@ class ProcessCcdWithFakesConnections(PipelineTaskConnections, dimensions=("instr
     fakeCat = cT.Input(
         doc="Catalog of fake sources to draw inputs from.",
         name="{CoaddName}Coadd_fakeSourceCat",
-        storageClass="Parquet",
+        storageClass="DataFrame",
         dimensions=("tract", "skymap")
```
kfindeisen, line 57: @sr525, just to make sure, is it ok for the Gen 3 code to only take a tract-level `*Coadd_fakeSourceCat`, and not a detector-level `fakeSourceCat`? That seems to be the only substantial difference between the Gen 2 and Gen 3 versions of the task at the moment.

morriscb: Okay, so I get what you're asking now. The catalogs are not different in how they are stored really. Both will be at the tract level, they just have different fluxes for the visit data vs coadd data.

sr525: Probably not? The visit level catalogues are needed for the variable sources, where the current way of using them is to make a catalogue specific to each calexp.

morriscb: Wait so are you saying we are going to produce specific catalogs for each ccdVisit vs cutting out the sources from a tract level catalog?

morriscb: Okay, this conversation doesn't seem to be continuing so I'm going to go ahead and merge the ticket.

PR #410, opened 2020-09-17
DM-26586: Skip initial cosmic ray detection if it fails

isullivan:
In very crowded fields, the cosmic ray detection algorithm can fail if it doesn't have a good PSF. We do an initial CR detection at the start of each iteration of measuring the PSF in image characterization, which means it uses the default PSF for the first iteration. By simply skipping CR masking if it fails at this stage we can get an improved PSF model, which can be used to perform CR masking in subsequent iterations.


RobertLuptonTheGood:
Please control this behaviour with a configuration option set by the user, defaulting to False

ebellm (approved):

PR #411, opened 2020-09-22
DM-26726: Properly set template names in AssembleCoadd

timj:
Previously the AssembleCoaddConnections class relied on config instances be unfrozen and modified the config's state. This commit sets the connection templates according to top level config in such a way as to not modify the config, allowing it to run on frozen configs.

### DIFF
--- a/python/lsst/pipe/tasks/assembleCoadd.py
+++ b/python/lsst/pipe/tasks/assembleCoadd.py
@@ -94,14 +94,20 @@ class AssembleCoaddConnections(pipeBase.PipelineTaskConnections,
     )
 
     def __init__(self, *, config=None):
-        # Override the connection's name template with config to replicate Gen2 behavior
-        config.connections.warpType = config.warpType
-        config.connections.warpTypeSuffix = makeCoaddSuffix(config.warpType)
-
-        if config.hasFakes:
-            config.connections.fakesType = "_fakes"
-
         super().__init__(config=config)
+
+        # Override the connection's name template with config to replicate Gen2 behavior
+        # This duplicates some of the logic in the base class, due to wanting Gen2 and
+        # Gen3 configs to stay in sync. This should be removed when gen2 is deprecated
+        templateValues = {name: getattr(config.connections, name) for name in self.defaultTemplates}
+        templateValues['warpType'] = config.warpType
+        templateValues['warpTypeSuffix'] = makeCoaddSuffix(config.warpType)
+        if config.hasFakes:
+            templateValues['fakesType'] = "_fakes"
+        self._nameOverrides = {name: getattr(config.connections, name).format(**templateValues)
+                               for name in self.allConnections}
+        self._typeNameToVarName = {v: k for k, v in self._nameOverrides.items()}
+        # End code to remove after deprecation
 
         if not config.doMaskBrightObjects:
             self.prerequisiteInputs.remove("brightObjectMask")


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -94,14 +94,20 @@ class AssembleCoaddConnections(pipeBase.PipelineTaskConnections,
...
     def __init__(self, *, config=None):
-        # Override the connection's name template with config to replicate Gen2 behavior
-        config.connections.warpType = config.warpType
-        config.connections.warpTypeSuffix = makeCoaddSuffix(config.warpType)
+        super().__init__(config=config)
 
+        # Override the connection's name template with config to replicate Gen2 behavior
+        # This duplicates some of the logic in the base class, due to waning Gen2 and
+        # Gen3 configs to stay in sync. This should be removed when gen2 is deprecated
+        templateValues = {name: getattr(config.connections, name) for name in self.defaultTemplates.keys()}
```
timj, line 102: is `.keys()` needed here? If `self.defaultTemplates` is dict-like then it shouldn't be there. I have the same comment on the `self.allConnections.keys()` line below.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -94,14 +94,20 @@ class AssembleCoaddConnections(pipeBase.PipelineTaskConnections,
     )
 
     def __init__(self, *, config=None):
-        # Override the connection's name template with config to replicate Gen2 behavior
-        config.connections.warpType = config.warpType
-        config.connections.warpTypeSuffix = makeCoaddSuffix(config.warpType)
+        super().__init__(config=config)
 
+        # Override the connection's name template with config to replicate Gen2 behavior
+        # This duplicates some of the logic in the base class, due to waning Gen2 and
```
timj, line 100: waning - > wanting

natelust:
fixed and amended.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -94,14 +94,20 @@ class AssembleCoaddConnections(pipeBase.PipelineTaskConnections,
...
+        templateValues['warpTypeSuffix'] = makeCoaddSuffix(config.warpType)
         if config.hasFakes:
-            config.connections.fakesType = "_fakes"
-
-        super().__init__(config=config)
+            templateValues['fakesType'] = "_fakes"
+        self._nameOverrides = {name: getattr(config.connections, name).format(**templateValues)
+                               for name in self.allConnections}
+        self._typeNameToVarName = {v: k for k, v in self._nameOverrides.items()}
+        # End code to remove after deprecation
```
TallJimbo, line 110: I can live with this if it's what we need to freeze configs without breaking existing on-disk configs, but I'm pretty uncomfortable with it.  How about instead we redefine `warpType` and `hasFakes` in `AssembleCoaddConfig` to
```
class AssembleCoaddConfig(...):
    ...
    @property
    def warpType(self):
        return self.connections.warpType

    @warpType.setter
    def warpType(self, value):
        self.connections.warpType = value
        self.connections.warpTypeSuffix = makeCoaddSuffix(self.warpType)

    _hasFakes = pexConfig.Field(
        dtype=bool,
        default=False,
        doc="Should be set to True if fake sources have been inserted into the input data."
    )

    @property
    def hasFakes(self):
        return self._hasFakes

    @hasFakes.setter
    def hasFakes(self, value):
        if value:
            self.connections.fakesType = "_fakes"
        else:
            self.connections.fakesType = ""
        self._hasFakes = value
```

I _think_ that achieves the same goals, while being the easier to understand?

timj: I like the look of that and I think it will keep @erykoff happy.

natelust: I am not 100% certain this is in general more readable. I understand it because I understand getters and setters and such, but it adds a layer to config files that practically no other configs have. You need to understand property getters and setters, and then  understand that these are actually instance methods, where most things declared on configs are fields. I do agree that what is in the init will not be comprehensible to some people reading it, I think the number of people looking at the init method of the connections class will be small, and they will likely be happy with the comment block (which will go away with gen2). People looking at the config class (which I contend is a much larger group of people) will find it more mysterious if they dont understand this aspect of python (which we could help with comments too). 

All that said I am mostly fine with either.

natelust: Why does it keep eli happy more than the other solution, because it is more visible?

timj: I believe he wanted the config to still have both values whereas in the new version the connections are used but the config never sees them.

erykoff: My comment/complaint was that we have two ways of setting the same thing, and the Gen3 one is going to supplant the Gen2 one (as I understand it).  So therefore, we _have_ to set the Gen3 versions.  And at some point (now?) we have to have a check that the Gen2 and Gen3 versions are in sync.  Simply overriding the gen3 value with the gen2 value is going to work until we cut out the gen2 code and then everybody's gen3 configs will suddenly break.  (Which I think is worse than updating configs now).  I'm not sure that either of these really addresses what I'm concerned with.

timj: I admit I don't understand this Task so don't really understand the concerns. Fixing this as @TallJimbo suggests isn't going to change anything for anybody as far as I can tell. It was a hack before and a hack now. I think this ticket can continue with the hack and science pipelines can worry about config migration some other time.

natelust: I don't have time to fully comment on this now, but either you or I are misunderstanding the situation, and I want to get the cleared up before this merges. So for now I am going to put this ticket back in review from being marked as reviewed.

What my code is doing, is the same as what Jims is doing, is the same as what is currently happening in master, none of which is what Eli wants. What Eli wants though will break existing configs which is not something we are allowed to do, and certainly not on a ticket like this.

I will be back around later to write more.

erykoff: I understand the concerns about what I'm suggesting, and it probably isn't for this ticket, and so I don't especially care about the swizzling involved.  

But I would argue that our configs are already broken if we have config variables that have no effect and are overwritten silently.

natelust: > My comment/complaint was that we have two ways of setting the same thing, and the Gen3 one is going to supplant the Gen2 one (as I understand it). So therefore, we _have_ to set the Gen3 versions. And at some point (now?) we have to have a check that the Gen2 and Gen3 versions are in sync. Simply overriding the gen3 value with the gen2 value is going to work until we cut out the gen2 code and then everybody's gen3 configs will suddenly break. (Which I think is worse than updating configs now). I'm not sure that either of these really addresses what I'm concerned with.

no in all these versions the gen2 config is used to set the value of the gen3 config, and whatever the user set for the gen3 config is just lost. I just realized Jims version is a bit different because it depends on the order that one sets the config override in the file as to which one "wins" which is just even more confusing, and so I dont support it.

natelust: @timj jim's solution is broken and we should not use it.

TallJimbo: Agree that mine is broken; we'd need to also make the .connections options into properties to make it work.  Just merge @natelust 's fix for now, and we'll clean this all up on DM-26484 or some related ticket.

timj: Fine with me. I can merge once @natelust unblocks the jira ticket.

natelust: Just unblocked it

PR #422, opened 2020-10-27
DM-5304: Use psfex as the default PSF determiner

parejkoj:


tests/test_processCcd.py
```diff
@@ -132,20 +132,22 @@ def testProcessCcd(self):
...
                     expectedPlaces = 7  # Tolerance for numerical comparisons
                     for name, var, val in [
-                        ("bgMean", bgMean, 191.48635852060525),
-                        ("bgStdDev", bgStdDev, 0.2399466881603354),
-                        ("numGoodPix", numGoodPix, 1966820),
-                        ("imMean", imMean, 1.1237668985230562),
-                        ("imStdDev", imStdDev, 85.81296241298496),
+                        ("bgMean", bgMean, 191.48623786891795),
+                        ("bgStdDev", bgStdDev, 0.23994185672586282),
+                        ("numGoodPix", numGoodPix, 1966606),
```
arunkannawadi, line 143: It's frankly beyond this particular PR, but I don't quite follow why ``bgMean``, ``bgStdDev``, ``numGoodPix`` aren't being tested in the `else` condition.

tests/test_processCcd.py
```diff
@@ -132,20 +132,22 @@ def testProcessCcd(self):
...
+                        ("imMean", imMean, 1.1242456954648634),
+                        ("imStdDev", imStdDev, 85.8129750182329),
                         ("varMean", varMean, 131.24003624152013),
                         ("varStdDev", varStdDev, 55.98012493452948),
-                        ("psfIxx", psfIxx, 2.769679536557131),
-                        ("psfIyy", psfIyy, 2.2013649766299324),
-                        ("psfIxy", psfIxy, 0.14797939531970852)
+                        ("psfIxx", psfIxx, 2.843329671276296),
+                        ("psfIyy", psfIyy, 2.2249941554078156),
+                        ("psfIxy", psfIxy, 0.16073332780683286)
```
arunkannawadi, line 150: Could you expand the comment above these lines to indicate how to obtain these values in the event of significant algorithm changes in the future?

arunkannawadi (commented):
I'm not quite sure why so many values need to be hard-coded. Couldn't we compare all the objects with the first object, i.e., i=0 without explicitly listing them?

parejkoj:
I don't want to refactor this test at all: it's a purely empirical test. I'm not a fan of how the test is implemented, but I'd much rather not dig into it here.

arunkannawadi:
Sure, I didn't necessarily mean to refactor the test. It'd be helpful to write some instructions as to how you came up with those numbers. If I made an algorithmic change that makes this test fail, I wouldn't know how to find the new set of numbers.

arunkannawadi:
Okay, I see the comments now.

arunkannawadi (approved):

PR #423, opened 2020-11-02
DM-24331 Replace deprecated getCentroidKey calls.

kgabor:


parejkoj (approved):
Huh. Don't know how I missed this with my search, but it likely is stale code. Thanks for catching it.

PR #425, opened 2020-11-03
DM-26615: remove afw filter wavelength dependence from DCR model

isullivan:


taranu (approved):
Looks fine again; consider putting units in the fields if they're required to be nm.

PR #426, opened 2020-11-05
DM-24638: Convert TransformSourceTableTask and friends to Gen3

yalsayyad:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -204,7 +204,25 @@ def writeMetadata(self, dataRefList):
...
+        storageClass="SourceCatalog",
+        dimensions=("instrument", "visit", "detector")
+    )
+    outputCatalog = pipeBase.connectionTypes.Output(
+        doc="Catalog of sources, `src` in Parquet format",
+        name="source",
+        storageClass="DataFrame",
+        dimensions=("instrument", "visit", "detector")
+    )
+
```
erykoff, line 222: So for DM-27164 (and this will need to be rebased of course), I defined `from lsst.pipe.base import connectionTypes` so you can use just `connectionTypes` here.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -436,7 +461,24 @@ def compute(self, dropna=False, pool=None):
...
 
-class TransformCatalogBaseConfig(pexConfig.Config):
+class TransformCatalogBaseConnections(pipeBase.PipelineTaskConnections,
+                                      dimensions=()):
+    """Expected Connections for subclasses of TransformCatalogBaseTask.
+
+    Must be subclassed.
+    """
+    inputCatalog = pipeBase.connectionTypes.Input(
+        name="",
```
erykoff, line 471: Same here for `connectionTypes`

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -532,6 +574,21 @@ def inputDataset(self):
...
+        if self.config.functorFile:
+            self.log.info('Loading tranform functor definitions from %s',
+                          self.config.functorFile)
+            self.funcs = CompositeFunctor.from_file(self.config.functorFile)
+            self.funcs.update(dict(PostprocessAnalysis._defaultFuncs))
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        result = self.run(parq=inputs['inputCatalog'], funcs=self.funcs,
+                          dataId=outputRefs.outputCatalog.dataId)
```
erykoff, line 588: I'm not sure if using dataId in this way is really kosher in Gen3.  Specifically, `filter` is now `physical_filter`.  So in this case, I think that this won't work correctly.  I'm trying to follow how `dataId` gets used downstream, if there's anything else that needs to come from it; if it's just returning the dataId in the struct that seems like a gen2 necessity not gen3 (though I could be wrong).

What I would suggest is specifically pulling out `physical_filter` for gen3 (and `filter` for gen2), and having that as a separate keyword which seems special and necessary.  And then I think the `dataId` can be set only in the gen2 code run.  Though I may be missing one of the uses here.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -762,7 +817,27 @@ def writeMetadata(self, dataRef):
         pass
 
 
-class TransformSourceTableConfig(TransformCatalogBaseConfig):
+class TransformSourceTableConnections(pipeBase.PipelineTaskConnections,
+                                      dimensions=("instrument", "visit", "detector")):
+
+    inputCatalog = pipeBase.connectionTypes.Input(
+        doc="Wide input catalog of sources produced by WriteSourceTableTask",
```
erykoff, line 824: Same here about `connectionTypes`.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -827,11 +897,29 @@ def ccdDataRefList(visitId):
         self.refList = outputRefList
 
 
-class ConsolidateSourceTableConfig(pexConfig.Config):
+class ConsolidateSourceTableConnections(pipeBase.PipelineTaskConnections,
+                                        dimensions=("instrument", "visit")):
+    inputCatalogs = pipeBase.connectionTypes.Input(
+        doc="Input per-detector Source Tables",
```
erykoff, line 903: And here.

erykoff (commented):
There are some updates for `physical_filter` replacing `filter` that need to be done, but otherwise looks good.  The default pipeline file should also be updated (this will be clear after a rebase).

python/lsst/pipe/tasks/postprocess.py
yalsayyad: It's not obvious to me that `connectionTypes` is more readable than `pipeBase.connectionTypes` when there are other `pipeBase` objects abounding, but sure, I'll change them all.

python/lsst/pipe/tasks/functors.py
```diff
@@ -296,8 +297,12 @@ def __call__(self, parq, **kwargs):
                 except Exception:
                     valDict[k] = f.fail(subdf)
         else:
-            columns = self.columns
-            df = parq.toDataFrame(columns=columns)
+            if isinstance(parq, DeferredDatasetHandle):
+                df = parq.get(columns=self.columns)
+            elif isinstance(parq, pd.DataFrame):
+                df = parq
+            else:
```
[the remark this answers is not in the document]
yalsayyad: The `parq` instance is for the Gen3 case if `deferLoad=False`.  The inputs datatypes are pd.DataFrames after loading. and the `parq.toDataFrame` is the original (Gen2). I'll add inline comments.

python/lsst/pipe/tasks/postprocess.py
erykoff: I don't want to makework!  But we don't have any standard convention on this, and that's unpleasant.

python/lsst/pipe/tasks/postprocess.py
yalsayyad: Ah right, because the regular dataId depends on what you pass on the command-line. OK I changed it `outputRefs.outputCatalog.dataId.full` so that it'll at least be the same keys are being added to the table every time.  Technically, all this info is redundant: everything will eventually be looked up with the "cccVisitId."  I've been adding these columns because we don't have a corresponding `ccdVisit` table yet to look up the full ccd info given a ccdVisitId. 

Passing an optional dataId for logging was kosher last I checked.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -566,14 +623,12 @@ def run(self, parq, funcs=None, dataId=None, band=None):
...
-        funcs = CompositeFunctor.from_file(self.config.functorFile)
-        funcs.update(dict(PostprocessAnalysis._defaultFuncs))
-        return funcs
+        return self.funcs
 
     def getAnalysis(self, parq, funcs=None, band=None):
         # Avoids disk access if funcs is passed
         if funcs is None:
-            funcs = self.getFunctors()
+            funcs = self.funcs
```
natelust, line 631: same

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -566,14 +623,12 @@ def run(self, parq, funcs=None, dataId=None, band=None):
         return df
 
     def getFunctors(self):
-        funcs = CompositeFunctor.from_file(self.config.functorFile)
-        funcs.update(dict(PostprocessAnalysis._defaultFuncs))
-        return funcs
+        return self.funcs
```
natelust, line 626: same

natelust (commented):
Sorry for the drive by review, just caught my eye while I was waiting for a batch job and looking at something else in my email. You might have this well handled already, just something to think about.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -530,6 +572,21 @@ def inputDataset(self):
...
+        super().__init__(*args, **kwargs)
+        if self.config.functorFile:
+            self.log.info('Loading tranform functor definitions from %s',
+                          self.config.functorFile)
+            self.funcs = CompositeFunctor.from_file(self.config.functorFile)
+            self.funcs.update(dict(PostprocessAnalysis._defaultFuncs))
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        result = self.run(parq=inputs['inputCatalog'], funcs=self.funcs,
```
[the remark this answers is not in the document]
yalsayyad: `self.config.functorFile` is optional in the sense that you should be able to instantiate the task in the notebook and run any functors you want (not necessarily from a yaml); and there's no default set of functors that make sense for the base class.

`self.config.functorFile` is not optional when running the subclasses as a commandlineTask or pipelineTask. I'll add an appropriate validation error message somewhere.

pipelines/_SingleFrame.yaml
```diff
@@ -3,6 +3,8 @@ tasks:
   isr: lsst.ip.isr.IsrTask
   characterizeImage: lsst.pipe.tasks.characterizeImage.CharacterizeImageTask
   calibrate: lsst.pipe.tasks.calibrate.CalibrateTask
+  writeSourceTable: lsst.pipe.tasks.postprocess.WriteSourceTableTask
+  transformSourceTable: lsst.pipe.tasks.postprocess.TransformSourceTableTask
```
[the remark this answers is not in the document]
yalsayyad: That makes sense to mirror what we had before. Can subsets refer to each other? e.g.
```
subsets:
  processCcd:
      - isr
      - characterizeImage
      - calibrate
  singleFrame:
    subset:
      - processCcd
      - writeSourceTable
      - transformSourceTable
    description: Single frame processing that includes table transformations
```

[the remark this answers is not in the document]
yalsayyad: Couldn't get the DECam repo going in time to add for all cameras, so @natelust take a look at https://github.com/lsst/obs_subaru/pull/341/files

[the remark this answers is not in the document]
natelust: So currently this is the only place that labeled subsets cant be used as a substitute for labels. This is because at the time we didnt think it was worth adding code and complexity to track and handle cyclical definitions, to save a few lines in a (mostly) static file. If you think this is a good feature to have, we can add it.

pipelines/_Forced.yaml
```diff
@@ -8,4 +8,3 @@ subsets:
       - forcedPhotCcd
       - forcedPhotCoadd
     description: A set of tasks to run when doing forced measurements
```
natelust, line 10: Does this pass github checks? I though yaml lint required a new line at the end (but maybe I have that backwards)

pipelines/_SingleFrame.yaml
```diff
@@ -3,10 +3,28 @@ tasks:
...
+    description: Set of tasks to generate parquet Source Tables from output of processCcd subset.
+  singleFrame:
+    subset:
+      - isr
+      - characterizeImage
+      - calibrate
+      - writeSourceTable
+      - transformSourceTable
+      - consolidateSourceTable
+    description: Set of tasks for complete single frame processing. Analogous to SingleFrameDriver.
```
natelust, line 30: I'm not sure if we should mention SingleFrameDriver in the docs, it does not help anyone unfamiliar with pipe_drivers, and its somewhat ties this package to that, potentially after the later is retired. It's up to your judgment though. If you feel it should be there then feel free to leave it.

natelust (approved):

pipelines/_Forced.yaml
yalsayyad: lint DOES require a newline at the end. You had 2 newlines before: https://github.com/lsst/pipe_tasks/blob/master/pipelines/_Forced.yaml

PR #427, opened 2020-11-06
DM-27008: Allow a DeferredDatasetHandle to be passed to a Functor

timothydmorton:
For gen3 compatibility, this allows for a DeferredDatasetHandle to be be passed to a functor's __call__ method, rather than the
gen2 parquet interface ParquetTable object.  All currently tested Functor functionality is also tested here for the gen3 model.  Backward compatibility with gen2 is maintained.  N.B.: the gen3 functor performance will likely be slower than the gen2 performance; remains to be seen if this matters at all.

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -1,12 +1,14 @@
 import yaml
 import re
+from itertools import product
 
 import pandas as pd
 import numpy as np
 import astropy.units as u
 
 from lsst.daf.persistence import doImport
-from .parquetTable import MultilevelParquetTable
+from lsst.daf.butler import DeferredDatasetHandle
+from .parquetTable import ParquetTable, MultilevelParquetTable
 
 
 def init_fromDict(initDict, basePath='lsst.pipe.tasks.functors',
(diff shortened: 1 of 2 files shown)


tests/test_functors.py
```diff
@@ -33,19 +33,36 @@
...
+    HsmFwhm,
+    LocalPhotometry,
+    LocalNanojansky,
+    LocalNanojanskyErr,
+    LocalMagnitude,
+    LocalMagnitudeErr,
+    LocalWcs,
+    ComputePixelScale,
+    ConvertPixelToArcseconds,
+)
```
sr525, line 65: Do these really all need to be on their own lines?

python/lsst/pipe/tasks/functors.py
```diff
@@ -280,24 +416,58 @@ def update(self, new):
...
+            set(
+                [
+                    x
+                    for y in [
+                        f.multilevelColumns(data, returnTuple=True, **kwargs) for f in self.funcDict.values()
+                    ]
+                    for x in y
+                ]
+            )
+        )
```
sr525, line 431: This was definitely easier to read in its previous format.

PR #428, opened 2020-11-10
DM-27168: Phase out use of FilterProperty

czwa:
End usage of FilterProperty, as per RFC-730.

Usage in dcrAssembleCoadd.py left as it is dealt with by DM-26615.

### DIFF
--- a/python/lsst/pipe/tasks/multiBandUtils.py
+++ b/python/lsst/pipe/tasks/multiBandUtils.py
@@ -138,13 +138,12 @@ def getInputSchema(task, butler=None, schema=None):
     return schema
 
 
+# TODO: DM-27170 should remove this function.
 def getShortFilterName(name):
     """Given a longer, camera-specific filter name (e.g. "HSC-I") return its shorthand name ("i").
     """
-    # I'm not sure if this is the way this is supposed to be implemented, but it seems to work,
-    # and its the only way I could get it to work.
     try:
-        return afwImage.Filter(name).getFilterProperty().getName()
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/multiBandUtils.py
```diff
@@ -144,7 +144,7 @@ def getShortFilterName(name):
     # I'm not sure if this is the way this is supposed to be implemented, but it seems to work,
     # and its the only way I could get it to work.
     try:
-        return afwImage.Filter(name).getFilterProperty().getName()
+        return afwImage.Filter(name).getName()
```
kfindeisen, line 147: I believe this should be `getCanonicalName()`; see the [definition of getFilterProperty()](https://github.com/lsst/afw/blob/master/src/image/Filter.cc#L359-L360).

kfindeisen (commented):
I'm a bit confused by your comment about `dcrAssembleCoadd.py`. DM-26615 has already been merged and is ancestral to this branch; are you still seeing residual use despite that?

python/lsst/pipe/tasks/multiBandUtils.py
kfindeisen: In fact, looking at the function definition, I'm pretty sure this could be entirely removed. Maybe remove the existing comment and mark it as TODO for DM-27170?

czwa:
The commit comment was fine before I rebased.  I've removed it as you mention it's inaccurate now.

kfindeisen (approved):

PR #436, opened 2020-11-24
DM-27684: Add butler register-skymap subcommand

n8pease:


timj:
I think we decided that this Task code should be pulled into the the script implementation and the Task removed since it was trivial. Also to remove the Config and move the config items to be command line options (I think there are only three of them)

cc/ @TallJimbo who made this suggestion initially and this was mentioned on the Jira ticket. Did I miss a discussion on this?

python/lsst/pipe/tasks/script/registerSkymap.py
```diff
@@ -0,0 +1,50 @@
...
+        Path to a config overrides file.
+
+    Raises
+    ------
+    RuntimeError
+        If a config overrides file is given and does not exist.
+    """
+    config = MakeGen3SkyMapConfig()
+    if config_file:
+        if not os.path.exists(config_file):
```
timj, line 45: Do we really need this check? The `config.load` will fail anyhow and tell you that the file can't be found. It will also give a FileNotFoundError which seems more accurate.

python/lsst/pipe/tasks/makeGen3SkyMap.py
```diff
@@ -43,33 +49,16 @@ def validate(self):
                              "skymap to use when writing to the butler")
 
 
-class MakeGen3SkyMapTask(pipeBase.Task):
-    ConfigClass = MakeGen3SkyMapConfig
-    _DefaultName = "makeGen3SkyMap"
-
-    """This is a task to construct and optionally save a SkyMap into a gen3
-    butler repository.
+def makeGen3SkyMap(butler, config):
```
timj, line 52: @TallJimbo do we want to continue to have Gen3 in the name for this function and the config?

python/lsst/pipe/tasks/cli/cmd/commands.py
```diff
@@ -37,20 +40,25 @@
...
               type=str, default=None)
+@options_file_option()
 def make_discrete_skymap(*args, **kwargs):
     """Define a discrete skymap from calibrated exposures in the butler registry."""
     cli_handle_exception(script.makeDiscreteSkyMap, *args, **kwargs)
+
+
+@click.command(cls=ButlerCommand)
+@repo_argument(required=True)
+@config_file_option(help="Path to a config file overrides file.")
```
timj, line 60: Should we also support `--config` like pipetask does?

timj (approved):

python/lsst/pipe/tasks/makeGen3SkyMap.py
TallJimbo: Let's drop the "Gen3" everywhere there isn't a direct clash with Gen2.  I think the only place there was a clash before was in the script name, so hopefully that means we can drop the Gen3s everywhere (around this script, at least).

python/lsst/pipe/tasks/cli/cmd/commands.py
n8pease: 🆗

PR #437, opened 2020-11-25
DM-27356: Implement a simple cutout task

SimonKrughoff:


python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,96 @@
...
+            pix = wcs.skyToPixel(pt)
+            # Clamp to LL corner of the LL pixel and draw extent from there
+            box = geom.Box2I(geom.Point2I(int(pix.x-size/2), int(pix.y-size/2)),
+                             geom.Extent2I(size, size))
+            if not mim.getBBox().contains(box):
+                if not self.config.skip_bad:
+                    raise ValueError(f'Cutout bounding box is not completely contained in the image: {box}')
+                else:
+                    continue
+            # I think we need to think about what we want the origin to be: LOCAL or PARENT
```
r-owen, line 90: Please file a ticket on this subject and reference the ticket number here.

But when in doubt: I suggest parent. It's more informative.

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,96 @@
...
+            sub = mim.Factory(mim, box)
+            stamp = Stamp(stamp_im=sub, position=pt)
+            cutout_list.append(stamp)
+        # We are using this stamp container as a place holder because it already has
+        # the storage class defined
```
r-owen, line 95: Placeholder for what? Again...please file a ticket (possibly the same one) and reference it here.

r-owen (approved):
This looks nice.

I have many requests for documentation enhancement and would also like to suggest taking a numpy structured array instead of a dict (or some kind of table) for the coordinates and sizes, if that is doable without a lot of work, because it is a much more natural fit to the data.

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,96 @@
...
+    def run(self, in_cat, calexp):
+        """Compute and return the cutouts.
+
+        Parameters
+        ----------
+        in_cat : `dict`
+            A dictionary containing at least the following columns: ra, dec, size.
+            The coordinates should be in ICRS degrees.  The size is in pixels.
+        calexp : `lsst.afw.image.ExposureF`
+            The calibrated exposure from which to extract cutouts
```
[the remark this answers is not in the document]
SimonKrughoff: For clarification, the script that makes the input table that is used by this pipeline task can't depend on `afw`, but this can.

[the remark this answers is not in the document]
SimonKrughoff: The intent is that this task will run on some distributed system that has the full stack installed and is simply triggered by the running service with the appropriate information for this task to find the necessary inputs

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,96 @@
...
+        dimensions=("instrument", 'visit', 'detector'),
+    )
+
+
+class CalexpCutoutTaskConfig(pipeBase.PipelineTaskConfig,
+                             pipelineConnections=CalexpCutoutTaskConnections):
+    """Configuration for CalexpCutoutTask
+    """
+    max_cutouts = Field(dtype=int, default=100, doc='Maximum number of entries to process. '
+                                                    'The result will be the first N in the catalog.')
```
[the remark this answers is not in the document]
SimonKrughoff: I'm attempting to take a numpy array and will update the documentation accordingly

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,96 @@
...
+    skip_bad = Field(dtype=bool, default=True, doc='Skip cutouts that do not fall completely within'
```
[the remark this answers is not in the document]
SimonKrughoff: I too hope we can live without the added complexity.  I'm not sure how useful partial cutouts are.

python/lsst/pipe/tasks/calexpCutout.py
SimonKrughoff: I just went with the default.  I'm happy to change if you think it would be better

python/lsst/pipe/tasks/calexpCutout.py
SimonKrughoff: That is outdated.  I have implemented the container class.

tests/data/cutout_data.json
```diff
@@ -0,0 +1 @@
+{"good": {"ra": [215.56293259755142, 215.5192147097086, 215.53885833546383, 215.5482202222798, 215.56314911887543, 215.58006865017555, 215.54844934953582, 215.57238596481827, 215.52166218322103, 215.53640511652264], "dec": [53.16981748080182, 53.17431015949814, 53.15315235360539, 53.16044636925184, 53.16869944567883, 53.158855254327655, 53.15563347471535, 53.13767921007825, 53.152238102494636, 53.17685157361139], "size": [14, 24, 19, 22, 18, 16, 17, 22, 21, 13]}, "edge": {"ra": [215.5056177083067], "dec": [53.18742661682197], "size": [18]}, "bad": {"ra": [215.41966428289527], "dec": [53.23892939122016], "size": [23]}}
```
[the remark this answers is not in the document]
SimonKrughoff: I have done as you suggest and implemented the test data creation in the unit test

tests/test_calexpCutout.py
```diff
@@ -0,0 +1,100 @@
...
+datadir = os.path.join(packageDir, 'tests', "data")
+
+
+class CalexpCutoutTestCase(lsst.utils.tests.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        with open(os.path.join(datadir, 'cutout_data.json'), 'r') as fh:
+            cls.data = json.load(fh)
+        cls.exp = ExposureF.readFits(os.path.join(datadir, 'v695833-e0-c000-a00.sci.fits'))
```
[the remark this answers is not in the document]
SimonKrughoff: Done

SimonKrughoff (commented):
I'm not sure I should have done this in a review, but what's done is done

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,113 @@
...
+            pix = wcs.skyToPixel(pt)
+            size = rec['size'].value
+            # Clamp to LL corner of the LL pixel and draw extent from there
+            box = geom.Box2I(geom.Point2I(int(pix.x-size/2), int(pix.y-size/2)),
+                             geom.Extent2I(size, size))
+            if not mim.getBBox().contains(box):
+                if not self.config.skip_bad:
+                    raise ValueError(f'Cutout bounding box is not completely contained in the image: {box}')
+                else:
+                    continue
```
r-owen, line 106: I strongly suggest storing metadata about the skipped cutouts. At the very least a count, but a list of coordinates would be better.

As a subtlety: it would naturally be empty if you raise an exception but I suggest returning it anyway, in case you ever have a need to handle handle partial cutouts differently than fully off-the-CCD cutouts.

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,96 @@
...
+    _DefaultName = "calexpCutoutTask"
+
+    def run(self, in_cat, calexp):
+        """Compute and return the cutouts.
+
+        Parameters
+        ----------
+        in_cat : `dict`
+            A dictionary containing at least the following columns: ra, dec, size.
+            The coordinates should be in ICRS degrees.  The size is in pixels.
```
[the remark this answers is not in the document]
r-owen: I think "colums" is fine for a structured array, though I don't actually know the correct idiom. Since you ended up with a table, columns seems fully appropriate.

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Returns
+        -------
+        output : `lsst.pipe.base.Struct`
+            A struct containing a `lsst.meas.algorithms.Stamps` object
+            that wraps a list of masked images of the cutouts and a
+            `PropertyList` containing the metadata to be persisted
+            with the cutouts.  The exposure metadata is preserved and,
+            in addition, arrays holding the RA and Dec of each stamp
+            in degrees are added to the metadata.
```
r-owen, line 72: I strongly suggest giving the field names (the user should not have to read the code to find these). For example:
```
A struct containing:

* cutouts: an `lsst.meas.algorithms.Stamps` object that wraps ...
* metadata: an `lsst.daf.base.PropertyList` containing ...
```

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,113 @@
...
+            # Clamp to LL corner of the LL pixel and draw extent from there
+            box = geom.Box2I(geom.Point2I(int(pix.x-size/2), int(pix.y-size/2)),
+                             geom.Extent2I(size, size))
+            if not mim.getBBox().contains(box):
+                if not self.config.skip_bad:
+                    raise ValueError(f'Cutout bounding box is not completely contained in the image: {box}')
+                else:
+                    continue
+            sub = mim.Factory(mim, box)
+            stamp = Stamp(stamp_im=sub, position=pt)
```
r-owen, line 108: You say you "went with the default" for PARENT vs LOCAL, but what is the default? I think PARENT is the best choice, since it works with the WCS for the calexp.

tests/test_calexpCutout.py
```diff
@@ -0,0 +1,145 @@
...
+from lsst.afw.image import ExposureF
+from lsst.pipe.tasks.calexpCutout import CalexpCutoutTask
+
+random.seed(208241138)
+
+packageDir = lsst.utils.getPackageDir('pipe_tasks')
+datadir = os.path.join(packageDir, 'tests', "data")
+
+
+def make_data(bbox, wcs, border=100, ngood=13, nbad=1):
```
r-owen, line 44: Why no an nedge parameter? Not a big deal but it is a surprising omission.

r-owen (approved):
Very nice. Thanks for switching to a table.

My main remaining requests which I hope you will consider:
* Return metadata about skipped cutouts, e.g. a list of sky points that were skipped.
* Document what the "size" column means. A trivial addition that will help users.
* Explicitly name the fields in the struct returned by `run`. Another trivial change that will help users.

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -0,0 +1,113 @@
...
+    _DefaultName = "calexpCutoutTask"
+
+    def run(self, in_table, calexp):
+        """Compute and return the cutouts.
+
+        Parameters
+        ----------
+        in_table : `astropy.QTable`
+            A table containing at least the following columns: position, size.
+            The position should be an `astropy.SkyCoord`.  The size is in pixels.
```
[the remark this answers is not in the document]
SimonKrughoff: Aha.  Sorry for missing that bit.  Will do.

python/lsst/pipe/tasks/calexpCutout.py
SimonKrughoff: Ergh.  I didn't mean to skip that request.  I'll do that.

tests/test_calexpCutout.py
SimonKrughoff: I was lazy and didn't want to figure out how to sample random points on the edge.  I can add it.

r-owen: Eh, no need. Not that it should be hard, but...not very important.

python/lsst/pipe/tasks/calexpCutout.py
r-owen: I have no idea what the default is. I think PARENT would be best. If that's the default then fine. Or if you have a good reason to prefer LOCAL. 

Given the ambiguity I suggest documenting your choice in the doc string for this task!

python/lsst/pipe/tasks/calexpCutout.py
SimonKrughoff: OK. I've confirmed the default is PARENT.  I'll mention this in the docstring.

SimonKrughoff:
OK, I've done everything you've asked (I hope).  The only caveat is that I'm returning a list of positions of cutouts that *were* skipped.  I think this is actually what you intended in your comment.

r-owen:
That sounds fine. Any sort of list of missed patches seems reasonable (RA/Dec or pixel positions).

PR #448, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


python/lsst/pipe/tasks/coaddInputRecorder.py
```diff
@@ -112,7 +112,7 @@ def addCalExp(self, calExp, ccdId, nGoodPix):
             self._setExposureInfoInRecord(exposure=calExp, record=record)
             if self.task.config.saveCcdWeights:
                 record.setD(self.task.ccdWeightKey, 1.0)  # No weighting or overlap when warping
-            record.set(self.task.ccdFilterKey, calExp.getFilter().getName())
+            record.set(self.task.ccdFilterKey, calExp.getFilterLabel().bandLabel)
```
kfindeisen, line 115: This one's a bit problematic, since IIRC the `obs_subaru` coadd code depends on this key distinguishing HSC-R from HSC-R2 (previously accomplished through the use of `afw_name`). So I think you may actually need the physical filter here.

python/lsst/pipe/tasks/coaddInputRecorder.py
```diff
@@ -181,7 +181,7 @@ def __init__(self, *args, **kwargs):
         self.visitFilterKey = self.visitSchema.addField("filter", type=str, size=32,
                                                         doc="Filter associated with this visit.")
         self.ccdFilterKey = self.ccdSchema.addField("filter", type=str, size=32,
-                                                    doc="Filter associated with this visit.")
+                                                    doc="Filter (band) associated with this visit.")
```
kfindeisen, line 184: See above. The description should probably be synchronized with `self.visitFilterKey`.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -953,9 +953,8 @@ def _combineExposureMetadata(self, visit, dataRefs, isGen3=True):
             rec.setDetector(detector)
             rec.setValidPolygon(validPolygon)
 
-            # TODO: When RFC-730 is implemented we can fill both of these.
-            rec['physical_filter'] = filter_.getName()
-            rec['band'] = ''
+            rec['physical_filter'] = filterLabel.physicalLabel
+            rec['band'] = filterLabel.bandLabel
```
kfindeisen, line 957: Should `None` be converted to `''`?

python/lsst/pipe/tasks/colorterms.py
```diff
@@ -238,17 +236,11 @@ def getColorterm(self, filterName, photoCatName, doRaise=True):
                         "No colorterm dict found with photoCatName %r" % photoCatName)
             ctDict = ctDictConfig.data
             if filterName not in ctDict:
-                # Perhaps it's an alias
-                try:
-                    filterName = Filter(Filter(filterName).getId()).getName()
```
kfindeisen, line 243: 😵

python/lsst/pipe/tasks/mergeDetections.py
```diff
@@ -230,8 +230,8 @@ def __init__(self, butler=None, schema=None, initInputs=None, **kwargs):
         self.makeSubtask("skyObjects")
         self.schema = self.getInputSchema(butler=butler, schema=schema)
 
-        filterNames = [getShortFilterName(name) for name in self.config.priorityList]
-        filterNames += [self.config.skyFilterName]
+        filterNames = list(self.config.priorityList)
+        filterNames.append(self.config.skyFilterName)
```
kfindeisen, line 234: What do the configs for `priorityList` look like in practice? Do they use bands to begin with?

tests/test_coaddInputs.py
```diff
@@ -107,6 +107,7 @@ def makeExposure(self, universalId):
         expInfo.setPhotoCalib(lsst.afw.image.makePhotoCalibFromCalibZeroPoint(1.1e12, 2.2e10))
         expInfo.setApCorrMap(self.makeApCorrMap())
         expInfo.setValidPolygon(lsst.afw.geom.Polygon(lsst.geom.Box2D(bbox).getCorners()))
+        expInfo.setFilterLabel(lsst.afw.image.FilterLabel(physical="fakeFilter", band="fake"))
```
kfindeisen, line 110: Is this needed, given that the original test didn't set a `Filter`?

kfindeisen (changes requested):
Most of the changes look good, but I think the changes to coadd inputs will break `obs_subaru` (see discussion on lsst/obs_subaru#333).

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -829,7 +829,7 @@ def assembleMetadata(self, coaddExposure, tempExpRefList, weightList):
                            for tempExpRef in tempExpRefList]
         numCcds = sum(len(tempExp.getInfo().getCoaddInputs().ccds) for tempExp in tempExpList)
 
-        coaddExposure.setFilter(tempExpList[0].getFilter())
+        coaddExposure.setFilterLabel(tempExpList[0].getFilterLabel())
```
kfindeisen, line 832: I just realized that this will give coadds a physical filter (in general). I'm fine with that (especially if all the inputs use exactly the same filter), but I think it was argued at some point that coadds should be by band?

RobertLuptonTheGood: I think that we do need to use band.  For HSC this shows up in e.g. i and i2 physical filters;  for Rubin it'll be ITL and E2V CCDs, which have different effective bandpasses.

kfindeisen: Sorry, just to be clear, are you saying the code should _remove_ any physical filter in the `FilterLabel` it assigns to `coaddExposure`? A band will be present either way.

parejkoj: @RobertLuptonTheGood , following up on @kfindeisen 's comment: should coadds *only* have `band`, or can they have both `band` and `physical` (as this change implements)?

I also don't understand your comment about different CCDs: that has nothing to do with the physical/band FilterLabels.

python/lsst/pipe/tasks/colorterms.py
parejkoj: I'm not sure what to make of the face? Are you mad at what I removed, or that it was removed?

For the record, the `pass  # this will be handled shortly` just below this was added 6 years ago, so "shortly" in astronomical time, I guess.

python/lsst/pipe/tasks/postprocess.py
parejkoj: Done, but this class has no tests. :-(

tests/test_coaddInputs.py
parejkoj: Yes: ExposureInfo always had a default Filter before (it was the one non-shared_ptr in there), so the code was getting by with a default constructed Filter until now.

python/lsst/pipe/tasks/colorterms.py
kfindeisen: 😵  = "It makes my head spin that the old code made a `Filter` in order to make a `Filter` in order to standardize the filter name."

python/lsst/pipe/tasks/assembleCoadd.py
TallJimbo: > Sorry, just to be explicit, are you saying the code should remove any physical filter in the `FilterLabel` it assigns to `coaddExposure`? A band should be present either way.

This isn't something I'd considered before now, but I do think it would probably be best to clear the physical filter label from any coadd image, since it isn't always going to be meaningful and we don't have a clear use case for it.  

I think I probably would continue to use `physical_filter` in coaddInputs, however, as that will convey more information, and I can't think of any case where a single `coaddInput` record might have an ambiguous `physical_filter`.

I think we're already propagating ITL vs E2V via the `Detector` object that's also attached to the coaddInputs records (we certainly could be; there is definitely a slot for it).  And we're already propagating the combined throughput via the attached `TransmissionCurve` (which are actually combine quantitatively in coaddition, unlike the more descriptive `FilterLabel` and `Detector` information).

python/lsst/pipe/tasks/mergeDetections.py
parejkoj: It was physical in HSC, but band in CFHT and DECam. @TallJimbo stated on slack "I believe in Gen3 the priorityList is already band, and has to be, because coadds always use band."

I am changing it to `band` and updating obs_subaru's priorityList entries to reflect this.

python/lsst/pipe/tasks/coaddInputRecorder.py
parejkoj: This is now correctly `physical_filter`, but that opened up a whole host of other questions that @yalsayyad helped me sort out.

python/lsst/pipe/tasks/multiBandUtils.py
```diff
@@ -164,12 +164,12 @@ def readCatalog(task, patchRef):
...
     @param[in]  patchRef   data reference for patch
-    @return tuple consisting of the filter name and the catalog
+    @return tuple consisting of the band name and the catalog
     """
-    filterName = patchRef.dataId["filter"]
+    band = patchRef.get(task.config.coaddName + "Coadd_filterLabel", immediate=True).bandLabel
     catalog = patchRef.get(task.config.coaddName + "Coadd_" + task.inputDataset, immediate=True)
-    task.log.info("Read %d sources for filter %s: %s" % (len(catalog), filterName, patchRef.dataId))
-    return filterName, catalog
+    task.log.info("Read %d sources for filter %s: %s" % (len(catalog), band, patchRef.dataId))
```
kfindeisen, line 171: Should the text say "band" instead of "filter"? Same for `postprocess.py`.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -38,14 +38,14 @@
...
 
-def flattenFilters(df, filterDict, noDupCols=['coord_ra', 'coord_dec'], camelCase=False):
+def flattenFilters(df, noDupCols=['coord_ra', 'coord_dec'], camelCase=False):
     """Flattens a dataframe with multilevel column index
     """
     newDf = pd.DataFrame()
-    for filt, filtShort in filterDict.items():
-        subdf = df[filt]
+    for band in set(df.columns.to_frame()['band']):
+        subdf = df[band]
```
kfindeisen, line 46: I'm not familiar with Dataframes or with `postprocess.py`. Is it guaranteed that `df.columns.to_frame()` contains `'band'`? Or does a lookup of a nonexistent column(?) return an empty collection?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -555,6 +551,8 @@ def run(self, parq, funcs=None, dataId=None):
             Functors to apply to the table's columns
         dataId : dict, optional
             Used to add a `patchId` column to the output dataframe.
+        band : `str`, optional
+            Filter band that is being processed.
```
kfindeisen, line 555: Is this ever passed to `run`? `runDataRef` doesn't use it, and it looks like this task doesn't support Gen 3 yet.

python/lsst/pipe/tasks/functors.py
```diff
@@ -73,18 +73,18 @@ class Functor(object):
     * `name`: A name appropriate for a figure axis label
     * `shortname`: A name appropriate for use as a dictionary key
 
-    On initialization, a `Functor` should declare what filter (`filt` kwarg)
+    On initialization, a `Functor` should declare what band (`filt` kwarg)
```
kfindeisen, line 76: In other places you renamed `filt` to `band`. Do you want to do the same here? It's a breaking change in both cases, mind you.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -573,15 +570,15 @@ def getFunctors(self):
         funcs.update(dict(PostprocessAnalysis._defaultFuncs))
         return funcs
 
-    def getAnalysis(self, parq, funcs=None, filt=None):
+    def getAnalysis(self, parq, funcs=None, band=None):
```
kfindeisen, line 573: Breaking API change...

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -607,18 +604,28 @@ class TransformObjectCatalogConfig(TransformCatalogBaseConfig):
...
+             "input data actually contain."),
+        deprecated=("Coadds are now identified by the band, so this transform is unused."
+                    "Will be removed after v22.")
+    )
+    outputBands = pexConfig.ListField(
+        dtype=str,
+        default=[],
+        doc=("These bands and only these bands will appear in the output,"
+             " NaN-filled if the input does not include them."
+             " If an empty list, then just send bands from the input to the output.")
```
kfindeisen, line 623: I don't like the special-casing of how `[]` means the _opposite_ of the general spec. Maybe use `None` as the special value instead? I think it would be less confusing.

kfindeisen: I also suggest rewording "then just send bands from the input to the output" to "then use all bands found in the input" (assuming that's what you meant).

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -630,7 +637,8 @@ class TransformObjectCatalogConfig(TransformCatalogBaseConfig):
 
 
 class TransformObjectCatalogTask(TransformCatalogBaseTask):
-    """Compute Flatted Object Table as defined in the DPDD
+    """Produce a flattened Object Table to match the format specified in
+    sdm_schemas.
```
kfindeisen, line 641: English is fun. 😆

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -657,34 +665,32 @@ def run(self, parq, funcs=None, dataId=None):
...
         analysisDict = {}
         templateDf = pd.DataFrame()
-        # Perform transform for data of filters that exist in parq and are
-        # specified in config.filterMap
-        for filt in parq.columnLevelNames['filter']:
-            if filt not in self.config.filterMap:
-                self.log.info("Ignoring %s data in the input", filt)
+        # Perform transform for data of filters that exist in parq.
+        for band in parq.columnLevelNames['band']:
+            if self.config.outputBands != [] and band not in self.config.outputBands:
```
kfindeisen, line 670: I think the code would be a bit cleaner if you turned `outputBands` into the set of bands you're actually going with, and then used that consistently, instead of constantly referring back to the raw config.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -790,6 +796,15 @@ def _makeArgumentParser(cls):
                                help="data ID, e.g. --id visit=12345 ccd=0")
         return parser
 
+    def runDataRef(self, dataRef):
+        """Override to specify band label to run()."""
+        parq = dataRef.get()
+        funcs = self.getFunctors()
+        band = dataRef.get("calexp_filterLabel", immediate=True).bandLabel
+        df = self.run(parq, funcs=funcs, dataId=dataRef.dataId, band=band)
```
kfindeisen, line 804: `TransformObjectCatalogTask.run` doesn't take a `band` keyword. Did you mean to put this in `TransformCatalogBaseTask`, or did you forget to add the keyword here?

kfindeisen (approved):
Looks much better. 🙂

python/lsst/pipe/tasks/postprocess.py
parejkoj: Yes, the dataframe that is built as part of these tasks is constructed with `'band'` as a column.

python/lsst/pipe/tasks/postprocess.py
parejkoj: If this is used anywhere outside this file (I'd be shocked though), I think it's better that someone see the breakage immediately, because the parquet multi-index column name changed too.

python/lsst/pipe/tasks/postprocess.py
parejkoj: Ugh, and of course `TransformSourceTableTask` has no tests...

Wait, this is fine: it's calling the parent `TransformCatalogBaseTask.run()`, which does take `band`, and this is exactly why.

python/lsst/pipe/tasks/functors.py
parejkoj: Some of these functors also operate on non-coadds. I'm torn on this: I think the places I changed are the necessary ones, but I think not all of this is actually "band" (some of it is just "whatever filter name you want to specify, so long as it's consistent with the input data"). Maybe @yalsayyad has insight?

python/lsst/pipe/tasks/postprocess.py
kfindeisen: Aha, I got confused by all the classes in the file. 😵 

However, it's still a violation of substitutability for `TransformObjectCatalogTask.run` to not take `band`, so please fix that.

python/lsst/pipe/tasks/functors.py
yalsayyad: I'm OK with leaving `filt` as a variable/kwarg name for a band name. The docstring  "Filter upon which to do the calculation" is a little ambiguous.  I don't really see the case where `filt` is specified on something other than a `deepCoadd_obj` working anyway, so it's prob extra work for flexibility that isn't there in the first place.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -607,19 +604,29 @@ class TransformObjectCatalogConfig(TransformCatalogBaseConfig):
...
         doc=("Dictionary mapping full filter name to short one for column name munging."
              "These filters determine the output columns no matter what filters the "
-             "input data actually contain.")
+             "input data actually contain."),
+        deprecated=("Coadds are now identified by the band, so this transform is unused."
+                    "Will be removed after v22.")
+    )
+    outputBands = pexConfig.ListField(
+        dtype=str,
+        default=None,
```
yalsayyad, line 620: Once upon a time you had to set`optional=True` in order to have a config option be None.

yalsayyad (approved):
I only looked at the postprocessing stuff. As long as the transformSourceTable.py and transformObjectCatalog.py that we ran on Monday runs on your ci_hsc, fine with me.

python/lsst/pipe/tasks/postprocess.py
parejkoj: Oh, good point! None of the existing code actually uses this in a way that triggers `Config.validate()`, so it wasn't failing. I've added the `optional` now though.

PR #449, opened 2021-01-06
DM-27896: Add summary statistics component to Exposure

erykoff:


python/lsst/pipe/tasks/computeExposureSummary.py
```diff
@@ -0,0 +1,198 @@
...
+                                lon=observatory.getLongitude().asDegrees()*units.deg,
+                                height=observatory.getElevation()*units.m)
+            obstime = Time(visitInfo.getDate().get(system=DateTime.MJD),
+                           location=loc, format='mjd')
+            coord = SkyCoord(ra*units.degree, decl*units.degree, obstime=obstime, location=loc)
+            with warnings.catch_warnings():
+                warnings.simplefilter('ignore')
+                altaz = coord.transform_to(AltAz)
+
+            zenithDistance = altaz.alt.degree
```
timj, line 157: Given that this is already known to ObservationInfo (and possibly directly calculated from the FITS headers)  it seems a bit wasteful to calculate it here from the ra/dec/time. Can we add zenith distance to visitInfo (like it is already in the butler exposure table)?

timj (approved):
Seems fine.

python/lsst/pipe/tasks/computeExposureSummary.py
erykoff: So the fits header records boresight info, but we want this per-detector (which can vary significantly for the LSST FOV).  So "visitInfo" isn't the right place because it's not per-visit information.

timj: Ah. Can you add a comment to that effect then?

kfindeisen (approved):
No objections to the parts related to `ExposureInfo`.

PR #455, opened 2021-01-27
DM-28323: Make scarlet the default deblender

fred3m:


pipelines/_Multiband.yaml
```diff
@@ -2,7 +2,7 @@ description: Multiband
 tasks:
   detection: lsst.pipe.tasks.multiBand.DetectCoaddSourcesTask
   mergeDetections: lsst.pipe.tasks.mergeDetections.MergeDetectionsTask
-  deblend: lsst.pipe.tasks.deblendCoaddSourcesPipeline.DeblendCoaddSourcesSingleTask
+  deblend: lsst.pipe.tasks.deblendCoaddSourcesPipeline.DeblendCoaddSourcesMultiTask
```
parejkoj, line 5: I would like to see these pipelines be more documented. A comment here to the effect of "Use Scarlet for deblending following RFC-745" or something would be useful. `DeblendCoaddSourcesMultiTask` itself doesn't even have a docstring, so the relationship between this change and scarlet is unclear unless one already knows that "DeblendMulti" means "scarlet".

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -353,7 +353,7 @@ class DeblendCoaddSourcesConfig(Config):
                                           doc="Deblend sources separately in each band")
     multiBandDeblend = ConfigurableField(target=ScarletDeblendTask,
                                          doc="Deblend sources simultaneously across bands")
-    simultaneous = Field(dtype=bool, default=False, doc="Simultaneously deblend all bands?")
+    simultaneous = Field(dtype=bool, default=True, doc="Simultaneously deblend all bands?")
```
parejkoj, line 356: The doc for this field needs to say that this used to switch between the `singleBandDeblend` and `multiBandDeblend` subtasks listed above.

parejkoj (commented):
Two comments about improved documentation.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -353,7 +353,11 @@ class DeblendCoaddSourcesConfig(Config):
                                           doc="Deblend sources separately in each band")
     multiBandDeblend = ConfigurableField(target=ScarletDeblendTask,
                                          doc="Deblend sources simultaneously across bands")
-    simultaneous = Field(dtype=bool, default=False, doc="Simultaneously deblend all bands?")
+    simultaneous = Field(dtype=bool,
+                         default=True,
+                         doc="Simultaneously deblend all bands? "
+                             "True uses the 'ScarletDeblendTask' while False performs deblending in "
```
parejkoj, line 359: I think this docstring should just refer to the ConfigurableField names, not what they are default-configured to. You also need `()` around this docstring, because I'm pretty sure `doc=` is only being assigned to the first line here.

timj: I don't think you need parens around string concatenation. Looks fine to me.

timj:
```
milo:~ $ python xxx.py 
abcdef
milo:~ $ cat xxx.py
def p(x):
    print(x)

p("abc"
  "def")
```

parejkoj: Thanks, I never quite figured out when it was and wasn't necessary.

PR #459, opened 2021-02-02
DM-28609: Fix Zero Size Bounding Box Crash in InsertFakes.py

leeskelvin:


morriscb (approved):
This feels worth approving based on the reduction of lines of code alone.

I'll drop a comment on the ticket regarding the difference in the catalogs.

PR #461, opened 2021-02-02
DM-28570: Take both width and height for cutouts

SimonKrughoff:


python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -104,10 +104,11 @@ def run(self, in_table, calexp):
             pt = geom.SpherePoint(geom.Angle(ra, geom.degrees),
                                   geom.Angle(dec, geom.degrees))
             pix = wcs.skyToPixel(pt)
-            size = rec['size'].value
+            width = rec['xspan'].value
+            height = rec['yspan'].value
```
r-owen, line 108: Please consider using `xspan` and `yspan` for variable names, in order to make it more obvious where they come from.

I can see arguments both ways. If you prefer what you have then by all means keep it.

python/lsst/pipe/tasks/calexpCutout.py
```diff
@@ -45,7 +45,7 @@ class CalexpCutoutTaskConfig(pipeBase.PipelineTaskConfig,
 
 class CalexpCutoutTask(pipeBase.PipelineTask):
     """Task for computing cutouts on a specific calexp given
-    positions and sizes of the stamps.
+    positions, widths, and heights of the stamps.
```
r-owen, line 48: I strongly suggest adding the words xspan and yspan here, to make it clearer which table columns are needed. For example:

given position, xspan (width), and yspan (height).

r-owen (approved):
This looks like a useful update. I have two minor suggestions for your consideration.

PR #464, opened 2021-02-03
DM-14875: Add README to pipe_tasks

MorganSchmitz:
[Jira ticket](https://jira.lsstcorp.org/browse/DM-14875).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,0 +1,16 @@
+##########
+pipe_tasks
+##########
+``pipe_tasks`` is a package in the `LSST Science Pipelines <https://pipelines.lsst.io>`_.
+
+``pipe_tasks`` provides many of the `Task <https://pipelines.lsst.io/py-api/lsst.pipe.base.Task.html#lsst.pipe.base.Task>`_ classes that drive the LSST Science Pipelines. The `command-line tasks listed here <https://pipelines.lsst.io/modules/lsst.pipe.tasks/index.html#lsst-pipe-tasks-command-line-tasks>`_ are useful data processing entry points for most users. These scripts can be found in the ``bin.src/`` folder. After building and installing the package, they can also be found in the created ``bin/`` folder.
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -0,0 +1,16 @@
+##########
+pipe_tasks
+##########
+``pipe_tasks`` is a package in the `LSST Science Pipelines <https://pipelines.lsst.io>`_.
+
+``pipe_tasks`` provides many of the `Task <https://pipelines.lsst.io/py-api/lsst.pipe.base.Task.html#lsst.pipe.base.Task>`_ classes that drive the LSST Science Pipelines. The `command-line tasks listed here <https://pipelines.lsst.io/modules/lsst.pipe.tasks/index.html#lsst-pipe-tasks-command-line-tasks>`_ are useful data processing entry points for most users. These scripts can be found in the ``bin.src/`` folder.
```
parejkoj, line 6: I think we should mention why they are in `bin.src/` instead of `bin/` (i.e., when built and installed they are in `bin/`).

parejkoj (commented):
Thank you for adding this.

I was surprised to see it as an RST file; I'd thought all our readmes were markdown, but it looks like we do have a handful of `README.rst`s. Not sure it really matters, but markdown is a little less clunky for simple things like this.

MorganSchmitz:
We wondered about .rst vs .md. I think in the end we went with rst because that's what's used in the [examples of the auto-generated template](https://github.com/lsst/templates/tree/884f3705eea9ea9d2957a6e6a775c627ed995d6e/project_templates/stack_package). Though looking at it again today, I see the README for that page is a md...

I certainly can change if you think it's best.

parejkoj:
I've asked for guidance on the rst vs. md question on #dm-docs.

parejkoj:
According to @jonathansick, we are slowly moving to rst because that's what the stack template generates, so stick with .rst here.

MorganSchmitz:
Thanks, John!
Okay to merge then?

PR #468, opened 2021-02-10
DM-26372: Units wrong in deepDiff_diaSrc catalog

mrawls:
This PR adds units of "count" to instFlux and instFluxErr. Previously, they errantly had no units.


kherner (approved):
Looks fine.

PR #469, opened 2021-02-18
DM-28863: Create VisitSummary tables for testdata_jointcal

parejkoj:


parejkoj:
Wrong branch name. Will delete branch.

PR #472, opened 2021-02-23
DM-28759: Support driver to write new parquet tables from old src tables

yalsayyad:
We have two specialized use cases for this:
1) Ingesting `src` catalogs for DC2
2) The current HSC data-release is going to be an incremental data release. We will not rerun singleFrameDriver for ccds run in the 2020 data release.


timj:
It looks to me like this work was redone in gen3 on DM-34466 (#675) and so we can close this PR.

PR #473, opened 2021-02-25
DM-20695: Convert SelectImageTasks to Gen3

yalsayyad:
* Refactor SelectImageTasks, adding a run method that takes lists
  of calexp properties. runDataRef remains the Gen2 entry point
  which still takes dataRefs.
* MakeWarp now calls the configured selectImages subtask in
  runQuantum. The calexp list was changed to deferLoad, because the padding
  needed in the gen3 skymap is providing many more calexps than actually
  overlap the patch. Loading is performed after image selection is run,
  so that only the calexps that will be used are loaded into memory.

### DIFF
--- a/python/lsst/pipe/tasks/makeCoaddTempExp.py
+++ b/python/lsst/pipe/tasks/makeCoaddTempExp.py
@@ -29,6 +29,7 @@ import lsst.pipe.base as pipeBase
 import lsst.pipe.base.connectionTypes as cT
 import lsst.log as log
 import lsst.utils as utils
+import lsst.geom
 from lsst.meas.algorithms import CoaddPsf, CoaddPsfConfig
 from lsst.skymap import BaseSkyMap
 from .coaddBase import CoaddBaseTask, makeSkyInfo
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -672,25 +709,48 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+                           dataIdList=[dataIdList[i] for i in goodIndices],
                            skyInfo=skyInfo)
         if self.config.makeDirect:
             butlerQC.put(results.exposures["direct"], outputRefs.direct)
         if self.config.makePsfMatched:
             butlerQC.put(results.exposures["psfMatched"], outputRefs.psfMatched)
 
-    def prepareCalibratedExposures(self, calExpList, backgroundList=None, skyCorrList=None):
+    def filterInputs(self, indices, inputs):
+        """Return task inputs with their lifsts filtered by indices
```
ebellm, line 742:
```suggestion
        """Return task inputs with their lists filtered by indices
```

ebellm (approved):

PR #474, opened 2021-03-01
DM-28429: Convert MultiProFitTask into gen3 PipelineTask

taranu:


python/lsst/pipe/tasks/fit_multiband.py
```diff
@@ -0,0 +1,314 @@
...
+                            bands_extra = bands_data.difference(bands_needed)
+                elif bands_set != bands_data:
+                    raise ValueError(
+                        f'Datarefs={ref_d} have data with bands in the set={bands_set}'
+                        f' which differs from the previous={bands_data}); bandsets must be identical.'
+                    )
+                if bands_extra:
+                    for dref in ref_d:
+                        if dref.dataId['band'] in bands_extra:
+                            ref_d.remove(dref)
```
taranu, line 161: I added logic here to drop the exposure dataref from bands only in bands_read_only, but then I remembered that I actually need the photoCalib to calibrate the catalog for magnitude-based priors.

Would it be possible to do that and add a photoCalib dataref that isn't specified in the connections, or is that beyond what adjustQuantum is meant to do? (I'm probably not going to do it either way as it over-complicates things for a marginal benefit - the exposure object doesn't actually read fits images into memory until you first retrieve them, right?)

python/lsst/pipe/tasks/fit_multiband.py
```diff
@@ -0,0 +1,314 @@
...
+        return datasetRefMap
+
+
+class MultibandFitSubConfig(pexConfig.Config):
+    def bands_read_only(self) -> Set:
```
taranu, line 166: I couldn't make this an ABC, but I wanted a default implementation anyway. PyCharm complains that this could be a staticmethod - adding that wouldn't affect subclasses' overrides, would it?

python/lsst/pipe/tasks/fit_multiband.py
```diff
@@ -0,0 +1,314 @@
...
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        input_refs_objs = [(inputRefs.cats_meas, inputs['cats_meas']), (inputRefs.coadds, inputs['coadds'])]
+        cats, exps = [
+            {dRef.dataId: obj for dRef, obj in zip(refs, objs)}
+            for refs, objs in input_refs_objs
+        ]
+        dataIds = set(cats).union(set(exps))
+        catexps = [
+            CatalogExposure(catalog=cats.get(dataId), exposure=exps.get(dataId), dataId=dataId)
```
taranu, line 281: Now that CatalogExposures' catalogs and exposures are optional (I can see future use cases for just wanting one or the other attached to a dataId), I went with defensively putting inputs into dicts as you suggested and in case future changes to valid subtasks decide to drop one or the other. Does that seem reasonable?

PR #475, opened 2021-03-01
DM-27613: Add gen3 ability to add external SkyWcs and PhotoCalib

erykoff:


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -703,12 +749,63 @@ def prepareCalibratedExposures(self, calExpList, backgroundList=None, skyCorrLis
...
         for calexp, background, skyCorr in zip(calExpList, backgroundList, skyCorrList):
             mi = calexp.maskedImage
             if not self.config.bgSubtracted:
                 mi += background.getImage()
+
+            if externalSkyWcsCatalog is not None or externalPhotoCalibCatalog is not None:
+                detectorId = calexp.getInfo().getDetector().getId()
+
+            # Find and apply external photoCalib
+            if externalPhotoCalibCatalog is not None:
```
parejkoj, line 777: Could you break out the external calibrations stuff into some short methods? e.g. `_getExternalPhotoCalib()` that returns the correct thing? I think that would tidy this up a bunch.

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -703,12 +749,63 @@ def prepareCalibratedExposures(self, calExpList, backgroundList=None, skyCorrLis
             Sequence of backgrounds to be added back in if bgSubtracted=False
         skyCorrList : `list` of `lsst.afw.math.backgroundList`
             Sequence of background corrections to be subtracted if doApplySkyCorr=True
+        externalSkyWcsCatalog : `lsst.afw.table.ExposureCatalog`
+            Exposure catalog with external skyWcs to be applied
+            if config.doApplyExternalSkyWcs=True.
+        externalPhotoCalibCatalog : `lsst.afw.table.ExposureCatalog`
```
parejkoj, line 755: These two are optional, right?

python/lsst/pipe/tasks/makeCoaddTempExp.py
fred3m: Is that necessary? I don't see the need to create a separate function unless there is reason to believe that it will be used somewhere else, which doesn't seem likely in this case.

fred3m (approved):
This looks fine to me (after it is rebased), just make sure to resolve the questions raised by @parejkoj .

python/lsst/pipe/tasks/coaddBase.py
```diff
@@ -82,9 +74,16 @@ class CoaddBaseConfig(pexConfig.Config):
...
+        dtype=bool,
+        default=True,
+        doc=("When using doApplyExternalPhotoCalib, use global calibrations. "
+             "When False, per-tract calibrations will be used instead.")
+    )
     externalPhotoCalibName = pexConfig.ChoiceField(
         dtype=str,
-        doc="Type of external PhotoCalib if `doApplyExternalPhotoCalib` is True.",
+        doc=("Type of external PhotoCalib if `doApplyExternalPhotoCalib` is True. "
+             "This field is only used for Gen2 middleware."),
```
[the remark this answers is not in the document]
timj: DM-20572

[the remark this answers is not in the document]
erykoff: Perfect, I'll note this with a TODO.

python/lsst/pipe/tasks/makeCoaddTempExp.py
erykoff: I'm not going to split this into separate mini-functions, but the routine has been made simpler by the use of `catalog.find(detectorId)` with the new format.

PR #477, opened 2021-03-03
DM-29017: Update visitSummary catalogs to use detector id as index.

erykoff:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1041,7 +1043,7 @@ def _combineExposureMetadata(self, visit, dataRefs, isGen3=True):
 
             rec['physical_filter'] = filterLabel.physicalLabel if filterLabel.hasPhysicalLabel() else ""
             rec['band'] = filterLabel.bandLabel if filterLabel.hasBandLabel() else ""
-            rec['detector_id'] = detector.getId()
+            rec['id'] = detector.getId()
```
parejkoj, line 1046: I don't know that it makes any real difference, but you can also do `rec.setId(detector.getId())` here, since every Catalog has `id`.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1058,6 +1060,12 @@ def _combineExposureMetadata(self, visit, dataRefs, isGen3=True):
             rec['raCorners'][:] = [sph.getRa().asDegrees() for sph in sph_pts]
             rec['decCorners'][:] = [sph.getDec().asDegrees() for sph in sph_pts]
 
+        metadata = dafBase.PropertyList()
+        metadata.add("COMMENT", "Catalog id is detector id, sorted.")
+        metadata.add("COMMENT", "Only detectors with data have entries.")
```
parejkoj, line 1065: I guess this latter comment is true in this case because we're iterating over dataRefs. It wasn't entirely obvious to me at first.

parejkoj (commented):
Two comments that may not require any actual changes.

PR #479, opened 2021-03-05
DM-29065: Sort all input lists by calexpList detector order in makeWarp

yalsayyad:


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -875,3 +881,37 @@ def prepareCalibratedExposures(self, calExpList, backgroundList=None, skyCorrLis
...
+        Sorted inputRefs.
+    """
+    for key, refs in inputRefs:
+        if isinstance(refs, Iterable):
+            if hasattr(refs[0], "dataId"):
+                inputKeyOrder = [ref.dataId[dataIdKey] for ref in refs]
+            else:
+                inputKeyOrder = [ref.datasetRef.dataId[dataIdKey] for ref in refs]
+            if inputKeyOrder != outputKeyOrder:
+                setattr(inputRefs, key, reorderAndPadList(refs, inputKeyOrder, outputKeyOrder))
```
erykoff, line 916: Can this be `inputRefs[key] = reorderAndPadList(refs, inputKeyOrder, outputKeyOrder)`?

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -875,3 +881,37 @@ def prepareCalibratedExposures(self, calExpList, backgroundList=None, skyCorrLis
...
+
+    Returns:
+    --------
+    inputRefs: `lsst.pipe.base.connections.InputQuantizedConnection`
+        Sorted inputRefs.
+    """
+    for key, refs in inputRefs:
+        if isinstance(refs, Iterable):
+            if hasattr(refs[0], "dataId"):
+                inputKeyOrder = [ref.dataId[dataIdKey] for ref in refs]
```
erykoff, line 912: This also applies to the function above which has just been moved and therefore might be outside the scope of this ticket.  But the term `key` is used in two different ways here.  One is a dict-like `key` (as in `for key, refs in inputRefs`) which is very pythonic.  The other is a sorting `key` which I guess python `sort` functions also use the term `key` in this way, but in this particular method I find the overloading of the term confusing.
It may be verbose, but maybe call `inputKeyOrder` -> `inputSortKeyOrder`?

erykoff (approved):
This is fine, but I did get a bit confused about the overloaded term `key` in here.

python/lsst/pipe/tasks/makeCoaddTempExp.py
yalsayyad: Nope.
```(Pdb) inputRefs[key] 
*** TypeError: 'InputQuantizedConnection' object is not subscriptable
```

python/lsst/pipe/tasks/makeCoaddTempExp.py
yalsayyad: The plain `key` -->  `connectionName`,
`inputKeyOrder` --> `inputSortKeyOrder`, 
`outputKeyOrder` --> `outputSortKeyOrder`

PR #481, opened 2021-03-10
DM-28597: Fix colorterm/photoCal filterLabel confusion

parejkoj:


python/lsst/pipe/tasks/colorterms.py
```diff
@@ -58,15 +59,16 @@ class Colorterm(Config):
     c1 = Field(dtype=float, default=0.0, doc="First-order parameter")
     c2 = Field(dtype=float, default=0.0, doc="Second-order parameter")
 
-    def getCorrectedMagnitudes(self, refCat, filterName):
+    def getCorrectedMagnitudes(self, refCat, filterName="deprecatedArgument"):
```
erykoff, line 62: Am I crazy, or was `filterName` not used in this method at all previously?  How strange...

erykoff (approved):

python/lsst/pipe/tasks/colorterms.py
parejkoj: I think it was used some time ago and got refactored out at some point. I don't feel like digging into the history to check (but I'll bet good money I was responsible for said refactoring).

PR #482, opened 2021-03-11
DM-28942: Create dipMeanFlux/dipFluxDiff functor.

morriscb:


tests/test_functors.py
```diff
@@ -533,6 +534,71 @@ def _testLocalPhotometryFunctors(self, functor, parq, testValues):
...
 
+    def testDipPhotometry(self):
+        """Test calibrated flux calculations for dipoles."""
+        fluxNeg = -100
+        fluxPos = 101
+        fluxErr = 10
+        calib = 10
+        calibErr = 1
+
+        absMean = 0.5 * (fluxPos - fluxNeg) * calib
```
sr525, line 547: See style guide about spaces around operators: https://developer.lsst.io/python/style.html#binary-operators-should-be-surrounded-by-a-single-space-except-for

PR #485, opened 2021-03-15
DM-29225: Add instrument to quantum dimensions

MorganSchmitz:


arunkannawadi (commented):
Looks good. It might be better to have them in the same order as you have them for `skyCorr` for readability, although I don't suppose the ordering matters.

PR #489, opened 2021-03-26
tickets/DM-29031 Add new Config and Field types

natelust:
Add new Config and Field types. These config types known as
ConfigurableActions, are callables that can have state specified
at configuration time. Two container Fields,
ConfigurableActionField and ConfigurableActionStructField are
introduced to allow adding ConfigurableActions to Config classes.

These classes are meant to serve as the basis for future work on
specific actions to interact with DataFrames.

### DIFF
--- a/python/lsst/pipe/tasks/configurableActions/__init__.py
+++ b/python/lsst/pipe/tasks/configurableActions/__init__.py
@@ -1,0 +1,23 @@
+# This file is part of pipe_tasks.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from ._configurableAction import *
+from ._configurableActionStructField import *
+from ._configurableActionField import *
--- a/python/lsst/pipe/tasks/configurableActions/_configurableAction.py
+++ b/python/lsst/pipe/tasks/configurableActions/_configurableAction.py
@@ -1,0 +1,49 @@
+# This file is part of pipe_tasks.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ["ConfigurableAction"]
+
+from typing import Any
+
+from lsst.pex.config.config import Config
+
+
+class ConfigurableAction(Config):
+    """A `ConfigurableAction` is an interface that extends a
+    `lsst.pex.config.Config` class to include a `__call__` method.
+
+    This interface is designed to create an action that can be used at
+    runtime with state that is determined during the configuration stage. A
+    single action thus may be assigned multiple times, each with different
+    configurations.
+
+    This allows state to be set and recorded at configuration time,
+    making future reproduction of results easy.
+
+    This class is intended to be an interface only, but because of various
+    inheritance conflicts this class can not be implemented as an Abstract
+    Base Class. Instead, the `__call__` method is purely virtual, meaning that
+    it will raise a `NotImplementedError` when called. Subclasses that
+    represent concrete actions must provide an override.
+    """
+    def __call__(self, *args, **kwargs) -> Any:
+        raise NotImplementedError("This method should be overloaded in subclasses")
--- a/python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
+++ b/python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
@@ -1,0 +1,331 @@
+# This file is part of pipe_tasks.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("ConfigurableActionStructField", "ConfigurableActionStruct")
+
+from typing import Iterable, Mapping, Optional, TypeVar, Union, Type, Tuple, List, Any, Dict
+
+from lsst.pex.config.config import Config, Field, FieldValidationError, _typeStr, _joinNamePath
+from lsst.pex.config.comparison import compareConfigs, compareScalars, getComparisonName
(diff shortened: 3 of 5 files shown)


python/lsst/pipe/tasks/configurableActions/__init__.py
```diff
@@ -0,0 +1,3 @@
+from ._configurableAction import *
```
TallJimbo, line 1: Should add a license preamble here, too.

python/lsst/pipe/tasks/configurableActions/_configurableAction.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import Any
+
+from lsst.pex.config.config import Config
+
+
+class ConfigurableAction(Config):
+    """A `ConfigurableAction` is an interface that extends a
+    `lsst.pex.config.Config` class to include a `__call__` method.
+
+    This interfaced is designed to create an action that can be used at
```
TallJimbo, line 34:
```suggestion
    This interface is designed to create an action that can be used at
```

python/lsst/pipe/tasks/configurableActions/_configurableAction.py
```diff
@@ -0,0 +1,49 @@
...
+    runtime with state that is determined during the configuration stage. A
+    single action thus may be assigned multiple times, each with different
+    configurations.
+
+    In addition to allowing state to be set at configuration time these
+    actions allow the state of execution to be recorded alongside all the
+    other program configuration settings, allowing easy reproduction of
+    results at some future time.
```
TallJimbo, line 42: I find "state of execution" here a bit confusing, though I know what you mean from our earlier discussion; how about,
```
This allows state to be set and recorded at configuration time, making future reproduction of results easy.
```

python/lsst/pipe/tasks/configurableActions/_configurableAction.py
```diff
@@ -0,0 +1,49 @@
...
+
+    This class is intended to be an interface only, and the `__call__` method
+    is purely virtual. Subclasses that represent concrete actions must
+    provide an override.
```
TallJimbo, line 46: I know this can't be a real `ABC` that uses `abstractmethod` because it'd be a metaclass conflict, but maybe we should add a note to that effect?

python/lsst/pipe/tasks/configurableActions/__init__.py
TallJimbo: Might even be a rare case where a module docstring actually makes sense, too, just to serve as sort of README introduction to the subpackage.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -0,0 +1,330 @@
...
+    `ConfigurableActionStruct` on which the attribute is invoked such that it
+    will be updated to contain the entries specified by the structure on the
+    right hand side of the equals sign.
+
+    The second convenience attribute is named remove. You may assign an
+    iterable of strings which correspond to attribute names on the
+    `ConfigurableActionStruct`. All of the corresponding attributes will then
+    be removed. If any attribute does not exist, an `AttributeError` will be
+    raised. Any attributes in the Iterable prior to the name which raises will
+    have been removed from the `ConfigurableActionStruct`
```
TallJimbo, line 115: Why special decorators instead of just methods?

I think e.g.
```
config.my_action.update(dictionary)
```
reads a lot more naturally than
```
config.my_action.update = dictionary
```
Is this just for command-line/pipeline overrides again?  If so, they're giving me some reservations about whether catering to their needs is a good idea.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -0,0 +1,330 @@
...
+        if name in self._attrs:
+            del self._attrs[name]
+        else:
+            super().__delattr__(name)
+
+    def __iter__(self) -> Generator[ConfigurableAction, None, None]:
+        yield from self._attrs.values()
+
+    def items(self) -> Iterable[Tuple[str, ConfigurableAction]]:
+        yield from self._attrs.items()
```
TallJimbo, line 184: I don't think this should be a generator.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -0,0 +1,330 @@
...
+            super().__getattribute__(attr)
+
+    def __delattr__(self, name):
+        if name in self._attrs:
+            del self._attrs[name]
+        else:
+            super().__delattr__(name)
+
+    def __iter__(self) -> Generator[ConfigurableAction, None, None]:
+        yield from self._attrs.values()
```
TallJimbo, line 181: Most curiousity: why the extra generator indirection, rather than just `return iter(self._attrs.values())`?

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class ConfigurableActionStructField(Field):
+    """`ConfigurableActionStructField` is a `~lsst.pex.config.Field` subclass
+    that allows `ConfigurableAction` s to be organized in a
+    `~lsst.pex.config.Config` class in a manor similar to how a
+    `~lsst.pipe.base.Struct` works.
+
+    This class implements a `ConfigurableActionStruct` as an intermediary
+    object to organize the `ConfigurableActions`. See it's documentation for
```
TallJimbo, line 197:
```suggestion
    object to organize the `ConfigurableAction`\ s. See its documentation for
```

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -0,0 +1,330 @@
...
+            history = instance._history.setdefault(self.name, [])
+            history.append((value, at, label))
+
+        if not isinstance(value, ConfigurableActionStruct):
+            raise FieldValidationError(self, instance,
+                                       "Can only assign things that are subclasses of Configurable Action")
+        instance._storage[self.name] = value
+
+    def __get__(self: T, instance, owner=None, at=None, label="default"
+                ) -> Union[None, T, ConfigurableActionStruct]:
```
TallJimbo, line 239: I'm a bit surprised you've found a tool that's happy with the level of type annotations here (and useful with just that).  I thought they all complained when annotated some arguments/results but not others even within the same method.

tests/test_configurableActions.py
```diff
@@ -0,0 +1,226 @@
...
+
+class ConfigurableActionsTestCase(unittest.TestCase):
+    def _createConfig(self, default=None, singleDefault=None):
+        class TestConfig(Config):
+            actions = ConfigurableActionStructField(doc="Actions to be tested", default=default)
+            singleAction = ConfigurableActionField(doc="A configurable action", default=singleDefault)
+        return TestConfig
+
+    def testConfigInstatiation(self):
+        # This will raise if there is an issue instatiating something
```
TallJimbo, line 70:
```suggestion
        # This will raise if there is an issue instantiating something
```

tests/test_configurableActions.py
```diff
@@ -0,0 +1,226 @@
...
+        config = configClass()
+        self.assertEqual(config.toDict(), {'actions': {'test1': {'var': 0}}, 'singleAction': {'var': 0}})
+
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
+
+class MatchMemoryTestCase(lsst.utils.tests.MemoryTestCase):
+    pass
```
TallJimbo, line 221: Name of this class looks like a copy-paste relic.  I'm also not sure if it's useful, or just old boilerplate.

TallJimbo (approved):
Looks good.  Comments are all detail-oriented.  It's possible we already had the debate about whether `update` and `remove` should be regular methods or special decorators, and I'm just going to have to be sad about that, and that's the only issue that's remotely structural.

python/lsst/pipe/tasks/configurableActions/__init__.py
natelust: I agree with module docstring, but I talked with yusra and to not block the people who are looking to start using this, we were going to get it merged, as that need not be a blocker.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
natelust: I had this in initially but flake8 on github rejected this, saying it was invalid, and the test failed so I removed it. Maybe the repo is miss configured somehow.

timj: I think you need to declare the docstring as a raw-string for it to be allowed.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
timj: are we running mypy now in pipe_tasks?

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
natelust: There is no compelling reason that either of these two methods do this. Its left over from some random early on idea that I moved on from and just never thought to change here. One note I will make on this, there is a subtle usage difference when it comes to `return iter` `yield from` `yield` when it come to what information is available in traceback messages, but in this case it really does not matter.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
natelust: No, I am using the language server support in VS code that gives me something like the feedback from just running mypy as I go.

natelust: @TallJimbo yeah, pylance treats it as an "unknown" type. It does restrict some of the things it can do with downstream analysis, but it does not stop it from running on what it still can. That said I had intended to annotate what I could.

python/lsst/pipe/tasks/configurableActions/__init__.py
natelust: I will create a ticket for that so it does not get forgotten.

PR #490, opened 2021-03-31
DM-22277: Gen3 object tables

yalsayyad:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -98,6 +133,26 @@ def runDataRef(self, patchRefList):
...
+                    zip(inputRefs.inputCatalogMeas, inputs['inputCatalogMeas'])}
+        forcedSourceDict = {ref.dataId['band']: {'forced_src': cat} for ref, cat in
+                            zip(inputRefs.inputCatalogForcedSrc, inputs['inputCatalogForcedSrc'])}
+
+        catalogs = {}
+        for band, catDict in measDict.items():
+            catalogs[band] = {}
+            catalogs[band]['meas'] = measDict[band]['meas']
+            catalogs[band]['forced_src'] = forcedSourceDict[band]['forced_src']
+            catalogs[band]['ref'] = inputs['inputCatalogRef']
```
erykoff, line 149: Would this read a bit better with: 
```
catalogs[band] = {'meas': measDict[band]['meas'],
                  'forced_src': forcedSourceDict[band]['forced_src'],
                  'ref': inputs['inputCatalogRef']}
```

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -98,6 +133,26 @@ def runDataRef(self, patchRefList):
...
+
+        dataId = butlerQC.quantum.dataId
+        parq = self.run(catalogs=catalogs, tract=dataId['tract'], patch=dataId['patch'])
+        outputs = pipeBase.Struct(outputCatalog=parq.toDataFrame())
```
erykoff, line 153: I don't know if there's a big overhead to converting to and from parquet and dataFrames, but I think it's odd that the last thing the `run` method does is convert the dataframe to parquet, and the first thing this code does is convert it back to a dataframe.  I know that's what the gen2 code expects, but in terms of planning for gen2 removal, I think that `run` should return a dataframe and `runDataRef` could then parquet-ify the table?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -55,7 +56,41 @@ def flattenFilters(df, noDupCols=['coord_ra', 'coord_dec'], camelCase=False):
...
-class WriteObjectTableConfig(pexConfig.Config):
+class WriteObjectTableConnections(pipeBase.PipelineTaskConnections,
+                                  defaultTemplates={"coaddName": "deep"},
+                                  dimensions=("tract", "patch", "skymap")):
+    inputCatalogMeas = connectionTypes.Input(
+        doc="Catalog of source measurements on the deepCoadd.",
+        dimensions=("tract", "patch", "band", "skymap"),
+        storageClass="SourceCatalog",
+        name="{coaddName}Coadd_meas",
+        multiple=True
```
erykoff, line 67: I guess there's no advantage to these being `deferLoad=True` because we need to hold them all in memory at the same time anyway?

erykoff (approved):
Some minor comments below, otherwise looks good (assuming it works with ci_hsc_gen3).

python/lsst/pipe/tasks/postprocess.py
yalsayyad: That's right. We're vertically concatenating them.

python/lsst/pipe/tasks/postprocess.py
yalsayyad: I agree.  ParquetTable will disappear with along with Gen2.  
Was avoiding API changes on this one, but if the reviewer advocates for API changes, well then I guess I have to :)

PR #491, opened 2021-03-31
DM-25305: Add stacking of saved bright star stamps to build extended PSF model

MorganSchmitz:
[Jira ticket](https://jira.lsstcorp.org/browse/DM-25305); Jenkins run [#33904](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/33904/pipeline).

~Main one of two PRs, the other being in daf_butler~
(edit: the daf_butler PR has now been moved to DM-29583)

### DIFF
--- a/python/lsst/pipe/tasks/extended_psf.py
+++ b/python/lsst/pipe/tasks/extended_psf.py
@@ -28,7 +28,12 @@ from typing import List
 
 from lsst.afw import image as afwImage
 from lsst.afw import fits as afwFits
+from lsst.afw import math as afwMath
 from lsst.daf.base import PropertyList
+from lsst.pipe import base as pipeBase
+from lsst.pipe.tasks.assembleCoadd import AssembleCoaddTask
+import lsst.pex.config as pexConfig
+from lsst.geom import Extent2I
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -239,3 +245,250 @@ def readFits(cls, filename):
...
+        }
+    )
+    num_sigma_clip = pexConfig.Field(
+        dtype=float,
+        doc="Sigma for outlier rejection; ignored if stacking_statistic != 'MEANCLIP'.",
+        default=4
+    )
+    num_iter = pexConfig.Field(
+        dtype=int,
+        doc="Number of iterations of outlier rejection; ignored if atackingStatistic != 'MEANCLIP'.",
```
leeskelvin, line 276: typo: `stackingStatistic`

python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -239,3 +245,250 @@ def readFits(cls, filename):
...
+        pipeBase.CmdLineTask.__init__(self, *args, **kwargs)
+
+    def _set_up_stacking(self, example_stamp):
+        """Configure stacking statistic and control from config fields.
+        """
+        stats_control = afwMath.StatisticsControl()
+        stats_control.setNumSigmaClip(self.config.num_sigma_clip)
+        stats_control.setNumIter(self.config.num_iter)
+        bad_masks = self.config.bad_mask_planes
+        if bad_masks:
```
leeskelvin, line 313: Perhaps personal preference, but this could be combined into:
`if bad_masks := self.config.bad_mask_planes:`

python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -239,3 +245,250 @@ def readFits(cls, filename):
...
+        ----------
+        bss_ref_list : `list` of
+                `lsst.daf.butler._deferredDatasetHandle.DeferredDatasetHandle`
+            List of available bright star stamps data references.
+        region_name : `str`, optional
+            Name of the focal plane region, if applicable. Only used for
+            logging purposes, when running over multiple such regions
+            (typically from `MeasureExtendedPsfTask`)
+        """
+        log_message = f'Building extended PSF from stamps extracted from {len(bss_ref_list)} detector images '
```
leeskelvin, line 337: Does there need to be an extra space at the end here?

python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -239,3 +245,250 @@ def readFits(cls, filename):
...
+        if region_name:
+            log_message += f'for region "{region_name}".'
```
leeskelvin, line 339: Maybe it would be cleaner to move the extra space at the end of L337 to the beginning of the f-string in L339.

python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -239,3 +245,250 @@ def readFits(cls, filename):
...
+        # read in example set of full stamps
+        for bss_ref in bss_ref_list:
+            if not isinstance(bss_ref, DeferredDatasetHandle):
+                if not bss_ref.datasetExists("brightStarStamps"):
+                    self.log.warn("Could not find %s %s; skipping it", "brightStarStamps", bss_ref.dataId)
+                    bss_ref_list.remove(bss_ref)
+                    continue
+            bss = bss_ref.get(datasetType="brightStarStamps", immediate=True)
+            break
+        example_stamp = bss[0].stamp_im
```
leeskelvin, line 350: From reading this block, it seems like the purpose is to generate `example_stamp`. I was a little confused as to why two `if` statements were required, however, after having checked with you offline, it appears that it may be possible to inject a valid `DeferredDatasetHandle` that does not contain any `brightStarStamps` dataset types. In such a case however, L350 would still raise an exception and crash.

It might be cleaner to replace the two `if` statements with a try/except block here, i.e.:
```
for bss_ref in bss_ref_list:
    try:
        bss = bss_ref.get(datasetType="brightStarStamps", immediate=True)
        example_stamp = bss[0].stamp_im
    except ...
```

python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -239,3 +245,250 @@ def readFits(cls, filename):
...
+        # perform stacking
+        for jbbox, bbox in enumerate(sub_bboxes):
+            all_stars = None
+            for bss_ref in bss_ref_list:
+                if not isinstance(bss_ref, DeferredDatasetHandle):
+                    if not bss_ref.datasetExists("brightStarStamps"):
+                        self.log.warn("Could not find %s %s; skipping it", "brightStarStamps", bss_ref.dataId)
+                        bss_ref_list.remove(bss_ref)
+                        continue
+                read_stars = bss_ref.get(datasetType="brightStarStamps", parameters={'bbox': bbox})
```
leeskelvin, line 372: As this is the second time you're checking each entry in `bss_ref_list` to check for `bss_ref`s that are valid `DeferredDatasetHandle` types and that also contain `brightStarStamps` dataset types, perhaps this functionality should be stripped out and placed somewhere higher up in the run task (before L342) to check and remove invalid entries at the outset. That way, you can raise an exception and quit early if no valid entries are found, and reduce the complexity of using `bss_ref_list` elsewhere in the code.

python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -239,3 +245,250 @@ def readFits(cls, filename):
...
+        stats_control, stats_flags = self._set_up_stacking(example_stamp)
+        # perform stacking
+        for jbbox, bbox in enumerate(sub_bboxes):
+            all_stars = None
+            for bss_ref in bss_ref_list:
+                if not isinstance(bss_ref, DeferredDatasetHandle):
+                    if not bss_ref.datasetExists("brightStarStamps"):
+                        self.log.warn("Could not find %s %s; skipping it", "brightStarStamps", bss_ref.dataId)
+                        bss_ref_list.remove(bss_ref)
+                        continue
```
leeskelvin, line 371: As this is the second time you're checking each entry in `bss_ref_list` to check for `bss_ref`s that are valid `DeferredDatasetHandle` types and that also contain `brightStarStamps` dataset types, perhaps this functionality should be stripped out and placed somewhere higher up in the run task (before L342) to check and remove invalid entries at the outset. That way, you can raise an exception and quit early if no valid entries are found, and reduce the complexity of using `bss_ref_list` elsewhere in the code.

python/lsst/pipe/tasks/extended_psf.py
leeskelvin, line 372: As above, this might want to be wrapped in a try/except block. (However, if the `bss_ref_list` checking is moved to its own initial section, then you might be safe leaving this line as-is, due to the fact that you'd then be able to guarantee this dataset type definitely already exists in `bss_ref_list`.

python/lsst/pipe/tasks/extended_psf.py
MorganSchmitz: Mystery solved!

This is just vestigial code left from when I had a gen2 version of this Task. The first `if` clause was only ever `True` when the inputs to `run` were those given by the gen2 `runDataRef`. 
I should have removed all this (and the code in lines 367-371)  when I got rid of the gen2 version. Apologies!

Note this also (sort of) explains the weird syntax I went with; I was probably cargoculting what is done in many other places in the stack when we want stuff to be able to run on both gen2 and gen3 (eg [here](https://github.com/lsst/pipe_tasks/blob/387f8f07a2b66205f9fa6bda9a89dcdbbef3f64c/python/lsst/pipe/tasks/dcrAssembleCoadd.py#L542-L547)), where the first `if` is really just the beginning of the handle-gen2 code block. Except in my case, I just had the one check, so it really could have been a single `if` statement (in fact I used  to deal with this [somewhat differently](https://github.com/lsst/pipe_tasks/blob/b64fabfa9b71ed0b1609c282480c22d7da685a86/python/lsst/pipe/tasks/measureExtendedPsf.py#L142-L145), and closer to what you suggested here, which is probably why I wrote things this way before switching to the syntax we see here).

Anyway - I've removed it all now, as I should have from the get go. Sorry again.

The good thing is this pushed me into testing a bunch of things, including rerunning on dataIds where I used to have trouble in gen2, and can confirm it does not occur in gen3. It also made me realize I had to file and fix [DM-29862](https://jira.lsstcorp.org/browse/DM-29862)


TL;DR: this code should have never made it to the PR (it's how I used to solve a gen2-specific problem)

PR #492, opened 2021-03-31
DM-28803: Add capability to discard stars with NaN annuli

leeskelvin:


python/lsst/pipe/tasks/processBrightStars.py
```diff
@@ -163,6 +163,11 @@ class ProcessBrightStarsConfig(pipeBase.PipelineTaskConfig,
         doc="Apply full focal plane sky correction before extracting stars?",
         default=True
     )
+    discardNanFluxStars = pexConfig.Field(
+        dtype=bool,
+        doc="Should stars with NaN annular flux be discarded?",
+        default=False
```
taranu, line 169: I'm confused as to why this default is False when the function default is True, which seems like the more reasonable choice.

MorganSchmitz: See my [comment on the Jira ticket](https://jira.lsstcorp.org/browse/DM-28803?focusedCommentId=273501&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-273501); I think this is how we want things.

leeskelvin: Thanks @MorganSchmitz. Does this address your concern @taranu?

PR #493, opened 2021-04-02
DM-28245: Update ConsolidateVisitSummaryTask to use calexp summaryStats.

erykoff:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1081,12 +1080,12 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         butlerQC.put(expCatalog, outputRefs.visitSummary)
 
     def _combineExposureMetadata(self, visit, dataRefs, isGen3=True):
-        """Make a combined exposure catalog from a list of dataRefs.
+        """Make a combined exposure catalog from a list of calexp dataRefs.
```
MorganSchmitz, line 1083: I think I remember something about trying to avoid using the word "calexp" in user-facing places like docstrings, or did I make that up?

MorganSchmitz (approved):
LGTM! My only comment is not about the change itself, just nitpicking on a docstring clarification you made along the way.

It looks like we're adding a few new fields (zenithDistance etc) we got from the new summaryStatistics. Maybe it's worth mentioning this in the body of the commit message, just so we can easily pinpoint when that was done?

PR #497, opened 2021-04-13
DM-29673 Add centre of mass centroiding to support donut images

mfisherlevine:
SdssCentroids fit donuts very poorly, and a simple CoM
centroid provides a pretty robust measurement when seeded
with the position using the existing algorithm.

This is not expected to deal with overlapping donuts as it
is used to simply to give a cutout to CWFS analysis.

### DIFF
--- a/python/lsst/pipe/tasks/quickFrameMeasurement.py
+++ b/python/lsst/pipe/tasks/quickFrameMeasurement.py
@@ -231,7 +231,7 @@ class QuickFrameMeasurementTask(pipeBase.Task):
         return medianXx, medianYy
 
     @staticmethod
-    def _getCentreOfMass(exp, nominalCentroid, boxSize):
+    def _getCenterOfMass(exp, nominalCentroid, boxSize):
         """Get the centre of mass around a point in the image.
 
         Parameters
@@ -455,7 +455,7 @@ class QuickFrameMeasurementTask(pipeBase.Task):
         ----------
         exp : `lsst.afw.image.Exposure`
             The exposure in which to find and measure the brightest star.
-        donutSize : `int`, optional
+        donutDiameter : `int`, optional
             The expected diameter of donuts in pixels for use in the centre of
             mass centroid measurement. If None is provided, the config option
             is used.
@@ -563,7 +563,7 @@ class QuickFrameMeasurementTask(pipeBase.Task):
         result.brightestObjCentroidCofM = None
         try:
             boxSize = donutDiameter * 1.3  # allow some slack, as cutting off side of donut is very bad
-            centreOfMass = self._getCentreOfMass(exp, brightestObjCentroid, boxSize)
+            centreOfMass = self._getCenterOfMass(exp, brightestObjCentroid, boxSize)
             result.brightestObjCentroidCofM = centreOfMass
         except Exception:
             pass
--- a/python/lsst/pipe/tasks/quickFrameMeasurement.py
+++ b/python/lsst/pipe/tasks/quickFrameMeasurement.py
@@ -20,8 +20,11 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import numpy as np
+import scipy.ndimage as ndImage
+
 import lsst.afw.detection as afwDetect
 import lsst.afw.table as afwTable
+import lsst.geom as geom
 import lsst.meas.base as measBase
 import lsst.daf.base as dafBase
 import lsst.pipe.base as pipeBase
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/quickFrameMeasurement.py
```diff
@@ -227,6 +237,42 @@ def _calcMedianXxYy(objData):
...
+        Returns
+        -------
+        com : `tuple` of `float`
+            The locaiton of the centre of mass of the brightest source in pixel
+            coordinates.
+        """
+        centroidPoint = geom.Point2I(nominalCentroid)
+        extent = geom.Extent2I(1, 1)
+        bbox = geom.Box2I(centroidPoint, extent)
+        bbox = bbox.dilatedBy(int(boxSize//2))
```
timj, line 265: Isn't `boxsize//2` already going to be an `int`? I think this means the casting you are doing here is because you can't trust the caller to give you an int despite the parameter typing. Might be best to change the docstring to say that it accepts int or float.

python/lsst/pipe/tasks/quickFrameMeasurement.py
```diff
@@ -227,6 +237,42 @@ def _calcMedianXxYy(objData):
...
 
+    @staticmethod
+    def _getCentreOfMass(exp, nominalCentroid, boxSize):
+        """Get the centre of mass around a point in the image.
+
+        Parameters
+        ----------
+        exp : `lsst.afw.image.Exposure`
+            The exposure in question.
+
```
timj, line 248: Remove the blank lines between parameters.

python/lsst/pipe/tasks/quickFrameMeasurement.py
```diff
@@ -408,7 +455,7 @@ def _makeEmptyReturnStruct():
         result.medianXxYy = (np.nan, np.nan)
         return result
 
-    def run(self, exp, doDisplay=False):
+    def run(self, exp, *, donutDiameter=None, doDisplay=False):
```
timj, line 458: Am I missing where donutDiameter is documented below?

python/lsst/pipe/tasks/quickFrameMeasurement.py
```diff
@@ -227,6 +237,42 @@ def _calcMedianXxYy(objData):
         medianYy = np.nanmedian([element['xx'] for element in objData.values()])
         return medianXx, medianYy
 
+    @staticmethod
+    def _getCentreOfMass(exp, nominalCentroid, boxSize):
```
timj, line 241: Don't we have to call this `_getCenterOfMass` ?

timj (commented):
Some quick comments. Don't call this a review because I haven't really looked at the algorithmic code.

python/lsst/pipe/tasks/quickFrameMeasurement.py
mfisherlevine: The type of a (float//2) is still a float, albeit one with a decimal part equal to zero, so the `int` cast for the `dilatedBy()` call which only accepts true ints is still sadly necessary.
I can change the doc too though, as it really doesn't matter what's passed (it gets multiplied by 1.3 anyway so likely gets floatified anyhow)

python/lsst/pipe/tasks/quickFrameMeasurement.py
mfisherlevine: Ah, didn't realise that was the spec, thanks.

python/lsst/pipe/tasks/quickFrameMeasurement.py
mfisherlevine: Not _exactly_, I renamed the parameter from `donutSize` and did it everywhere in the code but just forgot the docstring 🤦

python/lsst/pipe/tasks/quickFrameMeasurement.py
mfisherlevine: Oh yeah, probably.

python/lsst/pipe/tasks/quickFrameMeasurement.py
timj: Yes, my point was that the int exists because you don't trust the caller to give you an int and therefore you may as well give up and admit that you routine accepts any number type.

mfisherlevine: Well, that's not primarily why that's there - even if I get an int, I multiply it by 1.3, so it's no longer an int at that point, so it always need that call regardless. But sure, I'll make the doc read int or float too, because it really doesn't matter.

timj: Now that I look, you should have simply said that parameter was a float and not an int and everything would be fine. It was incorrectly documenting the parameter as an int that started this whole discussion.

mfisherlevine: Are you saying I should go and change the type in the pexConfig at the top too? Why does any of this matter at all?

timj: I'm a tad confused. All I wanted was the docstring to say it works for int and float -- you were already calling this routine with a float and not an int. pex_config will validate the type you want it to be and the context for that config item doesn't seem to have any bearing on the call signature of this function.

mfisherlevine: OK, it just seemed like, given that they're used in the same place in the same way (depending on which is supplied), that they should match. Honestly though, none of this actually matters, it's not even that likely that anyone will ever even change or supply this parameter in reality (let alone have its type matter), it's just not OK to hard code things.

mfisherlevine: I'll leave it as an int in the config, and an "int of float" in the `run()` docstring for now, unless you think something else needs changing.

python/lsst/pipe/tasks/quickFrameMeasurement.py
```diff
@@ -517,4 +559,13 @@ def _run(self, exp, doDisplay=False):
         result.brightestObjApFlux70 = brightestObjApFlux70
         result.brightestObjApFlux25 = brightestObjApFlux25
         result.medianXxYy = medianXxYy
+
+        result.brightestObjCentroidCofM = None
+        try:
+            boxSize = donutDiameter * 1.3  # allow some slack, as cutting off side of donut is very bad
```
czwa, line 565: If missing the side is bad, should you check for that, or will you just get an exception and return None, indicating that there was a problem?  I'm also not clear what would cause this block to raise.  _getCenterOfMass doesn't directly raise, and I don't see why anything else would raise either.  Am I just missing something obvious>

czwa (approved):

python/lsst/pipe/tasks/quickFrameMeasurement.py
mfisherlevine: It's not that bad, it just means the result won't be very good, and checking for it is definitely not easy (all of which we discussed OOB, but just noting here too). Also, downstream code with see that, and it's not a huge problem anyway (plus that code is responsible for passing the boxsize anyway).

You're right about the `try`, and I've removed it completely now (having run some tests to make sure I'm definitely OK if the boxsize goes off the edge of the chip).

PR #498, opened 2021-04-13
DM-28853: Improve WCS functionality within fake source injection

jmeyers314:


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -41,6 +41,110 @@
 __all__ = ["InsertFakesConfig", "InsertFakesTask"]
 
 
+def _add_fake_sources(image, objects, calibFluxRadius=12.0, logger=None):
+    """Add fake sources to the given image
+
+    Parameters
+    ----------
+    image : `lsst.afw.image.exposure.exposure.ExposureF`
```
leeskelvin, line 49: Looking elsewhere in the stack (e.g., in [meas_algorithms](https://github.com/lsst/meas_algorithms/blob/6590b197636fb1452f56009ad2efc69934c71ecb/python/lsst/meas/algorithms/detection.py#L211)), I think that an exposure type normally seems to be referred to as `exposure`, or similar. I worry that `image` is a little confusing, especially as `image.image` is called elsewhere in the script.

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -41,6 +41,110 @@
 __all__ = ["InsertFakesConfig", "InsertFakesTask"]
 
 
+def _add_fake_sources(image, objects, calibFluxRadius=12.0, logger=None):
```
leeskelvin, line 44: Top level comment: as we've decided we want to move away from use of the term 'fake', this should probably be changed in the future (e.g., to 'synthetic'). This is probably best addressed in future tickets such as DM-29628 however, so happy to leave as-is for now.

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -365,27 +468,135 @@ def run(self, fakeCat, image, wcs, photoCalib):
...
-            image = self.addFakeSources(image, galImages, "galaxy")
-            image = self.addFakeSources(image, starImages, "star")
+                generator = self._generateGSObjectsFromImages(image, fakeCat)
+            _add_fake_sources(image, generator, calibFluxRadius=self.config.calibFluxRadius, logger=self.log)
         elif len(fakeCat) == 0 and self.config.doProcessAllDataIds:
             self.log.warn("No fakes found for this dataRef; processing anyway.")
         else:
             raise RuntimeError("No fakes found for this dataRef.")
 
+        # restore originals
```
leeskelvin, line 480: or, more explicitly, `restore original exposure WCS and photoCalib`

leeskelvin (approved):

python/lsst/pipe/tasks/insertFakes.py
jmeyers314: I agree.  I'll switch this one to `exposure` and other similar cases that are newly introduced in this ticket.  For now I'll leave `image` where it's similarly used in the pre-existing methods though.

PR #499, opened 2021-04-14
DM-29703: Add imageDifference and forcedPhotDiffim to DRP pipeline

yalsayyad:


pipelines/_DiffimDRP.yaml
```diff
@@ -1,21 +1,50 @@
...
+    config:
+        doDecorrelation: True
+        detection.thresholdValue: 5.0
+        doSelectSources: False
+        doScaleDiffimVariance: True
+        coaddName: goodSeeing
+        getTemplate.coaddName: goodSeeing
+        getTemplate.warpType: direct
+        connections.coaddName: goodSeeing
+        # outputCoaddName isn't filling template for contract check
```
erykoff, line 27: I don't understand this comment.

erykoff (approved):

pipelines/_DiffimDRP.yaml
yalsayyad: Yeah..... I'll file a ticket and put a TO DO:
`coaddExposures="{fakesType}{coaddName}Coadd{warpTypeSuffix}"` and the defaults are:
```
                                     defaultTemplates={"coaddName": "deep",
                                                       "skyMapName": "deep",
                                                       "warpTypeSuffix": "",
                                                       "fakesType": ""}
```
If I specify the template as `coaddName: goodSeeing`, I'd expect that `coaddExposures` would resolve to `goodSeeingCoadd`. And I believe it does, but the contracts checker doesn't recognize it. So I'm over-specifying the connections here, so make the contracts checker happy.

PR #508, opened 2021-05-04
DM-29344: Investigate the CI differences between Gen 2 and 3 in COSMOS field

mrawls:
This PR makes `doScaleDiffimVarianceTask` True to be the new default behavior in imageDifferenceTask. We won't need to override this config in any pipeline or ap_verify-style dataset to get consistent results.


isullivan (approved):
This change overlaps with DM-29869, but once you and @kherner decide which one should be merged, this is fine.

PR #515, opened 2021-05-12
DM-29869: Make doScaleDiffimVariance=True the default in ImageDifferenceTask

kherner:
move some task defaults from DRP.yaml to setDefaults in imageDifference.py. Should be quick.


mrawls (commented):
These changes look self-consistent and OK. Please clean up the commit history, I think all this can be in just 1-2 commits, and I'm confused why you have some kind of merge commit. Keep these guidelines in mind https://developer.lsst.io/work/flow.html#appendix-commit-organization-best-practices

mrawls (approved):

PR #516, opened 2021-05-12
DM-30149: Check if the TransformDiaSourceCatalog pipeline task can be used on DC2 data for use in DP0

morriscb:
Add TransformDiaSourceCatalogTask to DRP pipeline.

### DIFF
--- a/pipelines/DRP.yaml
+++ b/pipelines/DRP.yaml
@@ -58,6 +58,14 @@ tasks:
         # Setting template 'coaddName' is not sufficient to satisfy contract
         connections.coaddExposures: goodSeeingCoadd
         connections.subtractedExposure: goodSeeingDiff_differenceExp
+  transformDiaSourceCat:
+    class: lsst.ap.association.TransformDiaSourceCatalogTask
+    config:
+      connections.coaddName: goodSeeing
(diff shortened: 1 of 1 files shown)


pipelines/DRP.yaml
```diff
@@ -58,6 +58,14 @@ tasks:
         # Setting template 'coaddName' is not sufficient to satisfy contract
         connections.coaddExposures: goodSeeingCoadd
         connections.subtractedExposure: goodSeeingDiff_differenceExp
+  transformDiaSourceCat:
+    class: lsst.ap.association.TransformDiaSourceCatalogTask
+    config:
+      connections.coaddName: goodSeeing
+      connections.diaSourceSchema: goodSeeingDiff_diaSrc_schema
+      connections.diaSourceCat: goodSeeingDiff_diaSrc
+      connections.diffIm: goodSeeingDiff_differenceExp
```
yalsayyad, line 67: Other than the template `connections.coaddName: goodSeeing`, the only connection that needs to be explicitly set here is `connections.diffIm` which is *only* to satisfy the contract checker.  I filed DM-30210 for the contract-checker issue.

yalsayyad (approved):

PR #525, opened 2021-06-02
DM-30193: Add MakeVisitTableTask to postprocess.py

cmsaunders:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1300,75 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+        for visitSummary in visitSummaries:
+            visitSummary = visitSummary.get()
+
+            visitRow = visitSummary[0]
+            visitInfo = visitRow.getVisitInfo()
+            visitEntry = {}
+            visitEntry["visitId"] = visitRow['visit']
+            visitEntry["filterName"] = visitRow['physical_filter']
+            visitEntry["ra"] = np.mean(visitSummary['ra'])
+            visitEntry["decl"] = np.mean(visitSummary['decl'])
```
erykoff, line 1360: The position obtained from the mean of all the detectors is not going to be the same as the boresight position.  I'm not sure which one we want, but we need to be consistent.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1300,75 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+            visitEntry["skyRotation"] = visitInfo.getBoresightRotAngle().asRadians()
+            azimuth, altitude = visitInfo.getBoresightAzAlt()
```
erykoff, line 1362: The problem here is that the az/alt is going to be inconsistent with the ra/dec above, and that could cause much confusion.  We need to make sure that they are consistent.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1300,75 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+            visitEntry["azimuth"] = azimuth.asRadians()
+            visitEntry["altitude"] = altitude.asRadians()
```
erykoff, line 1364: All units in this table according to the baseline schema should be degrees, e.g. https://github.com/lsst/sdm_schemas/blob/master/yml/baselineSchema.yaml#L5527

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1300,75 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+            visitEntry["zenithDistance"] = visitRow['zenithDistance']
+            visitEntry["airmass"] = visitInfo.getBoresightAirmass()
```
erykoff, line 1366: Note that the zenith distance in the visit summary table is computed at the center of each ccd.  It is not the boresight zenith distance.  I'm not sure which we want for this table (I personally don't know what "observation mid-point" means technically).  But right now this code snippet will set the zenith distance to the first detector, and the airmass to the boresight.  We need to ensure these are consistent.

Furthermore, "airmass" is a slippery quantity.  There are many ways of computing airmass, and I don't know which one is supposed to be here.  "zenithDistance" is unambiguous, thankfully.

erykoff (changes requested):
Apologies for the drive-by review, but I noticed the quantities used were not self-consistent.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1299,76 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
         """No config to write.
         """
         pass
+
+
+class MakeVisitTableConnections(pipeBase.PipelineTaskConnections,
+                                dimensions=("instrument",),
+                                defaultTemplates={}):
+    visitSummaries = pipeBase.connectionTypes.Input(
```
leeskelvin, line 1307: could this just be `connectionTypes`, instead of `pipeBase.connectionTypes`?

leeskelvin (approved):

PR #527, opened 2021-06-04
Add MakeCcdVisitTableTask to postprocess.py

cmsaunders:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1299,82 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+
+
+class MakeCcdVisitTableTask(CmdLineTask, pipeBase.PipelineTask):
+    """Produce a `ccdVisitTable` from the `visitSummary` exposure catalogs.
+    """
+    _DefaultName = 'makeCcdVisitTable'
+    ConfigClass = MakeCcdVisitTableConfig
+
+    def run(self, visitSummaries):
+        """ Make a table of ccd information from the `visitSummary` catalogs
```
erykoff, line 1335: Need a space after this line, no space at the beginning of the docstring.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1299,82 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+        Parameters
+        ----------
+        visitSummaries : `list` of `lsst.afw.table.ExposureCatalog`
```
erykoff, line 1338: This is a list of deferred datasets, not exposure catalogs, right?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1299,82 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+            List of exposure catalogs with per-detector summary information.
+        Returns
+        -------
+        result : `lsst.pipe.Base.Struct`
+            Results struct with attribute:
+                - `outputCatalog`
+                    Catalog of ccd and visit information.
+        """
+        ccdEntries = []
+        for visitSummary in visitSummaries:
```
erykoff, line 1348: This should be `visitSummaryRef in visitSummaryRefs` or something like that.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1299,82 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+            visitSummary = visitSummary.get()
+            visitInfo = visitSummary[0].getVisitInfo()
+
+            ccdEntry = {}
+            summaryTable = visitSummary.asAstropy()
+            selectColumns = ['id', 'visit', 'physical_filter', 'ra', 'decl', 'zenithDistance', 'zeroPoint',
+                             'skyBg', 'skyNoise', ]
+            ccdEntry = summaryTable[selectColumns].to_pandas().set_index('id')
+            ccdEntry.rename({"physical_filter": "filterName", "visitId": "visit"})
```
erykoff, line 1357: Should this be `"visit": "visitId"`?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1299,82 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+            ccdEntry['ccdVisitId'] = [row.getVisitInfo().getExposureId() for row in visitSummary]
```
erykoff, line 1358: This is not correct.  The `getExposureId()` call does not give what you want, and should not be used for anything (it's per visit but combines ccd info, but from a totally random ccd I think?).  See https://jira.lsstcorp.org/browse/RFC-459 which has never been implemented (sigh).  Right now I think we need to use the butler packer infrastructure for this.  But this is something that should be brought up on slack on what value we want here.  But it's is definitely not this.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1299,3 +1299,82 @@ def writeConfig(self, butler, clobber=False, doBackup=True):
...
+            ccdEntry['llcra'] = summaryTable['raCorners'][:, 0]
+            ccdEntry['llcdec'] = summaryTable['decCorners'][:, 0]
+            ccdEntry['ulcra'] = summaryTable['raCorners'][:, 1]
+            ccdEntry['ulcdec'] = summaryTable['decCorners'][:, 1]
+            ccdEntry['urcra'] = summaryTable['raCorners'][:, 2]
+            ccdEntry['urcdec'] = summaryTable['decCorners'][:, 2]
+            ccdEntry['lrcra'] = summaryTable['raCorners'][:, 3]
+            ccdEntry['lrcdec'] = summaryTable['decCorners'][:, 3]
+            # TODO: DM-30618, Add raftName, nExposures, ccdTemp, binX, binY, seeing, and flags,
+            # and decide if WCS, and llcx, llcy, ulcx, ulcy, etc. values are actually wanted.
```
erykoff, line 1376: seeing should be obtainable from the psf size in the summary.

erykoff (changes requested):
A few tweaks here, looks good.  The most serious issue is what to put in `ccdVisitId`.

python/lsst/pipe/tasks/postprocess.py
cmsaunders: Ok, I compared `getExposureId()` with the gen2 method `dataRef.get('ccdExposureId')` (based on what is used here: https://github.com/lsst/pipe_tasks/blob/1e789892caf5a6dcf760974bf4037902b9f858dc/python/lsst/pipe/tasks/postprocess.py#L302) and got the same results for a few visits and ccds, but maybe I just got lucky. I'll bring it up on slack.

erykoff: If this is from the `visitInfo`, then there's just one number per visit.  Presumably we need something that is per detector here?

cmsaunders: For this, `getVisitInfo()` is called for each ccd in the visitSummary. Most of the attributes in `getVisitInfo` are the same for all of the rows, but ExposureId is different. For example, for visit 30500, you get 
```
ipdb> ccdEntry['ccdVisitId']
id
0      6100000
1      6100001
2      6100002
3      6100003
4      6100004
        ...   
99     6100099
100    6100100
101    6100101
102    6100102
103    6100103
Name: ccdVisitId, Length: 103, dtype: int64
```

cmsaunders: @TallJimbo helped with this question and a packer is used now.

erykoff (approved):
Thank you, all the comments are addressed.  I think we also should add this task to the pipeline and then it'll be automatically run, and tested with `ci_hsc_gen3` (otherwise, this is not tested at all as far as I can tell.). If this is what we decide to do, you should add a comment that it is tested as part of `ci_hsc_gen3`.

PR #529, opened 2021-06-08
DM-30054: Add LoadReferenceCatalogTask

erykoff:
This task loads multi-band reference sources, applies color terms, and transforms them in one convenient task.

### DIFF
--- a/doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask.rst
+++ b/doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask.rst
@@ -1,0 +1,42 @@
+.. lsst-task-topic:: lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask
+
+########################
+LoadReferenceCatalogTask
+########################
+
+``LoadReferenceCatalogTask`` is a convenience task that combines the reference object loader adapted for multiple filters; color term application; and reference selection in one task.
+This task returns a numpy record array with the magnitude information for each of the physical filters requested by the caller.
+
+If appropriate, all proper motion corrections are handled by the reference object loader as configured.
+Mapping from physical filter to reference filter is handled by the ``filterMap`` configured with the reference object loader, as well as the color term library.
+
+The format of the reference catalogs returned by this task will be a numpy record array with the following datatype:
+
+.. code-block:: python
+
+    import numpy as np
+    dtype = [('ra', 'np.float64'),
+             ('dec', 'np.float64'),
+             ('refMag', 'np.float32', (len(filterList), )),
+             ('refMagErr', 'np.float32', (len(filterList), ))]
+
+Reference magnitudes (AB) and errors will be 99 for non-detections for a given reference filter.
+
+.. _lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask-api:
+
+Python API summary
+==================
+
+.. lsst-task-api-summary:: lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask
+
+.. _lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask-subtasks:
+
+Retargetable subtasks
+=====================
+
+.. lsst-task-config-subtasks:: lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask
+
+Configuration fields
+====================
+
+.. lsst-task-config-fields:: lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask
--- a/python/lsst/pipe/tasks/loadReferenceCatalog.py
+++ b/python/lsst/pipe/tasks/loadReferenceCatalog.py
@@ -48,7 +48,7 @@ class LoadReferenceCatalogConfig(pexConfig.Config):
         doc="Reference object loader for photometry",
     )
     doApplyColorTerms = pexConfig.Field(
-        doc=("Apply photometric color terms to reference stars?"
+        doc=("Apply photometric color terms to reference stars? "
              "Requires that colorterms be set to a ColorTermLibrary"),
         dtype=bool,
         default=True
@@ -115,9 +115,9 @@ class LoadReferenceCatalogTask(pipeBase.Task):
                            bboxToSpherePadding=None):
         """Get a multi-band reference catalog by specifying a bounding box and WCS.
 
-        The catalog will be in `numpy.ndarray`, with positions proper-motion corrected to
-        "epoch" (if specified, and if the reference catalog has proper motions);
-        sources cut on a reference selector (if
+        The catalog will be in `numpy.ndarray`, with positions proper-motion
+        corrected to "epoch" (if specified, and if the reference catalog has
+        proper motions); sources cut on a reference selector (if
         "config.doReferenceSelection = True"); and color-terms applied (if
         "config.doApplyColorTerms = True").
 
@@ -128,7 +128,8 @@ class LoadReferenceCatalogTask(pipeBase.Task):
                  ('refMag', 'np.float32', (len(filterList), )),
                  ('refMagErr', 'np.float32', (len(filterList), ))]
 
-        Reference magnitudes (AB) will be 99 for non-detections in a given band.
+        Reference magnitudes (AB) and errors will be 99 for non-detections
+        in a given band.
 
         Parameters
         ----------
@@ -140,8 +141,8 @@ class LoadReferenceCatalogTask(pipeBase.Task):
         filterList : `List` [ `str` ]
             List of camera physicalFilter names to retrieve magnitudes.
         epoch : `astropy.time.Time`, optional
-            Epoch to which to correct proper motion and parallax (if available),
-            or `None` to not apply such corrections.
+            Epoch to which to correct proper motion and parallax
+            (if available), or `None` to not apply such corrections.
         bboxToSpherePadding : `int`, optional
             Padding to account for translating a set of corners into a
             spherical (convex) boundary that is certain to encompass the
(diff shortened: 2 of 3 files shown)


tests/test_loadReferenceCatalog.py
```diff
@@ -0,0 +1,204 @@
...
+
+_synthFlux = 100.0
+_synthCenter = lsst.geom.SpherePoint(30, -30, lsst.geom.degrees)
+
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
+
+# class TrivialLoader(LoadReferenceObjectsTask):
```
leeskelvin, line 30: Does this comment still need to be here?

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+
+
+class LoadReferenceCatalogConfig(pexConfig.Config):
+    """Config for LoadReferenceCatalogTask"""
+    refObjLoader = pexConfig.ConfigurableField(
+        target=LoadIndexedReferenceObjectsTask,
+        doc="Reference object loader for photometry",
+    )
+    doApplyColorTerms = pexConfig.Field(
+        doc=("Apply photometric color terms to reference stars?"
```
leeskelvin, line 51: This needs an additional space following the '?'

doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.loadReferenceCatalog.LoadReferenceCatalogTask.rst
```diff
@@ -0,0 +1,42 @@
...
+
+.. code-block:: python
+
+    import numpy as np
+    dtype = [('ra', 'np.float64'),
+             ('dec', 'np.float64'),
+             ('refMag', 'np.float32', (len(filterList), )),
+             ('refMagErr', 'np.float32', (len(filterList), ))]
+
+Reference magnitudes (AB) will be 99 for non-detections for a given reference filter.
```
leeskelvin, line 23: Reference magnitude errors will also be set to this sentinel value too, right? Would it be helpful to mention that explicitly here?

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+        "config.doApplyColorTerms = True").
+
+        The format of the reference catalog will be of the format:
+
+        dtype = [('ra', 'np.float64'),
+                 ('dec', 'np.float64'),
+                 ('refMag', 'np.float32', (len(filterList), )),
+                 ('refMagErr', 'np.float32', (len(filterList), ))]
+
+        Reference magnitudes (AB) will be 99 for non-detections in a given band.
```
leeskelvin, line 131: As per the comment above, perhaps adding "and errors" here (and elsewhere this is mentioned) could be useful.

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+            else:
+                # Gen2
+                refCatName = self.refObjLoader.ref_dataset_name
+
+            for i, (filterName, fluxField) in enumerate(zip(self._fluxFilters, self._fluxFields)):
+                if fluxField is None:
+                    # There is no matching reference band.
+                    # This will leave the column filled with 99s
+                    continue
+                self.log.debug(f"Applying color terms for filtername={filterName}")
```
leeskelvin, line 278: This should probably be camelCase `filterName`, to match the rest of the script. Also, a nice shortcut here to avoid typing out the object name twice would be to use `self.log.debug(f"Applying color terms for {filterName=}")` (which also adds single-quotes around the string output too).

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+        self._fluxFilters = None
+        self._fluxFields = None
+        self._referenceFilter = None
+
+    def getPixelBoxCatalog(self, bbox, wcs, filterList, epoch=None,
+                           bboxToSpherePadding=None):
+        """Get a multi-band reference catalog by specifying a bounding box and WCS.
+
+        The catalog will be in `numpy.ndarray`, with positions proper-motion corrected to
+        "epoch" (if specified, and if the reference catalog has proper motions);
```
leeskelvin, line 119: These docstrings are flagged up by my flake8 as being too long (>79 characters). Is this your understanding too from reading [the dev guide](https://developer.lsst.io/python/style.html#docstring-and-comment-line-length-must-be-less-than-or-equal-to-79-columns)?

python/lsst/pipe/tasks/loadReferenceCatalog.py
leeskelvin, line 131: This docstring is also too long (by 1 char).

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+        ----------
+        bbox : `lsst.geom.Box2I`
+            Box which bounds a region in pixel space.
+        wcs : `lsst.afw.geom.SkyWcs`
+            Wcs object defining the pixel to sky (and reverse) transform for
+            the supplied bbox.
+        filterList : `List` [ `str` ]
+            List of camera physicalFilter names to retrieve magnitudes.
+        epoch : `astropy.time.Time`, optional
+            Epoch to which to correct proper motion and parallax (if available),
```
leeskelvin, line 143: This docstring is also too long (by 1 char).

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+            refCat = skyBox.refCat
+
+        return self._formatCatalog(refCat, filterList)
+
+    def getSkyCircleCatalog(self, center, radius, filterList, epoch=None,
+                            catalogFormat='numpy'):
+        """Get a multi-band reference catalog by specifying a center and radius.
+
+        The catalog will be in `numpy.ndarray`, with positions proper-motion corrected to
+        "epoch" (if specified, and if the reference catalog has proper motions);
```
leeskelvin, line 176: These docstrings are also too long.

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+        "config.doApplyColorTerms = True").
+
+        The format of the reference catalog will be of the format:
+
+        dtype = [('ra', 'np.float64'),
+                 ('dec', 'np.float64'),
+                 ('refMag', 'np.float32', (len(filterList), )),
+                 ('refMagErr', 'np.float32', (len(filterList), ))]
+
+        Reference magnitudes (AB) will be 99 for non-detections in a given band.
```
leeskelvin, line 188: Also adding "and errors" here, as per comments above.

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+                    continue
+                self.log.debug(f"Applying color terms for filtername={filterName}")
+
+                colorterm = self.config.colorterms.getColorterm(filterName, refCatName, doRaise=True)
+
+                refMag, refMagErr = colorterm.getCorrectedMagnitudes(refCat)
+
+                # nan_to_num replaces nans with zeros, and this ensures that we select
+                # magnitudes that both filter out nans and are not very large (corresponding
+                # to very small fluxes), as "99" is a commen sentinel for illegal magnitudes.
```
leeskelvin, line 286: These docstrings are >79 characters.

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+                good, = np.where((np.nan_to_num(refMag[selected]) < 90.0)
+                                 & (np.nan_to_num(refMagErr[selected]) < 90.0)
+                                 & (np.nan_to_num(refMagErr[selected]) > 0.0))
+
+                npRefCat['refMag'][good, i] = refMag[selected][good]
+                npRefCat['refMagErr'][good, i] = refMagErr[selected][good]
+        else:
+            # No color terms to apply
+            for i, (filterName, fluxField) in enumerate(zip(self._fluxFilters, self._fluxFields)):
+                # nan_to_num replaces nans with zeros, and this ensures that we select
```
leeskelvin, line 297: This docstrings is >79 characters.

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+                self.log.debug(f"Applying color terms for filtername={filterName}")
+
+                colorterm = self.config.colorterms.getColorterm(filterName, refCatName, doRaise=True)
+
+                refMag, refMagErr = colorterm.getCorrectedMagnitudes(refCat)
+
+                # nan_to_num replaces nans with zeros, and this ensures that we select
+                # magnitudes that both filter out nans and are not very large (corresponding
+                # to very small fluxes), as "99" is a commen sentinel for illegal magnitudes.
+
```
leeskelvin, line 287: I think it might make more sense for this newline to be removed, so the above comment block is clearly attached to the code it annotates.

python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -0,0 +1,380 @@
...
+                good, = np.where((np.nan_to_num(refMag[selected]) < 90.0)
+                                 & (np.nan_to_num(refMagErr[selected]) < 90.0)
+                                 & (np.nan_to_num(refMagErr[selected]) > 0.0))
```
leeskelvin, line 290: I'm curious why you're not using `np.nan_to_num(..., nan=99)` here (and elsewhere) to set NaNs to your sentinel value of 99, rather than 0?

python/lsst/pipe/tasks/loadReferenceCatalog.py
erykoff: That's handy!  But on another ticket I was admonished (again) that for logging (especially debug) I should be using the `%s` formatting for lazy-evaluation.  (I doubt it matters much for performance, but it's the "right" way to log).

python/lsst/pipe/tasks/loadReferenceCatalog.py
erykoff: This is very useful here, good point, I didn't know this.  But below, with fluxes, I want to protect against 0/negative fluxes so the default nan -> 0 is appropriate.

tests/test_loadReferenceCatalog.py
erykoff: This was leftover from earlier code, and should be removed.  Thanks!

python/lsst/pipe/tasks/loadReferenceCatalog.py
leeskelvin: Ah, I wasn't aware that there was a 'correct' way to log debug data, thank you for letting me know! In which case, the proposal you suggest sounds good.

PR #532, opened 2021-06-14
DM-30673: Update docs and function names in associationUtils.py

morriscb:


python/lsst/pipe/tasks/associationUtils.py
```diff
@@ -1,39 +1,116 @@
...
 
 
 def toRaDec(nside, index):
-    """Convert from healpix index to ra,dec in degrees"""
+    """Convert from healpix index to ra,dec in degrees
+
+    Parameters
+    ----------
+    nside : `int`
+        Resolution of healpixel "grid">
```
isullivan, line 57: typo: >

python/lsst/pipe/tasks/associationUtils.py
```diff
@@ -1,39 +1,116 @@
...
+    index : `int`
+        Index of the healpix we want to find the location of.
```
isullivan, line 59: "Index of the healpix pixel"...

python/lsst/pipe/tasks/associationUtils.py
```diff
@@ -45,36 +122,72 @@ def eq2vec(ra, dec):
...
+    nside : `int`
+        Resolution of the healpixels to search/return.
+    ra : `float`
+        RA in degrees.
+    dec : `float`
+        Declination in degrees
+    max_rad : `float`
+        Max distance in radians to search nearby healpixels.
+    min_rad : `float`, optional
+        Minimum distance to search healpixels. Default = 0.
```
isullivan, line 176: "Minimum distance in radians"...

python/lsst/pipe/tasks/associationUtils.py
```diff
@@ -45,36 +122,72 @@ def eq2vec(ra, dec):
...
+    Parameters
+    ----------
+    ra : `float`
+        RA in degrees.
+    dec : `float`
+        Declination in degrees
+    """
+    return np.dstack([np.cos(dec*np.pi/180)*np.cos(ra*np.pi/180),
+                      np.cos(dec*np.pi/180)*np.sin(ra*np.pi/180),
+                      np.sin(dec*np.pi/180)])[0]
```
isullivan, line 138: Use`np.radians()`.

python/lsst/pipe/tasks/associationUtils.py
```diff
@@ -1,39 +1,116 @@
...
     return np.array([x, y, z])
 
 
-def eq2vec(ra, dec):
-    """Convert equatorial ra,dec in degrees to x,y,z on the unit sphere parameters"""
+def eq2xyzVec(ra, dec):
+    """Convert equatorial ra,dec in degrees to x,y,z on the unit sphere
+    parameters
+
+    Vetorized version of ``eq2xyz``
```
isullivan, line 100: "Vectorized"

python/lsst/pipe/tasks/associationUtils.py
```diff
@@ -45,36 +122,72 @@ def eq2vec(ra, dec):
...
+        Declination in degrees
+    """
+    return np.dstack([np.cos(dec*np.pi/180)*np.cos(ra*np.pi/180),
+                      np.cos(dec*np.pi/180)*np.sin(ra*np.pi/180),
+                      np.sin(dec*np.pi/180)])[0]
 
 
 def convert_spherical_array(array):
-    """Convert from ra,dec to spherical from array"""
+    """Convert from and a array ra,dec to spherical.
```
isullivan, line 142: "Convert from an array of ra,dec to spherical"?

tests/test_associationUtils.py
```diff
@@ -0,0 +1,63 @@
...
+
+class TestAssociationUtils(lsst.utils.tests.TestCase):
+
+    def test_queryDisc(self):
+        """Test that doing a disc query of healpy using our wrapper works as
+        expected.
+
+        Most other calculations are very simple and not worth testing.
+        """
+        # Find nearest pixel to r=45, dec=45.
```
isullivan, line 38: Please define the test RA and Dec as variables, and pass those to your functions. Also, please use different numbers for Ra and Dec, so that this test will catch if the order of parameters gets switched. Maybe `raInit = 225` and `decInit = 45`, for example.

tests/test_associationUtils.py
```diff
@@ -0,0 +1,63 @@
...
+        nside = 128
+        centerPixNumber = hp.ang2pix(nside, 45, 45, lonlat=True)
+        ra, dec = hp.pix2ang(nside, centerPixNumber, lonlat=True)
```
isullivan, line 41: Perhaps add a test that the healpix pixel ra and dec are less than the size of a pixel away from the initial ra and dec.

isullivan (approved):
I found a few typos, and had a couple suggestions for the unit test.

tests/test_associationUtils.py
morriscb: Okay, this may lead to a change in the number of pixels found in the test just as a warning.

tests/test_associationUtils.py
morriscb: This would be testing healpy directly vs testing the wrappers that we have written so I'm not included to add this test. If ang2pix isn't working properly there would be more fundamental problems with healpy going on.

PR #534, opened 2021-06-16
DM-30386: Make DRP wrapper task for DiaCalculation task

morriscb:


python/lsst/pipe/tasks/drpDiaCalculationPipe.py
```diff
@@ -0,0 +1,139 @@
...
+
+import pandas as pd
+
+from lsst.meas.base import DiaObjectCalculationTask
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+
+
+class DrpDiaCalculationPipeConnections(pipeBase.PipelineTaskConnections,
+                                       dimensions=("tract", "patch", "skymap"),
```
kherner, line 38: You pass in skymap here to the PipelinTaskConnections but then it's never used again; the other Input calls use only tract and patch. Is that intentional?

python/lsst/pipe/tasks/drpDiaCalculationPipe.py
```diff
@@ -0,0 +1,139 @@
...
+        dimensions=("tract", "patch"),
+    )
+
+
+class DrpDiaCalculationPipeConfig(
+        pipeBase.PipelineTaskConfig,
+        pipelineConnections=DrpDiaCalculationPipeConnections):
+    filterNames = pexConfig.ListField(
+        dtype=str,
+        default=['u', 'g', 'r', 'i', 'z', 'y'],
```
kherner, line 69: See my comment in the ap_association review about whether you could end up with a list of lists.

kherner (commented):
See comments in drpDiaCalculationPipe.py

python/lsst/pipe/tasks/drpDiaCalculationPipe.py
morriscb: See my response there, the tasks have different context and the list of lists problem won't happen.

python/lsst/pipe/tasks/drpDiaCalculationPipe.py
morriscb: The skymap here defines the tracts/patches. I think I want it here as if I did not and a repo has more than one skymap and therefore data with different tract/patch definitions, I wouldn't want them to grab tracts/patches with the same id numbers across different skymaps as they likely represent different areas of the sky.

kherner (approved):

PR #535, opened 2021-06-16
DM-30701: Add astromOffsetMean and astromOffsetStd to ExposureSummaryStats.

erykoff:
The ComputeExposureSummaryTask has also been moved from characterize to calibrate, where it should be to have the latest-and-greatest photometric calibration and wcs.  In the previous location it was computing incorrect values.

### DIFF
--- a/python/lsst/pipe/tasks/calibrate.py
+++ b/python/lsst/pipe/tasks/calibrate.py
@@ -40,6 +40,8 @@ from lsst.meas.deblender import SourceDeblendTask
 from lsst.pipe.tasks.setPrimaryFlags import SetPrimaryFlagsTask
 from .fakes import BaseFakeSourcesTask
 from .photoCal import PhotoCalTask
+from .computeExposureSummaryStats import ComputeExposureSummaryStatsTask
+
 
 __all__ = ["CalibrateConfig", "CalibrateTask"]
 
@@ -286,6 +288,15 @@ class CalibrateConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Calibrate
         doc="Injection of fake sources for testing purposes (must be "
             "retargeted)"
     )
+    doComputeSummaryStats = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Run subtask to measure exposure summary statistics?"
+    )
+    computeSummaryStats = pexConfig.ConfigurableField(
+        target=ComputeExposureSummaryStatsTask,
+        doc="Subtask to run computeSummaryStats on exposure"
+    )
     doWriteExposure = pexConfig.Field(
         dtype=bool,
         default=True,
@@ -522,6 +533,8 @@ class CalibrateTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
                 photoRefObjLoader = self.photoRefObjLoader
             self.makeSubtask("photoCal", refObjLoader=photoRefObjLoader,
                              schema=self.schema)
+        if self.config.doComputeSummaryStats:
+            self.makeSubtask('computeSummaryStats')
 
         if initInputs is not None and (astromRefObjLoader is not None or photoRefObjLoader is not None):
             raise RuntimeError("PipelineTask form of this task should not be initialized with "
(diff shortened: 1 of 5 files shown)


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -40,6 +40,8 @@
 from lsst.pipe.tasks.setPrimaryFlags import SetPrimaryFlagsTask
 from .fakes import BaseFakeSourcesTask
 from .photoCal import PhotoCalTask
+from .computeExposureSummaryStats import ComputeExposureSummaryStatsTask
+
```
laurenam, line 44: Extra newline?  Although...looking at https://developer.lsst.io/python/style.html?highlight=__all__#standard-code-order-should-be-followed, the `__all__` is in the wrong place (but I think this is pervasive...and certainly not on you on this tickte!)

python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -161,11 +161,15 @@ class CharacterizeImageConfig(pipeBase.PipelineTaskConfig,
     doComputeSummaryStats = pexConfig.Field(
         dtype=bool,
         default=True,
-        doc="Run subtask to measure exposure summary statistics"
+        doc="Run subtask to measure exposure summary statistics",
+        deprecated=("This subtask has been moved to CalibrateTask "
+                    "with DM-30701.")
```
laurenam, line 166: Do you need a "Will be removed after vNN." declaration here (and below)?

I'm glad you caught this before it may have caused some serious puzzlement!  However, this does mean that anyone running `characterizeImage`, but not `calibrate` will not benefit from getting any summary stats that are appropriately computed in that task (namely the PSF related values).   In fact, I'm now wondering if it actually is more appropriate to actually compute the summary stats in both places (where any values that get updated within `calibrate` would get updated to the new ones?

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -193,6 +201,8 @@ def run(self, exposure, sources, background):
                                                 skyNoise=float(skyNoise),
                                                 meanVar=float(meanVar),
                                                 raCorners=raCorners,
-                                                decCorners=decCorners)
+                                                decCorners=decCorners,
+                                                astromOffsetMean=astromOffsetMean,
+                                                astromOffsetStd=astromOffsetStd)
```
laurenam, line 206: Add this new addition to the class level docs above?

tests/test_processCcd.py
```diff
@@ -164,8 +164,8 @@ def testProcessCcd(self):
                         ("summary_decl", summary.decl, -9.800258687592303),
                         ("summary_zenithDistance", float('%.6f' % (summary.zenithDistance)), 42.361001),
                         ("summary_zeroPoint", summary.zeroPoint, 30.940228147639207),
-                        ("summary_skyBg", summary.skyBg, 191.4352211728692),
-                        ("summary_skyNoise", summary.skyNoise, 12.512330367008024),
+                        ("summary_skyBg", summary.skyBg, 191.37726892903447),
+                        ("summary_skyNoise", summary.skyNoise, 12.512272992606531),
```
laurenam, line 168: These are the only two that changed after the (significant!) move of the task from `characterizeImage` to `calibrate`?!  Is the test image basic enough that no other values in the summary stats got modified by `calibrateTask`?  Also, can you now add checks for the newly added astrometry metrics as part of the test?

python/lsst/pipe/tasks/characterizeImage.py
erykoff: I thought about this a bit, and went back and forth (there is some overhead to compute these twice, but I don't think it's necessarily a burden).  But in the end I landed on this because the `characterizeImage` task will output the `icExp` which really isn't a standard data product, while `calibrate` will output a `calexp` which is.  So if I think of these as "calexp summary statistics" it doesn't need to be in both places.  However, I could be persuaded otherwise.

tests/test_processCcd.py
erykoff: These are the only two that change.  And unfortunately, I can't check the astrometry metrics because the astrometry task isn't run as part of `test_processCcd`.  (Nor should it, since that would add quite a bit of testing overhead.)

python/lsst/pipe/tasks/characterizeImage.py
laurenam: Ah, good point about `icExp` vs. `calexp`.  I do think it's good to leave as is (and if anybody decides they'd really want such summaries on the former, it can be added later -- and optionally, to avoid unnecessary duplication of the computation!)

python/lsst/pipe/tasks/calibrate.py
erykoff: There are supposed to be two blank lines between the imports and the `__all__`.

laurenam (approved):

PR #536, opened 2021-06-17
DM-30560: Handle cases of incomputable Psfs in source injection

jmeyers314:
Above error occasionally encountered when evaluating a
CoaddPsf in a region with no underlying exposures.


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -81,7 +81,32 @@ def _add_fake_sources(exposure, objects, calibFluxRadius=12.0, logger=None):
         mat = wcs.linearizePixelToSky(spt, geom.arcseconds).getMatrix()
         gsWCS = galsim.JacobianWCS(mat[0, 0], mat[0, 1], mat[1, 0], mat[1, 1])
 
-        psfArr = psf.computeKernelImage(pt).array
+        try:
+            psfArr = psf.computeKernelImage(pt).array
+        except InvalidParameterError:
+            logger.infof("Cannot compute Psf for object at {}; skipping", pt)
```
morriscb, line 87: The log here is a bit of a misnomer as it the object isn't being skipped just yet assuming that the new point creation doesn't return the same position.

morriscb (approved):
Slight change to one of the log outputs for clarity of what is happening.

PR #542, opened 2021-06-29
DM-30843: Raise RepeatableQuantumError if HealSparsePropertyMapTask given inconsistent inputs.

erykoff:


python/lsst/pipe/tasks/healSparseMapping.py
```diff
@@ -570,7 +578,13 @@ def run(self, sky_map, tract, band, coadd_dict, input_map_dict, visit_summary_di
                 total_inputs[inmap] += 1
 
                 # Retrieve the correct visitSummary row
+                if visit not in visit_summary_dict:
+                    msg = f"Visit {visit} not found in visit_summaries."
```
natelust, line 582: I go back and forth over time with this in my own code, so whatever you want to do is fine, but recently I have been more a fan of putting the message string right in the Exception constructor. Its sometimes less lines, and a bit clearer how they link together

python/lsst/pipe/tasks/healSparseMapping.py
```diff
@@ -570,7 +578,13 @@ def run(self, sky_map, tract, band, coadd_dict, input_map_dict, visit_summary_di
...
+                    raise pipeBase.RepeatableQuantumError(msg)
                 row = visit_summary_dict[visit].find(detector_id)
+                if row is None:
+                    msg = f"Visit {visit} / detector_id {detector_id} not found in visit_summaries."
```
natelust, line 586: same

natelust (approved):

python/lsst/pipe/tasks/healSparseMapping.py
erykoff: You know, I first wrote it as a single call (which I prefer in general!), but the message string was too long for the line and had to be split up anyway.  I think this is more readable than a multi-line string, so I'm going to leave it as is.

PR #548, opened 2021-07-02
DM-29965 Add AL likelihood image as separate output like zogy score image.

kgabor:
* Use scoreImage for AL, too. Deprecate doPreconvole option.
 * Update useScoreImageDetection option description.
 * Change Gaussian pre-convolution to default False.

### DIFF
--- a/python/lsst/pipe/tasks/imageDifference.py
+++ b/python/lsst/pipe/tasks/imageDifference.py
@@ -100,7 +100,6 @@ class ImageDifferenceTaskConnections(pipeBase.PipelineTaskConnections,
         storageClass="SourceCatalog",
         name="{fakesType}{coaddName}Diff_diaSrc_schema",
     )
-    # TODO DM-29965: Currently the AL likelihood image is not a separate data product
     subtractedExposure = pipeBase.connectionTypes.Output(
         doc="Output AL difference or Zogy proper difference image",
         dimensions=("instrument", "visit", "detector"),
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -176,21 +175,23 @@ class ImageDifferenceConfig(pipeBase.PipelineTaskConfig,
...
     doScaleTemplateVariance = pexConfig.Field(dtype=bool, default=False,
                                               doc="Scale variance of the template before PSF matching")
     doScaleDiffimVariance = pexConfig.Field(dtype=bool, default=True,
                                             doc="Scale variance of the diffim before PSF matching. "
                                                 "You may do either this or template variance scaling, "
                                                 "or neither. (Doing both is a waste of CPU.)")
-    useGaussianForPreConvolution = pexConfig.Field(dtype=bool, default=True,
-                                                   doc="Use a simple gaussian PSF model for pre-convolution "
-                                                       "(else use fit PSF)? Ignored if doPreConvolve false.")
+    useGaussianForPreConvolution = pexConfig.Field(
```
kherner, line 192: One thing to consider is whether it would be a good idea to raise a ValueError around here, similar to what's done below, if an invalid config is passed (e.g. using zogy and/or useScoreImageDetection is False, and this option is set to True). That said, there's nothing wrong with the code (it will just go back to using the exposure PSF and work fine) though it might lead to confusion down the line in that people are expecting it to be one way when it really is not. Granted, you've covered it in the doc string, so it's not like you're not informing people. Up to you, really.

kherner (commented):
See my comment around line 192.

PR #549, opened 2021-07-09
DM-31060: Clean up some log usage

timj:
* `self.log.warn` -> `self.log.warning`
* `log.x("message: %s" % msg)` -> `log.x("message: %s", msg)`


python/lsst/pipe/tasks/ingestCalibs.py
```diff
@@ -168,8 +168,8 @@ def fixSubsetValidity(self, conn, table, detectorData, validity):
         except Exception:
             det = " ".join("%s=%s" % (k, v) for k, v in zip(self.config.detector, detectorData))
             # Sqlite returns unicode strings, which cannot be passed through SWIG.
```
isullivan, line 170: Can we remove this comment, since we are no longer using SWIG?

isullivan (approved):
Looks good; I had one minor comment.

PR #551, opened 2021-07-19
DM-30202: Add option to do online coaddition for MEAN statistics.

erykoff:


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -995,6 +1008,81 @@ def assembleSubregion(self, coaddExposure, bbox, tempExpRefList, imageScalerList
...
+        """Assemble the coadd using the "online" method.
+
+        This method takes a running sum of images and weights to save memory.
+        It only works for MEAN statistics.
+
+        Parameters
+        ----------
+        coaddExposure : `lsst.afw.image.Exposure`
+            The target exposure for the coadd.
+        tempExpRefList : `list`
```
arunkannawadi, line 1022: Could you add the explicit type of the object these variables hold (although they are not present in `assembleSubregion`)?
https://developer.lsst.io/python/numpydoc.html#sequence-types

tests/test_assembleCoadd.py
```diff
@@ -284,6 +284,31 @@ def testDcrGen2Gen3Compatibility(self):
...
+        configOnline.validate()
+        assembleTaskOnline = MockInputMapAssembleCoaddTask(config=configOnline)
+
+        resultsOnline = assembleTaskOnline.runQuantum(self.skyInfo, dataRefList)
+        coaddOnline = resultsOnline.coaddExposure
+
+        self.assertFloatsAlmostEqual(coaddOnline.image.array,
+                                     coadd.image.array, rtol=1e-3)
+        self.assertFloatsAlmostEqual(coaddOnline.variance.array,
+                                     coadd.variance.array, rtol=1e-6)
```
arunkannawadi, line 309: Is this the maximum precision at which the images agree? I'd have expected it to be more precise.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -995,6 +1008,81 @@ def assembleSubregion(self, coaddExposure, bbox, tempExpRefList, imageScalerList
...
+        altMaskList : `list`
+            List of alternate masks to use rather than those stored with
+            tempExp, or None.  Each element is dict with keys = mask plane
+            name to which to add the spans.
+        statsCtrl : `lsst.afw.math.StatisticsControl`
+            Statistics control object for coadd
+        nImage : `lsst.afw.image.ImageU`, optional
+            Keeps track of exposure count for each pixel.
+        """
+        tempExpName = self.getTempExpDatasetName(self.warpType)
```
arunkannawadi, line 1037: The logging (at the debug level) that happens in `assembleSubregion` should happen here as well, and explicitly say it is using the online coaddition.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -828,13 +833,21 @@ def run(self, skyInfo, tempExpRefList, imageScalerList, weightList,
...
             try:
-                self.assembleSubregion(coaddExposure, subBBox, tempExpRefList, imageScalerList,
-                                       weightList, altMaskList, stats.flags, stats.ctrl,
-                                       nImage=nImage)
+                self.assembleOnlineMeanCoadd(coaddExposure, tempExpRefList, imageScalerList,
+                                             weightList, altMaskList, stats.ctrl,
+                                             nImage=nImage)
             except Exception as e:
-                self.log.fatal("Cannot compute coadd %s: %s", subBBox, e)
+                self.log.fatal("Cannot compute coadd %s", e)
```
arunkannawadi, line 842: The error message should indicate that it tried online coaddition. Do you think it makes sense to fall back to the 'offline' coaddition if online coaddition fails, or is it going to be futile anyway?

arunkannawadi (approved):
LGTM, Eli! Looking forward to using this in DESC work.

python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -828,13 +833,21 @@ def run(self, skyInfo, tempExpRefList, imageScalerList, weightList,
                                                  skyInfo.wcs,
                                                  coaddExposure.getInfo().getCoaddInputs().ccds)
 
-        for subBBox in self._subBBoxIter(skyInfo.bbox, subregionSize):
+        if self.config.doOnlineForMean and self.config.statistic == 'MEAN':
```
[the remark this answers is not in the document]
erykoff: I've changed it, but just so you know the single- or double- quoting of MEAN is all over the place in this file.

python/lsst/pipe/tasks/assembleCoadd.py
erykoff: This is left blank in all the docstrings ... and it could be the gen2 or gen3 types ... but I can add something.

erykoff: Or maybe not.  This is a much bigger kettle of fish, it has multiple names, and the pain point is the gen2/gen3 overlap which will be cleaned out at some point.

tests/test_assembleCoadd.py
erykoff: I _think_ that the difference is because the afw accumulator works with deltas (from the original mean) as opposed to actual summation, but that's not possible with the online version.  But if this level of precision is insufficient, then let me know and I can see if I can make it better.

python/lsst/pipe/tasks/assembleCoadd.py
erykoff: I think that's going to be futile, but if the user has configured things to use online coaddition I think it should do that.  Though the only real failures are somehow data not being there (which should be impossible) and that's going to be the same problem with the subregions.

PR #552, opened 2021-07-20
DM-30301: Switch lsst.log usage to logging

timj:


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -52,6 +52,8 @@
            "SafeClipAssembleCoaddTask", "SafeClipAssembleCoaddConfig",
            "CompareWarpAssembleCoaddTask", "CompareWarpAssembleCoaddConfig"]
 
+log = logging.getLogger(__name__.partition(".")[2])
```
ktlim, line 55: Slightly dangerous, as I think it fails catastrophically if the module name somehow doesn't contain `.`.

Maybe safer to do `__name__.replace("lsst.", "", 1)` ?

python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -39,6 +39,8 @@
 
 __all__ = ["MakeCoaddTempExpTask", "MakeWarpTask", "MakeWarpConfig"]
 
+log = logging.getLogger(__name__.partition(".")[2])
```
ktlim, line 42: See above.

ktlim (approved):

python/lsst/pipe/tasks/assembleCoadd.py
timj: We do this all over the place in ctrl_mpexec (and elsewhere) -- we don't do it in daf_butler because in daf_butler the `lsst.` is retained. I think we'd want to constrain the replacement to start of string as well.

A case could be made for a helper function to do this consistently in ctrl_mpexec and pipe_tasks but would that be pipe_base.task_logging? or lsst.daf.butler.core.logging that has it?

timj: I think by default I'm not going to change this. If ever this code moves out of `lsst.` module then I'm sure more code fixes will be needed and it's going to break instantly. It isn't going to break though until the code moves.

PR #554, opened 2021-07-21
DM-31163: Enable section for flags from forced_src

yalsayyad:
in the Object Table specification. Specify with forcedFlags header

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -551,6 +551,10 @@ class CompositeFunctor(Functor):
             for flag in translationDefinition['refFlags']:
                 funcs[cls.renameCol(flag, renameRules)] = Column(flag, dataset='ref')
 
+        if 'forcedFlags' in translationDefinition:
+            for flag in translationDefinition['forcedFlags']:
+                funcs[cls.renameCol(flag, renameRules)] = Column(flag, dataset='forced_src')
+
         if 'flags' in translationDefinition:
             for flag in translationDefinition['flags']:
                 funcs[cls.renameCol(flag, renameRules)] = Column(flag, dataset='meas')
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/functors.py
```diff
@@ -551,6 +551,10 @@ def from_yaml(cls, translationDefinition, **kwargs):
             for flag in translationDefinition['refFlags']:
                 funcs[cls.renameCol(flag, renameRules)] = Column(flag, dataset='ref')
 
+        if 'forcedFlags' in translationDefinition:
```
arunkannawadi, line 554: This should preferably occur after the block involving `flags`. That way, if a flag appears twice by mistaken, then the forced_src will be the one that in the object table, assuming that's what we'd want.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -455,12 +455,13 @@ class PostprocessAnalysis(object):
     _defaultFuncs = (('coord_ra', RAColumn()),
                      ('coord_dec', DecColumn()))
 
-    def __init__(self, parq, functors, filt=None, flags=None, refFlags=None):
+    def __init__(self, parq, functors, filt=None, flags=None, refFlags=None, forcedFlags=None):
```
arunkannawadi, line 458: The `forcedFlags` argument is not documented yet. Similarly, in line 605 it should be mentioned that `forcedFlags` will also be expanded per band.

arunkannawadi (approved):

arunkannawadi:
Approving it for now so as to not block the merge before tonight, but changes are requested.

python/lsst/pipe/tasks/functors.py
yalsayyad: ...and I do want the regular meas flags to be the default.

yalsayyad:
The "changes requested" button is usually for big refactoring changes that require a second review because the rework is new code.

PR #555, opened 2021-07-26
DM-23786: Add bright star subtraction task

MorganSchmitz:
[Jira ticket](https://jira.lsstcorp.org/browse/DM-23786); [Jenkins run #34670](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/34670/pipeline)

### DIFF
--- a/python/lsst/pipe/tasks/subtractBrightStars.py
+++ b/python/lsst/pipe/tasks/subtractBrightStars.py
@@ -39,19 +39,19 @@ from lsst import geom
 
 class SubtractBrightStarsConnections(pipeBase.PipelineTaskConnections,
                                      dimensions=("instrument", "visit", "detector"),
-                                     defaultTemplates={"outputExposureName": "subtracted",
+                                     defaultTemplates={"outputExposureName": "brightStar_subtracted",
                                                        "outputBackgroundName": "brightStars"}):
     inputExposure = cT.Input(
-        doc="Input exposure from which to subtract bright star stamps",
+        doc="Input exposure from which to subtract bright star stamps.",
         name="calexp",
         storageClass="ExposureF",
-        dimensions=("visit", "detector")
+        dimensions=("visit", "detector",),
     )
     inputBrightStarStamps = cT.Input(
         doc="Set of preprocessed postage stamps, each centered on a single bright star.",
         name="brightStarStamps",
         storageClass="BrightStarStamps",
-        dimensions=("visit", "detector")
+        dimensions=("visit", "detector",),
     )
     inputExtendedPsf = cT.Input(
         doc="Extended PSF model.",
@@ -60,22 +60,22 @@ class SubtractBrightStarsConnections(pipeBase.PipelineTaskConnections,
         dimensions=("band",),
     )
     skyCorr = cT.Input(
-        doc="Input Sky Correction to be subtracted from the calexp if doApplySkyCorr=True",
+        doc="Input Sky Correction to be subtracted from the calexp if ``doApplySkyCorr``=True.",
         name="skyCorr",
         storageClass="Background",
-        dimensions=("instrument", "visit", "detector")
+        dimensions=("instrument", "visit", "detector",),
     )
     outputExposure = cT.Output(
         doc="Exposure with bright stars subtracted.",
         name="{outputExposureName}_calexp",
         storageClass="ExposureF",
-        dimensions=("visit", "detector"),
+        dimensions=("visit", "detector",),
     )
     outputBackgroundExposure = cT.Output(
         doc="Exposure containing only the modelled bright stars.",
         name="{outputBackgroundName}_calexp_background",
         storageClass="ExposureF",
-        dimensions=("visit", "detector"),
+        dimensions=("visit", "detector",),
     )
 
     def __init__(self, *, config=None):
@@ -112,12 +112,14 @@ class SubtractBrightStarsConfig(pipeBase.PipelineTaskConfig,
             "lanczos3": "Lanczos kernel of order 3",
             "lanczos4": "Lanczos kernel of order 4",
             "lanczos5": "Lanczos kernel of order 5",
+            "lanczos6": "Lanczos kernel of order 6",
+            "lanczos7": "Lanczos kernel of order 7",
         }
     )
     scalingType = pexConfig.ChoiceField(
         dtype=str,
         doc="How the model should be scaled to each bright star; implemented options are "
-            "annularFlux to reuse the annular flux of each stamp, or leastSquare to perform "
+            "`annularFlux` to reuse the annular flux of each stamp, or `leastSquare` to perform "
             "least square fitting on each pixel with no bad mask plane set.",
         default="leastSquare",
         allowed={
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,325 @@
...
+        -------
+        subtractorExp : `afwImage.ExposureF`
+            An Exposure containing every bright star's profile; its image can
+            then be subtracted from the input exposure.
+        invImages : `list` [`afwImage.MaskedImageF`]
+            A list of small images ("stamps") containing the model, each scaled
+            to its corresponding input bright star.
+        """
+        inputExpBBox = inputExposure.getBBox()
+        if self.config.doApplySkyCorr:
```
arunkannawadi, line 273: Do we have to check if `skyCorr` is not None here (default), or can `applySkyCorr` handle that case?

arunkannawadi (changes requested):
I've left a ton of comments, but almost all of them are trivial. My major complaint is that there is no test code to run this task. I noticed at least one place where running this task in accordance with the documentation would have still crashed it, so some unit testing is necessary before merging this.

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,325 @@
...
+        self.statsControl, self.statsFlag = None, None
+
+    def _setUpStatistics(self, exampleMask):
+        """Configure statistics control and flag, for use if scalingType is
+        ``"leastSquare"``.
+        """
+        if self.config.scalingType == 'leastSquare':
+            self.statsControl = afwMath.StatisticsControl()
+            # set mask planes to be ignored
+            andMask = reduce(ior, (exampleMask.getPlaneBitMask(bm) for bm in self.config.badMaskPlanes))
```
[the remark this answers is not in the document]
MorganSchmitz: Yep, that's why I called it that. Note there's precedent [here](https://github.com/lsst/meas_algorithms/blob/master/python/lsst/meas/algorithms/brightStarStamps.py#L139) - not that that's a good reason to keep using it :)

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,359 @@
...
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        if not config.doApplySkyCorr:
+            self.inputs.remove("skyCorr")
+
+
+class SubtractBrightStarsConfig(PipelineTaskConfig, pipelineConnections=SubtractBrightStarsConnections):
+    """Configuration parameters for SubtractBrightStarsTask"""
+
+    doWriteSubtractor = Field(
```
arunkannawadi, line 114: Either in this ticket, or in the refactoring ticket that should follow soon, it'd be good to make it as `Field[bool]` so type annotations can work.

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,359 @@
...
+        doc="Mask planes that, if set, lead to associated pixels not being included in the computation of "
+        "the scaling factor (`BAD` should always be included). Ignored if scalingType is `annularFlux`, "
+        "as the stamps are expected to already be normalized.",
+        # Note that `BAD` should always be included, as secondary detected
+        # sources (i.e., detected sources other than the primary source of
+        # interest) also get set to `BAD`.
+        default=("BAD", "CR", "CROSSTALK", "EDGE", "NO_DATA", "SAT", "SUSPECT", "UNMASKEDNAN"),
+    )
+    doApplySkyCorr = Field(
+        dtype=bool, doc="Apply full focal plane sky correction before extracting stars?", default=True
```
arunkannawadi, line 164: Is this a black thing? I'd like it personally if these were in their own individual lines. One way to make black wrap it would be to give a trailing comma, as is the case in previous Fields.

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,359 @@
...
+            raise AttributeError(
+                f'Unknown scalingType "{self.config.scalingType}"; implemented options '
+                'are "annularFlux" and "leastSquare".'
+            )
+        if inPlace:
+            model.image *= scalingFactor
+        return scalingFactor
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        """Run a single quantum of the task."""
```
arunkannawadi, line 269: You can remove this docstring and replace it with a comment "#Docstring inherited." If it didn't have a docstring, it'd inherit it automatically from the parent `PipelineTask` which is probably sufficient here.

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,359 @@
...
+        output = Struct(outputExposure=outputExposure, outputBackgroundExposure=outputBackgroundExposure)
+        butlerQC.put(output, outputRefs)
+
+    def run(self, inputExposure, inputBrightStarStamps, inputExtendedPsf, dataId, skyCorr=None):
+        """Iterate over all bright stars in an exposure to scale the extended
+        PSF model before subtracting bright stars.
+
+        Parameters
+        ----------
+        inputExposure : `afwImage.exposure.exposure.ExposureF`
```
arunkannawadi, line 288:
```suggestion
        inputExposure : `~lsst.afw.image.exposure.exposure.ExposureF`
```

arunkannawadi: Similarly in the `Returns` section as well.

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,325 @@
...
+                          dataId)
+            self.applySkyCorr(inputExposure, skyCorr)
+        # Create an empty image the size of the exposure
+        # TODO: DM-31085 set mask planes
+        subtractorExp = afwImage.ExposureF(bbox=inputExposure.getBBox())
+        subtractor = subtractorExp.maskedImage
+        # make a copy of the input model
+        model = inputExtendedPsf(dataId["detector"]).clone()
+        modelStampSize = model.getDimensions()
+        inv90Rots = 4 - inputBrightStarStamps.nb90Rots
```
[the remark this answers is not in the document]
arunkannawadi: I'm not sure if there's any check to ensure `nb90Rots` is an integer between 0-3. So this might be better off as 
```python
inv90Rots = 4 - inputBrightStarStamps.nb90Rots%4
```

arunkannawadi (commented):
I want to give this another pass, so not marking it as Approved for now. Happy to work on unit tests for this.

python/lsst/pipe/tasks/subtractBrightStars.py
leeskelvin: Type annotations added to all `Config` options (including `ChoiceField`, which I think works now?)

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -0,0 +1,359 @@
...
+            # is E[XY]/E[XX].
+            xy = starIm.clone()
+            xy.image.array *= model.image.array
+            xx = starIm.clone()
+            xx.image.array = model.image.array**2
+            # Compute the least squares scaling factor.
+            xySum = makeStatistics(xy, self.statsFlag, self.statsControl).getValue()
+            xxSum = makeStatistics(xx, self.statsFlag, self.statsControl).getValue()
+            scalingFactor = xySum / xxSum if xxSum else 0
+        else:
```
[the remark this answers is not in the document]
leeskelvin: Removed. I'd like to replace this with a `match` during a refactor, as I think that's more readable, but leaving as-is for now sans the final `else` block.

arunkannawadi (approved):

PR #556, opened 2021-07-26
DM-31188: Fix LoadReferenceCatalogTask so it doesn't temporarily clobber its own config

erykoff:


bechtol (approved):
Thanks for this update. Looks good.

I have verified that this version can be used to load reference catalogs with proper motions applied (Gaia) and with color terms applied (PS1) when config is frozen. Glad to see that freezing the config is also now included in the unit tests.

PR #563, opened 2021-08-13
DM-30465 Implement decorrelation afterburner for maximum likelihood images

kgabor:


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -970,15 +970,26 @@ def run(self, exposure=None, selectSources=None, templateExposure=None, template
                 # useScoreImageDetection=True and doDecorrelation=True is not allowed in the config
```
isullivan, line 970: This comment should be removed

python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -970,15 +970,26 @@ def run(self, exposure=None, selectSources=None, templateExposure=None, template
...
+                        scoreExposure = decorrResult.correctedExposure
+                    else:
+                        decorrResult = self.decorrelate.run(exposureOrig, subtractRes.warpedExposure,
+                                                            subtractedExposure,
+                                                            subtractRes.psfMatchingKernel,
+                                                            spatiallyVarying=self.config.doSpatiallyVarying,
+                                                            preConvKernel=None,
+                                                            templateMatched=self.config.convolveTemplate,
+                                                            preConvMode=False)
+                        subtractedExposure = decorrResult.correctedExposure
```
isullivan, line 992: I think we want to decorrelate the subtracted exposure, even if we use the score image for detection.

isullivan (approved):
Please move decorrelation of the subtracted exposure out of the else statement. I believe we still want to decorrelate it even if we use the score image for detection.

PR #566, opened 2021-08-17
DM-31359: Add abs to greater-than statement in maskStreaks

cmsaunders:


czwa (approved):
This looks reasonable to me.

PR #568, opened 2021-08-24
DM-31507 Modify data-structures to aid ref counting

natelust:
It is desirable that variables get cleaned up when they go out of
scope, meaning that their ref count must go to zero. Some objects
kept closures or circular references which means they must be
cleaned by the garbage collector which can happen at unpredictable
times. Make these objects able to be dropped with reference counting.


python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -137,6 +139,10 @@ def __init__(self, config: Config, field: ConfigurableActionStructField,
             for k, v in value.items():
                 setattr(self, k, v)
 
+    @property
+    def _config(self) -> Optional[Config]:
+        return self._config_()
```
andy-slac, line 144: My general worry about this pattern is that previously `_config` was never `None`, while after this change it is possible (may be improbable) that weakref returns `None` if original `Config` is gone. Could you make it return `Config` instead of `Optional[Config]` and raise an exception if weakref object is deleted? This applies to other packages where you use the same method, I'm not going to comment on it there.

andy-slac (approved):
Looks OK, one comment may be worth to think about.

PR #570, opened 2021-09-08
DM-29628: Handle alignment of source injection from FITS files

jmeyers314:


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -263,6 +263,22 @@ class InsertFakesConfig(PipelineTaskConfig,
         default="sourceType",
     )
 
+    fits_alignment = pexConfig.ChoiceField(
+        doc="How should injections from FITS files be aligned?",
+        dtype=str,
+        allowed={
+            "wcs": (
+                "Local WCS from the input FITS image will align with local WCS "
```
leeskelvin, line 271: I think this would read better if it was written along the lines of "Local WCS from the input FITS image will be used to align it with the local WCS in the target image." maybe?

jmeyers314: Sure, here's a proposed rewrite with more detail for both "wcs" and "pixel" options:

```
            "wcs": (
                "Input image will be transformed such that the local WCS in "
                "the FITS header matches the local WCS in the target image."
                "I.e., North, East, and angular distances in the input image "
                "will match North, East, and angular distances in the target "
                "image."
            ),
            "pixel": (
                "Input image will _not_ be transformed.  Up, right, and pixel "
                "distances in the input image will match up, right and pixel "
                "distances in the target image."
            )
```

PR #573, opened 2021-09-13
DM-30927: Set calexp WCS to None for failed astrometric fit

laurenam:


python/lsst/pipe/tasks/makeWarp.py
```diff
@@ -576,6 +592,7 @@ def prepareCalibratedExposures(self, calExpList, backgroundList=None, skyCorrLis
                 calexp.maskedImage -= skyCorr.getImage()
 
             indices.append(index)
+            calExpList[index] = calexp
```
erykoff, line 595: This makes me nervous ... this is an update of the list as you enumerate over it?

But more importantly, I'm not sure why this API has to change at all.  Why can't we still read the calexps in the runQuantum and just filter here?  This might require some discussion.

python/lsst/pipe/tasks/propagateSourceFlags.py
```diff
@@ -189,6 +189,10 @@ def run(self, coadd_object_cat, ccd_inputs,
                     for row in ccd_inputs[ccd_inputs["visit"] == visit]:
                         detector = row["ccd"]
                         wcs = row.getWcs()
+                        if wcs is None:
+                            self.log.info("No WCS for visit %d detector %d, so can't match sources to "
+                                          "propagate flags.  Skipping...", visit, detector)
+                            continue
```
erykoff, line 195: This is fine as a belt-and-suspenders, but I don't know how a None wcs can possibly make it into the coadd input list.

erykoff (changes requested):
I think we need to discuss the `makeWarp` changes.  I think it can be done much more simply, but I may be missing something.

python/lsst/pipe/tasks/makeWarp.py
laurenam: Ok, so I think I know where the confusion lies (and am not fully convinced what I'm doing is ideal, but it seemed the best option at the time).  The main difference is that I'm now passing `calExpList` to `prepareCalibratedExposures()` as a `list` of  `DeferredDatasetHandle`s.  The reason for the delayed read is to potentially avoid unnecessarily reading in `calexps` that will never be used (via the `if externalSkyWcsCatalog is None and wcsList[index] is None:` clause at the beginning of the loop...and this is also how a `calexp` with a failed SFM WCS could still pick up the external calibration, if present).  Also, this way the "within patch" assessment is made on the final WCS (externally calibrated, if relevant), not just the one from SFM (I _think_ this is slightly preferred, although probably won't play a huge role in most circumstances).  Since the `calExpList` `list` was of `DeferredDatasetHandle`s, I update it here to contain the (now re-calibrated) actual `calexp`  `Exposure` object.

Does that make any sense?!

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -167,12 +170,12 @@ def getValidImageCorners(self, imageWcs, imageBox, patchPoly, dataId=None):
         except (pexExceptions.DomainError, pexExceptions.RuntimeError) as e:
             # Protecting ourselves from awful Wcs solutions in input images
             self.log.debug("WCS error in testing calexp %s (%s): deselecting", dataId, e)
-            return
+            return None
```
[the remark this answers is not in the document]
laurenam: It is, but I was following PEP8's:

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None...

and there is a `return imageCorners` in this function below (and, looking closer, I wonder if a final `return None` at the end of this function, or at least in an `else` to the `if patchPoly.intersects(imagePoly):` clause is needed?)

python/lsst/pipe/tasks/makeWarp.py
```diff
@@ -530,40 +561,75 @@ def filterInputs(self, indices, inputs):
...
+        """Deprecated function.
+
+        Please use _prepareCalibratedExposure(), which this delegates to and
+        noting its slightly updated API, instead.
+        """
+        # Read in all calexps.
+        calExpList = [ref.get() for ref in calExpList]
+        # Populate wcsList as required by new underscored version of function.
+        wcsList = [calexp.getWcs() for calexp in calExpList]
+        print("In prepareCalibratedExposures, wcsList = ", wcsList)
```
erykoff, line 580: Please remove!

laurenam: Whoops!  Thanks for the catch!

python/lsst/pipe/tasks/makeWarp.py
```diff
@@ -573,12 +638,20 @@ def prepareCalibratedExposures(self, calExpList, backgroundList=None, skyCorrLis
         for index, (calexp, background, skyCorr) in enumerate(zip(calExpList,
                                                                   backgroundList,
                                                                   skyCorrList)):
```
erykoff, line 640: I think this should be zipping `wcsList` as well, instead of doing `wcsList[index]`, for consistency/clarity if nothing else (unless there's a reason I don't see for that).

laurenam: It's really only the one instance below in 
`if externalSkyWcsCatalog is None and wcsList[index] is None:`
that would get updated as I still need to use the `wcsList[index]` to assign the new external calibrations, but if you think that's still worth it, I'm happy to add it to the zip list!

PR #574, opened 2021-09-16
DM-31313: Add HtmIndex20 functor

yalsayyad:


python/lsst/pipe/tasks/functors.py
```diff
@@ -709,6 +712,28 @@ def __call__(self, catalog, **kwargs):
...
 
 
+class HtmIndex20(Functor):
+    """Compute the level 20 HtmIndex for the catalog.
+    """
+    columns = ['coord_ra', 'coord_dec']
+    name = "Htm20"
+    htmLevel = 20
+
+    def __init__(self, **kwargs):
```
yalsayyad, line 722: We will maybe one day need to compute `HtmIndex20` on other column names (or in the case of a Multi-level parquet  on  `dataset` other than `ref`), but hardcoding is fine for now.

yalsayyad (commented):
Also remove the "debug unittest" part of the commit message.

PR #575, opened 2021-09-21
DM-31765: Transform forced sources on diaObjects during DRP

yalsayyad:
- Generalize the write|transform|consolidateForcedSourceTasks to
work regardless of reference catalog used to seed the forced phot.
- Make the error message more clear if a requested column is missing.


cmsaunders (approved):
These changes look good.

PR #576, opened 2021-09-23
DM-31491: Add default types to allow fakes propagation

sr525:


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -81,6 +83,13 @@ def _add_fake_sources(exposure, objects, calibFluxRadius=12.0, logger=None):
         mat = wcs.linearizePixelToSky(spt, geom.arcseconds).getMatrix()
         gsWCS = galsim.JacobianWCS(mat[0, 0], mat[0, 1], mat[1, 0], mat[1, 1])
 
+        # This check is here because sometimes the WCS
+        # is multivalued and objects that should not be
+        # were being included.
+        gsPixScale = np.sqrt(gsWCS.pixelArea())
+        if gsPixScale < pixScale/2 or gsPixScale > pixScale*2:
```
leeskelvin, line 90: Is a gsPixScale offset of a factor of 2 sufficient to always capture this multivalued WCS issue, or will you still expect some level of failure? If sufficient, then that's fine. If it might continue to be an issue however, then perhaps this factor should be configurable?

sr525: This was sufficient in testing, I think if it turns out not to be we need to do something more intelligent than this check. In which case I would like it to crash so we can figure out a better solution.

PR #577, opened 2021-09-24
DM-31217: Add ConsolidateHealSparsePropertyMapsTask

erykoff:


python/lsst/pipe/tasks/healSparseMapping.py
```diff
@@ -684,3 +687,232 @@ def _compute_nside_coverage_tract(self, tract_info):
...
+            doc=f"Minumum-value map of {name}",
+            name=f"{{coaddName}}Coadd_{name}_consolidated_map_sum",
+            storageClass="HealSparseMap",
+            dimensions=("skymap", "band"),
+        )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+
+        # Remove connections from those that are not configured
```
taranu, line 779: Why would there be unconfigured connections? Is this something a user should be warned about?

python/lsst/pipe/tasks/healSparseMapping.py
```diff
@@ -684,3 +687,232 @@ def _compute_nside_coverage_tract(self, tract_info):
...
+        self.property_maps["sky_noise"].do_weighted_mean = True
+        self.property_maps["sky_background"].do_weighted_mean = True
+        self.property_maps["dcr_dra"].do_weighted_mean = True
+        self.property_maps["dcr_ddec"].do_weighted_mean = True
+        self.property_maps["dcr_e1"].do_weighted_mean = True
+        self.property_maps["dcr_e2"].do_weighted_mean = True
+
+
+class ConsolidateHealSparsePropertyMapTask(pipeBase.PipelineTask):
+    """Task to consolidate HealSparse property maps."""
```
taranu, line 842: It would be helpful to document somewhere what exactly is being consolidated as it's not obvious to me.

python/lsst/pipe/tasks/healSparseMapping.py
```diff
@@ -684,3 +687,232 @@ def _compute_nside_coverage_tract(self, tract_info):
...
+        doc="Input definition of geometry/bbox and projection/wcs for coadded exposures",
+        name=BaseSkyMap.SKYMAP_DATASET_TYPE_NAME,
+        storageClass="SkyMap",
+        dimensions=("skymap",),
+    )
+
+    # Create input/output connections for all possible maps defined in the
+    # registry.
+    for name in BasePropertyMap.registry:
+        vars()[f"{name}_map_min"] = pipeBase.connectionTypes.Input(
```
taranu, line 705: Is this the preferred pattern for batch assignment of connections? It seems sensible on second thought but was unexpected to me at first.

taranu (commented):
Minor clarifications requested.

python/lsst/pipe/tasks/healSparseMapping.py
erykoff: I've improved the comments here and when I used the same pattern for the tract-based maps.  But the idea is that we have to create all possible connections that are in the registry, and then remove those that weren't configured to run to leave the subset of connections that we want to use.

python/lsst/pipe/tasks/healSparseMapping.py
erykoff: This is the only way that I know of to programatically make connections which must be class instance variables.  I've added additional comments, including a link to the magic stack overflow which told me how to do this: https://stackoverflow.com/questions/2519807/setting-a-class-attribute-with-a-given-name-in-python-while-defining-the-class

python/lsst/pipe/tasks/healSparseMapping.py
erykoff: I am expanding this, and adding rst task documentation which I had forgotten to add.

taranu (approved):

PR #579, opened 2021-09-28
DM-31825: Implement RFC-807 on  DRP DataFrame consistency

yalsayyad:
* all dataframes have primary keys as index only
* all final parquet table dataset names start with lowercase and end with `Table_visit` or `Table_tract` depending if its a visit or tract level dataset 
* object table columns look like `g_apFlux` instead of `gApFlux`
* All per-patch datasets are consolidated into a per-tract dataset

I promise to squash these before merging

### DIFF
--- a/pipelines/DRP.yaml
+++ b/pipelines/DRP.yaml
@@ -64,6 +64,7 @@ tasks:
   transformDiaSourceCat:
     class: lsst.ap.association.TransformDiaSourceCatalogTask
     config:
+      doPackFlags: false
       connections.coaddName: goodSeeing
       connections.diaSourceSchema: goodSeeingDiff_diaSrc_schema
       connections.diaSourceCat: goodSeeingDiff_diaSrc
--- a/pipelines/DRP.yaml
+++ b/pipelines/DRP.yaml
@@ -88,6 +88,16 @@ tasks:
       connections.assocDiaSourceTable: goodSeeingDiff_assocDiaSrcTable
       connections.diaObjectTable: goodSeeingDiff_diaObjTable
       connections.fullDiaObjectTable: goodSeeingDiff_fullDiaObjTable
+  consolidateAssocDiaSourceTable:
+    class: lsst.pipe.tasks.postprocess.ConsolidateTractTask
+    config:
+      connections.inputCatalogs: goodSeeingDiff_assocDiaSrcTable
+      connections.outputCatalog: diaSourceTable_tract
+  consolidateFullDiaObjectTable:
+    class: lsst.pipe.tasks.postprocess.ConsolidateTractTask
+    config:
+      connections.inputCatalogs: goodSeeingDiff_fullDiaObjTable
+      connections.outputCatalog: diaObjectTable_tract
   forcedPhotDiffim:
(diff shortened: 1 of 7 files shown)


sr525 (approved):

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1468,6 +1469,7 @@ def run(self, visitSummaries):
             visitEntry = {}
             visitEntry["visitId"] = visitRow['visit']
             visitEntry["filterName"] = visitRow['physical_filter']
+            visitEntry["band"] = visitRow['band']
```
sr525, line 1498: Mix of " and '

pipelines/DRP.yaml
```diff
@@ -90,6 +90,16 @@ tasks:
       connections.assocDiaSourceTable: goodSeeingDiff_assocDiaSrcTable
       connections.diaObjectTable: goodSeeingDiff_diaObjTable
       connections.fullDiaObjectTable: goodSeeingDiff_fullDiaObjTable
+  consolidateAssocDiaSourceTable:
+    class: lsst.pipe.tasks.postprocess.ConsolidateTractTask
+    config:
+      connections.inputCatalogs: goodSeeingDiff_assocDiaSrcTable
+      connections.outputCatalog: diaSourceTable_tract
+  consolidateFullDiaObjectTable:
```
sr525, line 99: Just to double check that these should be the same class.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -540,7 +540,15 @@ class TransformCatalogBaseConfig(pipeBase.PipelineTaskConfig,
                                  pipelineConnections=TransformCatalogBaseConnections):
     functorFile = pexConfig.Field(
         dtype=str,
-        doc='Path to YAML file specifying functors to be computed',
+        doc='Path to YAML file specifying Science DataModel functors to use '
```
sr525, line 543: DataModel or Data Model?

python/lsst/pipe/tasks/simpleAssociation.py
```diff
@@ -189,7 +189,9 @@ def run(self, diaSources, tractPatchId, skymapBits):
         diaSources.set_index("diaSourceId", inplace=True, verify_integrity=True)
 
         diaObjects = pd.DataFrame(data=diaObjectCat)
-        diaObjects.set_index("diaObjectId", inplace=True,  verify_integrity=True)
+        diaSources.reset_index(inplace=True)
+        if 'diaObjectId' in diaObjects.columns:
+            diaObjects.set_index("diaObjectId", inplace=True, verify_integrity=True)
```
sr525, line 194: Mix of ' and "

python/lsst/pipe/tasks/postprocess.py
yalsayyad: Keeping it consistent with what's there.

pipelines/DRP.yaml
yalsayyad: Feature not a bug :)

PR #582, opened 2021-10-08
DM-32065: Skip gen2 test_read_CuratedCalibs in pipe_tasks

leeskelvin:


tests/test_read_CuratedCalibs.py
```diff
@@ -33,12 +33,14 @@
 import lsst.daf.persistence as dafPersist
 
 ROOT = os.path.join(getPackageDir('obs_base'), 'tests')
```
timj, line 35: To prevent the duplication of the test how about adding:

```python
if not os.path.exists(ROOT):
    ROOT = None
```
and then the `skipIf` is `@unittest.skipIf(ROOT is None, ROOTSkipMsg)`
?

My try version was being a bit more defensive around `obs_base` not being available but you are right that if `obs_base` is not setup then everything is going to break anyhow.

PR #583, opened 2021-10-08
DM-31801: Use lsst prefix for loggers

timj:


tests/test_isPrimaryFlag.py
```diff
@@ -160,7 +160,7 @@ def setUp(self):
         self.charImResults = charImTask.run(self.exposure)
 
         # set log level so that warnings do not display
-        logging.getLogger("calibrate").setLevel(logging.ERROR)
+        logging.getLogger("lsst.calibrate").setLevel(logging.ERROR)
```
arunkannawadi, line 163: The logger name doesn't make it obvious that this is arising from the tests. I guess since it doesn't have `lsst.pipe.tasks` prefix, it may be linked to the unit test?

arunkannawadi (approved):
Just a minor comment about a plain `lsst` prefix in unit tests.

timj:
The `lsst.calibrate` is there because for the calibrate `Task` the logger is given that name.

PR #597, opened 2021-10-29
DM-32402: Update method to merge configs that allows registries.

erykoff:


TallJimbo (approved):

arunkannawadi (approved):
Just left a few minor suggestions, but the code as it stands should still work fine.

PR #600, opened 2021-11-09
DM-30831: DcrAssembleCoaddConnections should inherit from AssembleCoaddConnections

isullivan:


python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -97,8 +85,12 @@ class DcrAssembleCoaddConnections(pipeBase.PipelineTaskConnections,
 
     def __init__(self, *, config=None):
         super().__init__(config=config)
+        self.outputs.remove("coaddExposure")
```
morriscb, line 88: I assume these are part of AssembleCoaddConnections but this method does not output them, correct? If that's the case please add a comment here to the explaining why these two outputs are removed.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
```diff
@@ -45,35 +45,23 @@
 __all__ = ["DcrAssembleCoaddConnections", "DcrAssembleCoaddTask", "DcrAssembleCoaddConfig"]
 
 
-class DcrAssembleCoaddConnections(pipeBase.PipelineTaskConnections,
+class DcrAssembleCoaddConnections(AssembleCoaddConnections,
                                   dimensions=("tract", "patch", "band", "skymap"),
-                                  defaultTemplates={"inputCoaddName": "deep",
+                                  defaultTemplates={"inputWarpName": "deep",
```
morriscb, line 50: Is the inputWarpName template here in case you want to have different warps and coadds? Is that a possible or expected mode of operation?

morriscb (approved):
Looks fine in terms of changes. I do have a few questions and one request for an comment to be added to the code.

python/lsst/pipe/tasks/dcrAssembleCoadd.py
isullivan: Yes, these will be different in general. I'll add a comment.

PR #605, opened 2021-11-19
DM-31060: Log usage fixups

timj:
This is related to #549 but fixes some regressions.

### DIFF
--- a/python/lsst/pipe/tasks/imageDifference.py
+++ b/python/lsst/pipe/tasks/imageDifference.py
@@ -750,22 +750,22 @@ class ImageDifferenceTask(pipeBase.CmdLineTask, pipeBase.PipelineTask):
                 # TODO: DM-22762 This functional block should be moved into its own method
                 preConvPsf = None
                 if self.config.useScoreImageDetection:
-                    self.log.warn("AL likelihood image: pre-convolution of PSF is not implemented.")
+                    self.log.warning("AL likelihood image: pre-convolution of PSF is not implemented.")
                     convControl = afwMath.ConvolutionControl()
                     # cannot convolve in place, so need a new image anyway
                     srcMI = exposure.maskedImage
                     exposure = exposure.clone()  # New deep copy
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/imageDifference.py
```diff
@@ -750,22 +750,22 @@ def run(self, exposure=None, selectSources=None, templateExposure=None, template
...
                     convControl = afwMath.ConvolutionControl()
                     # cannot convolve in place, so need a new image anyway
                     srcMI = exposure.maskedImage
                     exposure = exposure.clone()  # New deep copy
                     srcPsf = sciencePsf
                     if self.config.useGaussianForPreConvolution:
-                        self.log.infof(
-                            "AL likelihood image: Using Gaussian (sigma={:.2f}) PSF estimation "
+                        self.log.info(
+                            "AL likelihood image: Using Gaussian (sigma=%.2f) PSF estimation "
```
leeskelvin, line 761: As with the other loggers, should this end with a full stop?

timj: I suppose it can. I hadn't looked at the content of the message or even noticed that everything else in this file had a full stop. Our use of periods at the end of log messages seems like a bit of a lottery.

timj: Actually, even in this file there are plenty of log messages that don't have the period so I don't think I want to look under that rock on this PR in case I end up making many changes throughout pipe_tasks.

leeskelvin (approved):

python/lsst/pipe/tasks/imageDifference.py
leeskelvin: That's a fair point - I only had in my scope the three loggers immediately relevant to this PR when I made the comment above. Looking at the entire script, I agree with you that it's probably best to avoid attempting any standardization here.

PR #608, opened 2021-11-29
DM-32714: Add backwards compatibility psf selection.

erykoff:


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -709,6 +716,11 @@ def __init__(self, *, config=None):
         # TODO DM-28769: add connection per selectImages connections
         if config.select.target != lsst.pipe.tasks.selectImages.PsfWcsSelectImagesTask:
             self.inputs.remove("visitSummary")
+            self.inputs.remove("srcList")
+        elif config.select.target == lsst.pipe.tasks.selectImages.PsfWcsSelectImagesTask:
```
arunkannawadi, line 720:
```suggestion
        elif not config.select.doLegacyStarSelectionComputation:
```

arunkannawadi: I don't see the need for an `elif` and another `if`.

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -315,6 +315,13 @@ class PsfWcsSelectImagesConfig(pipeBase.PipelineTaskConfig,
         deprecated=('This field has been moved to ComputeExposureSummaryStatsTask and '
                     'will be removed after v24.')
     )
+    doLegacyStarSelectionComputation = pexConfig.Field(
+        doc="Perform the legacy star selection computations (for backwards compatibility)",
+        dtype=bool,
+        default=False,
```
arunkannawadi, line 321: About the default value, should this be `True` by default making the users of `visitSummaryTable` use this config?

arunkannawadi (approved):
LGTM! Did a quick comparison with DM-32649 changes to get some context.

python/lsst/pipe/tasks/selectImages.py
erykoff: So this should only be set if you are in the case that you are running new coadd code on old calexp data, which I assume is not the norm.
Furthermore, this is an odd case (and you can complain!) that this is a new deprecated function (!!!) but in this way when somebody overrides it they will get a warning, and that will let them know that they are running in a specific backwards compatibility mode.

arunkannawadi: Fair enough!

PR #610, opened 2021-11-30
DM-32714-v23: Fix backwards compatibility broken by DM-32649

erykoff:


arunkannawadi (approved):
LGTM! This looks consistent with #608

PR #621, opened 2022-01-10
DM-27013: Change CalibrateTask refcat defaults to Gaia DR2 for astrometry and PS1 for photometry

parejkoj:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -317,6 +317,13 @@ def setDefaults(self):
         self.postCalibrationMeasurement.doReplaceWithNoise = False
         for key in self.postCalibrationMeasurement.slots:
             setattr(self.postCalibrationMeasurement.slots, key, None)
+        self.astromRefObjLoader.anyFilterMapsToThis = "phot_g_mean"
+        # The photoRefCat connection is the name to use for the colorterms.
+        self.photoCal.photoCatName = self.connections.photoRefCat
+
+        # NOTE: these two lines are for gen2, and are only here for compatibility.
+        self.astromRefObjLoader.ref_dataset_name = "gaia_dr2_20200414"
+        self.photoRefObjLoader.ref_dataset_name = "ps1_pv3_3pi_20170110"
```
leeskelvin, line 326: Do we still need to support gen2 functionality?

parejkoj: Yes, until the gen2 refcat code is completely removed we do need to support this (because there are consistency checks in various places that will remain until after that removal).

PR #622, opened 2022-01-20
DM-33327: insertOnlyStars option and option for adding fakes matched to visit

surhudm:
Addresses https://jira.lsstcorp.org/browse/DM-33327

1) Adds an insertOnlyStars option
2) The pixScale is checked at the center of the calibrated exposure than at the reference center of the WCS. This was causing many fake objects to not get inserted in CCDs near the edge of the mosaic.
3) Option to add fake objects based on their visit.


python/lsst/pipe/tasks/matchFakes.py
```diff
@@ -211,6 +221,28 @@ def composeFakeCat(self, fakeCats, skyMap):
...
+                    The exposure to add the fake sources to
+
+        Returns
+        -------
+        movingFakeCat : `pandas.DataFrame`
+            All fakes that satisfy conditions specified by matchExposureField
+        """
+        try:
+            selected = exposure.getInfo().getVisitInfo().getId() == fakeCat["visit"]
+        except Exception as e:
```
sr525, line 241: We probably shouldn't be catching generic exceptions.

sr525: Are there multiple typical exceptions or one main failure mode?

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -524,6 +533,29 @@ def composeFakeCat(self, fakeCats, skyMap):
...
+                    The exposure to add the fake sources to
+
+        Returns
+        -------
+        movingFakeCat : `pandas.DataFrame`
+            All fakes that satisfy conditions specified by matchExposureField
+        """
+        try:
+            selected = exposure.getInfo().getVisitInfo().getId() == fakeCat["visit"]
+        except Exception as e:
```
sr525, line 553: Same comment as above.

python/lsst/pipe/tasks/matchFakes.py
surhudm: Not really, the main one I was considering was if `visit` is not present in the `fakeCat`. I suppose I can just remove all these exceptions and let the task fail in case `visit` is not present in the catalog.

yalsayyad:
The commit with change was merged in `tickets/DM-33327`, but this PR won't close automatically.

PR #625, opened 2022-01-24
DM-13899: Fix AstrometryTask API in ImageDifferenceTask

isullivan:
This could also have been written to instead use `self.astrometer`, though that would not pick up the custom config value for the matcher.


parejkoj (approved):
Thank you for also passing `config` as a kwarg.

PR #634, opened 2022-02-14
DM-33221: Add DiffMatchedTractCatalogTask

taranu:


python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -208,6 +221,228 @@ def columns_in_target(self) -> Set[str]:
...
+        keyed by the column suffix.
+    prefix : `str`
+        A prefix for all column names (e.g. band).
+    skip_diff : `bool`
+        Whether to skip computing statistics on differences. Note that
+        differences will still be computed for chi statistics.
+
+    Returns
+    -------
+    row_stats : `numpy.ndarray`, (1, C)
```
sr525, line 345: row_with_stats? Make it more obvious that this is the original row with the stats and not just the stats for the row.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -301,5 +537,221 @@ def run(
             cat_matched.index = cat_left.index
         cat_matched.columns.values[len(cat_target.columns):] = [f'refcat_{col}' for col in cat_right.columns]
 
-        retStruct = pipeBase.Struct(cat_matched=cat_matched)
+        # Add/compute distance columns
+        coord1_target_err, coord2_target_err = config.columns_target_coord_err
+        column_dist, column_dist_err = 'dr', 'drErr'
```
sr525, line 542: Is this the name that the columns end up with in the table? If so could it be something more descriptive?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -301,5 +537,221 @@ def run(
...
+            ),
+            target.column_coord2: (
+                ref.column_coord2, target.column_coord2, coord2_target_err, False,
+            ),
+            column_dist: (column_dummy, column_dist, column_dist_err, False),
+        }
+
+        band_prev = None
+        for idx_band, (band, config_flux) in enumerate(config.columns_flux.items()):
+            mag_ref = -2.5 * np.log10(cat_ref[config_flux.column_ref_flux]) + config.mag_zeropoint_ref
```
sr525, line 621: Spaces around *

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -301,5 +537,221 @@ def run(
...
+                columns_target[f'mag_{column_target_flux}'] = (
+                    column_mag_temp, column_mag_temp_model, None, False,
+                )
+
+                if idx_band > 0:
+                    column_color_temp_model = f'{column_color_temp}{idx_model}'
+                    column_color_err_temp_model = f'{column_color_err_temp}{idx_model}'
+
+                    # e.g. if order is ugrizy, first color will be u - g
+                    cat_target[column_color_temp_model] = mag_prev[idx_model] - mag_model[idx_model]
```
sr525, line 665: Are they guaranteed to be in order?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: It was meant to be consistent with the default (pixel) coordinate naming of x and y, but I suppose `coord_x`,`coord_y` and `distance` would be clearer.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: They're in config.columns_flux order, so the user can specify the N-1 colours they want. I suppose I could output all N*(N-1)/2 unique colours instead, since that's "only" 15 instead of 5. I'll leave that for a future hypothetical feature request.

PR #636, opened 2022-02-23
DM-33786: Remove try block for stacking

yalsayyad:
because log.fatal does not exit.
There is still a log.debug that prints each bbox in the loop
if more info on which subregion it is working on is needed.


erykoff (approved):
Ah, so the thought was that the fatal log would exit, but instead it just continued with a bad section?  And removing this will let it die in peace.

erykoff (approved):

PR #639, opened 2022-03-04
DM-33888: Add config parameter to grow streak footprints

cmsaunders:
This adds a parameter that allows the streak footprints to be grown by the parameter times the PSF width.


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -2012,6 +2012,11 @@ class CompareWarpAssembleCoaddConfig(AssembleCoaddConfig,
         dtype=bool,
         default=False
     )
+    growStreakFp = pexConfig.Field(
+        doc="Grow streak footprints by this number x the PSF width",
```
leeskelvin, line 2016: Would this read more clearly if it were written as:
```
"Grow streak footprints by this number multiplied by the PSF width"
```
?

PR #640, opened 2022-03-07
DM-33966: Create a fractional residual action for PSF size residuals

arunkannawadi:


python/lsst/pipe/tasks/dataFrameActions/_actions.py
```diff
@@ -121,14 +122,22 @@ def __call__(self, df, flux_column=None, flux_mag_err=None, **kwargs):
...
                                                          "colB": SingleColumnAction},
                                            docstring=_docs)
 
+_docs = """This is a `MultiColumnAction` that is designed to divide two columns
+together, subtract one and return the result.
+"""
+FractionalResidualColumns = makeColumnExpressionAction("FractionalResidualColumns", "(colA-colB)/colB",
+                                                       exprDefaults={"colA": SingleColumnAction,
+                                                                     "colB": SingleColumnAction},
+                                                       docstring=_docs)
```
leeskelvin, line 139: Ah, I guess this is why you only need to specify `ixx` and `iyy` in the pipeline yaml, and not `ixy`. I don't know if that is the correct thing to do scientifically, but at least I understand why it doesn't crash now! :)

leeskelvin, line 139: Ah, I guess this is why you only need to specify `ixx` and `iyy` in the pipeline yaml, and not `ixy`. I don't know if that is the correct thing to do scientifically, but at least I understand why it doesn't crash now! :)

arunkannawadi: No, this is not why not specifying `ixy` didn't crash the pipeline (too many negatives!). The real reason was [this](https://github.com/lsst/analysis_drp/blob/main/python/lsst/analysis/drp/calcFunctors.py#L135-L137).

PR #644, opened 2022-03-10
DM-33948: Add EBV to Object table (and dustmaps_cache as a dep)

yalsayyad:


python/lsst/pipe/tasks/functors.py
```diff
@@ -1899,3 +1905,21 @@ def _func(self, df):
...
+    name = "E(B-V)"
+    shortname = "ebv"
+
+    def __init__(self, **kwargs):
+        self._columns = ['coord_ra', 'coord_dec']
+        self.sfd = SFDQuery()
+        super().__init__(**kwargs)
+
+    def _func(self, df):
+        coords = SkyCoord(df['coord_ra']*u.rad, df['coord_dec']*u.rad)
```
erykoff, line 1923: Aren't `coord_ra` and `coord_dec` (somewhat confusingly) in degrees in the dataframes?

yalsayyad: Nope. They're degrees in the Object Tables. But they're still radians in `deepCoadd_ref` and `deepCoadd_obj`.

Remember that these functors are the ones for transforming the columns from catalogs to standard data model columns.  i.e. these are the functors that translate radians to degrees.

TallJimbo (approved):

PR #645, opened 2022-03-14
DM-33980: adding psfFlux and psfFluxErr columns to isolatedStarAssociationTask

psferguson:
In order to use the output of` isolatedStarAssociationTask` for faro photometric repeatability metrics the `isolated_star_sources` output should contain the columns `psfFlux` and `psfFluxErr`.

### DIFF
--- a/python/lsst/pipe/tasks/isolatedStarAssociation.py
+++ b/python/lsst/pipe/tasks/isolatedStarAssociation.py
@@ -128,7 +128,9 @@ class IsolatedStarAssociationConfig(pipeBase.PipelineTaskConfig,
                  'apFlux_17_0_instFluxErr',
                  'apFlux_17_0_flag',
                  'localBackground_instFlux',
-                 'localBackground_flag']
+                 'localBackground_flag',
+                 'psfFlux',
+                 'psfFluxErr']
     )
     source_selector = sourceSelectorRegistry.makeField(
         doc='How to select sources.  Under normal usage this should not be changed.',


python/lsst/pipe/tasks/isolatedStarAssociation.py
```diff
@@ -128,7 +128,9 @@ class IsolatedStarAssociationConfig(pipeBase.PipelineTaskConfig,
                  'apFlux_17_0_instFluxErr',
                  'apFlux_17_0_flag',
                  'localBackground_instFlux',
-                 'localBackground_flag']
+                 'localBackground_flag',
+                 'psfFlux',
+                 'psfFluxErr']
```
erykoff, line 133: Do we want the psfFlux or the psf instflux?  Also you probably want the associated flag.

psferguson: I think the psfFlux if we are computing photometric repeatability from this (comparing measurements of a single source).

I guess we would want the psf_instflux If the calibration is going to be superseded and we should be applying updated zeropoints.

Agree on the flag.

erykoff: If you are going to be applying external calibrations (and I believe you are!) then you need the instFluxes.  Furthermore, at least at the current time, the "calibrated" fluxes aren't even calibrated according to the correct calibration (there's a ticket on this).  In the farther future, you'll be able to get fluxes with the best calibrations out of the source tables, but we are still far from that point.

bechtol: To confirm, is this the referenced ticket for consistent calibrations between the source table and calexp.photoCalib: https://jira.lsstcorp.org/browse/DM-34019 ?

Two more questions:
(1) Assuming the above ticket is resolved, should we be using `local_PhotoCalib` column for the source table, or should we be loading the photoCalib from the butler?
(2) Do the (ra, dec) columns in the source table represent the calibrated coordinates?

erykoff: Yeah, that's the ticket!

(1) Until we are running source tables with the best calibrations (not coming imminently), faro is going to want to run the (fgcm) recalibration.  And therefore should be taking in instfluxes rather than calibrated fluxes and the local photocalib, and then trying to remove the calibration before applying the new one.
2) The ra, dec are from the initial wcs in the single frame processing.  So faro will want to apply jointcal recalibration to the x, y positions.  (But because of the bug from that ticket, I believe that the local_wcs will not translate between x, y and ra, dec.)

PR #647, opened 2022-03-15
DM-33857: Make Piff the default PsfEstimator.

jmeyers314:


tests/test_isPrimaryFlag.py
```diff
@@ -161,6 +161,8 @@ def setUp(self):
 
         # Characterize the image (create PSF, etc.)
         charImConfig = CharacterizeImageConfig()
+        charImConfig.measurePsf.psfDeterminer = 'piff'
```
arunkannawadi, line 164: Should the default psfDeterminer in `CharacterImageConfig` also be switched to "piff", instead of setting it here explicitly? In any case, it should be in double quotes for consistency with the rest of the file and in other files that have changed.

tests/test_isPrimaryFlag.py
```diff
@@ -161,6 +161,8 @@ def setUp(self):
...
+        charImConfig.measurePsf.psfDeterminer['piff'].spatialOrder = 0
```
arunkannawadi, line 165: I could look it up, but I'm lazy. What's the default value for `spatialOrder`?

tests/test_processCcd.py
```diff
@@ -147,29 +147,32 @@ def testProcessCcd(self):
...
                     expectedPlaces = 7  # Tolerance for numerical comparisons
                     for name, var, val in [
-                        ("bgMean", bgMean, 191.48623786891795),
-                        ("bgStdDev", bgStdDev, 0.23994185672586282),
-                        ("numGoodPix", numGoodPix, 1966606),
-                        ("imMean", imMean, 1.1242456954648634),
-                        ("imStdDev", imStdDev, 85.8129750182329),
+                        ("bgMean", bgMean, 191.48618251407825),
+                        ("bgStdDev", bgStdDev, 0.23993854036946),
+                        ("numGoodPix", numGoodPix, 1966533.00000000000000),
```
arunkannawadi, line 152: `numGoodPix` was previously an integer. Did you mean to cast it to a float here consciously?

arunkannawadi (approved):
LGTM, and excited to see Piff results!

tests/test_isPrimaryFlag.py
jmeyers314: Same as PSFEx: 2.
Swapping it to zero here just for unit test which includes very few stars.

tests/test_processCcd.py
jmeyers314: Good catch.  Integer definitely makes sense for a number of pixels.  Will fix.

tests/test_isPrimaryFlag.py
jmeyers314: I set it here b/c it's a unit test, and I don't want to assume anything about the API for the next great PsfEstimator.

`CharacterizeImageTask` already gets its psf estimator via `MeasurePsfTask`, which is where the default got reset.

PR #648, opened 2022-03-15
DM-33949: Add DiffMatchedTractCatalogTask

taranu:


python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -0,0 +1,306 @@
...
+        doc="Configs for flux columns for each band",
+    )
+    columns_ref_copy = pexConfig.ListField(
+        dtype=str,
+        default=set(),
+        doc='Reference table columns to copy to copy into cat_matched',
+    )
+    columns_target_coord_err = pexConfig.ListField(
+        dtype=str,
+        listCheck=lambda x: (len(x) == 2) and (x[0] != x[1]),
```
sr525, line 189: I thought we had policies on not using lambdas? Are we allowed to do this?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -0,0 +1,306 @@
...
+        skymap = inputs.pop("skymap")
+
+        outputs = self.run(
+            catalog_ref=inputs['cat_ref'].get(parameters={'columns': self.config.columns_in_ref}),
+            catalog_target=inputs['cat_target'].get(parameters={'columns': self.config.columns_in_target}),
+            catalog_match_ref=inputs['cat_match_ref'].get(
+                parameters={'columns': ['match_candidate', 'match_row']},
+            ),
+            catalog_match_target=inputs['cat_match_target'].get(
+                # parameters={'columns': ['match_candidate']},
```
sr525, line 230: Should this be commented out?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: The dev guide says "should not be used", but I don't think a `def checkLengthTwoSet(listfield):` function would improve anything.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: It's from DM-33221, so I suppose I can just delete it for now.

PR #652, opened 2022-03-22
DM-32756: Compute average FWHM of the PSF on an exposure as the media…

surhudm:
…n of PSF determined on a grid of points

### DIFF
--- a/python/lsst/pipe/tasks/imageDifference.py
+++ b/python/lsst/pipe/tasks/imageDifference.py
@@ -345,6 +345,17 @@ class ImageDifferenceConfig(pipeBase.PipelineTaskConfig,
         target=SkyObjectsTask,
         doc="Generate sky sources",
     )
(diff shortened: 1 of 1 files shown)


ktlim:
Commit comments don't usually need to include the ticket number; the PR title is sufficient.  Commit comments should also be under 50 characters: https://developer.lsst.io/work/flow.html#writing-commit-summary-lines

arunkannawadi:
Changing the base to a corresponding ticket branch within `lsst` so we can test this more thoroughly and merge it to `main`. Thanks @surhudm !

isullivan:
Note that these changes need to be made in `lsst.ip.diffim.subtractImages` now, instead of in `pipe_tasks`

arunkannawadi:
Thanks Ian! ... I'll look at the refactoring that happened since these changes were made. Alternatively, I'd appreciate if we can paircode this ticket.

PR #653, opened 2022-03-22
DM-34019: Run local photocalib/wcs plugins after calibration

yalsayyad:


python/lsst/pipe/tasks/calibrate.py
```diff
@@ -780,6 +795,11 @@ def run(self, exposure, exposureIdInfo=None, background=None,
                 exposure=exposure,
                 exposureId=exposureIdInfo.expId
             )
+            self.postCalibrationMeasurement.run(
+                measCat=sourceCat,
+                exposure=exposure,
+                exposureId=exposureIdInfo.expId
+            )
```
TallJimbo, line 802: I don't want to complicate this ticket even a little bit, but this `doInsertFakes` block here is entirely about the old synpipe interface, right, not anything we're using now?  I'd forgotten it even existed, but we should probably get it removed. If you agree I'll RFC it, even if that's just a formality (it's a config API deprecation/change at least).

TallJimbo (approved):

python/lsst/pipe/tasks/calibrate.py
sr525: Pretty sure that it is from synpipe and we are refactoring the fakes stuff while we remove the gen 2 only code anyway.

PR #655, opened 2022-03-24
tickets/DM-33977 Fixes for ConfigurableActionStructField

natelust:
This ticket fixes some assignment logic in ConfigurableActionStructField wherein a local copy of a ConfigurableActionStruct was not being made, leaving a dangling reference to a config object that goes out of scope.

This ticket also makes a change to stay in line with other pex_config fields wherein it takes other python classes that duck like a ConfigurableActionStructField as an argument in assigment.

Finally this ticket adds a verification that a string used in the dict update interface is a valid python identifier.

### DIFF
--- a/python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
+++ b/python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
@@ -22,11 +22,13 @@ from __future__ import annotations
 
 __all__ = ("ConfigurableActionStructField", "ConfigurableActionStruct")
 
+from types import SimpleNamespace
 from typing import Iterable, Mapping, Optional, TypeVar, Union, Type, Tuple, List, Any, Dict
 
 from lsst.pex.config.config import Config, Field, FieldValidationError, _typeStr, _joinNamePath
 from lsst.pex.config.comparison import compareConfigs, compareScalars, getComparisonName
 from lsst.pex.config.callStack import StackFrame, getCallStack, getStackFrame
+from lsst.pipe.base import Struct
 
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -237,6 +247,21 @@ def __set__(self, instance: Config,
...
         else:
+            # An actual value is being assgigned check for what it is
+            if isinstance(value, self.StructClass):
+                # If this is a ConfigurableActionStruct, we need to make our own
+                # copy that references this current field
+                value = self.StructClass(instance, self, value._attrs, at=at, label=label)
+            elif isinstance(value, (SimpleNamespace, Struct)):
+                # If this is a a python analogous container, we need to make
+                # a ConfigurableActionStruct initialized with this data
+                value = self.StructClass(instance, self, vars(value), at=at, label=label)
```
TallJimbo, line 258: Are you sure that the call to `vars` doesn't return any "special" attributes for either `SimpleNamespace` or `Struct`?  Seems like something worth adding a test for.

TallJimbo (approved):
I think this needs a unit test.  Otherwise looks good, but fix the "assignements" type in one of the commit messages, too.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
natelust: I verified it though with a few tests in python manually, but I will add a unit test to guard against behavior changes (especially in Struct)

PR #656, opened 2022-03-29
DM-34217: Move ScaleVarianceTask to meas_algorithms

isullivan:


erykoff (changes requested):
The moving of the task from one place to another will work for our local code, but if anybody else using science pipelines tries `import lsst.pipe.tasks.ScaleVarianceTask`they're going to get a big surprise.  So I think we need to import `ScaleVarianceTask` into the pipe_tasks namespace with a deprecation warning.

isullivan:
> The moving of the task from one place to another will work for our local code, but if anybody else using science pipelines tries `import lsst.pipe.tasks.ScaleVarianceTask`they're going to get a big surprise. So I think we need to import `ScaleVarianceTask` into the pipe_tasks namespace with a deprecation warning.

OK, I've left a stub `ScaleVarianceTask` in pipe_tasks, which will call the new Task and print a deprecation warning. Is that what you had in mind?

erykoff:
Ah, we have a better way of doing deprecation: https://developer.lsst.io/stack/deprecating-interfaces.html . I don't have time to search for a specific example of a task being moved, but this is somewhat related: https://github.com/lsst/daf_butler/blob/90f5a21b374cee11ca52f65e10ce3972def880f0/python/lsst/daf/butler/core/_butlerUri.py#L33-L38

PR #660, opened 2022-04-06
DM-33157: Fix doxygen errors in pipe_tasks

kfindeisen:
This PR cleans up a very long list of Doxygen errors in `pipe_tasks`, as well as some unrelated warnings that were easy fixes. Given the ongoing (if stalled) documentation migration, I've prioritized error-free builds over correct ones, intervening for correctness only if I found an issue that badly corrupted the documentation (e.g., failure to properly close `@link` tags).

While few of the fixes overlap, some of them are non-obvious, so I recommend reviewing this PR commit-by-commit for context.

### DIFF
--- a/python/lsst/pipe/tasks/calibrate.py
+++ b/python/lsst/pipe/tasks/calibrate.py
@@ -359,8 +359,8 @@ class CalibrateTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
     @section pipe_tasks_calibrate_Purpose  Description
 
     Given an exposure with a good PSF model and aperture correction map
-    (e.g. as provided by @ref CharacterizeImageTask), perform the following
-     operations:
+    (e.g. as provided by @ref characterizeImage::CharacterizeImageTask "CharacterizeImageTask"),
+    perform the following operations:
     - Run detection and measurement
     - Run astrometry subtask to fit an improved WCS
     - Run photoCal subtask to fit the exposure's photometric zero-point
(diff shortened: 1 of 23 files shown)


python/lsst/pipe/tasks/measurePsf.py
```diff
@@ -121,15 +123,12 @@ class MeasurePsfTask(pipeBase.Task):
...
 @section pipe_tasks_measurePsf_Example	A complete example of using MeasurePsfTask
 
-This code is in @link measurePsfTask.py@endlink in the examples directory, and can be run as @em e.g.
+This code is in `measurePsfTask.py` in the examples directory, and can be run as @em e.g.
 @code
 examples/measurePsfTask.py --doDisplay
 @endcode
 @dontinclude measurePsfTask.py
 
-The example also runs SourceDetectionTask and SingleFrameMeasurementTask;
```
ktlim, line 130: Isn't this still true?

tests/test_finalizeCharacterization.py
```diff
@@ -64,6 +64,8 @@ def get(self, parameters={}, **kwargs):
 class TestFinalizeCharacterizationTask(FinalizeCharacterizationTask):
     """A derived class which skips the initialization routines.
     """
+    __test__ = False  # Stop Pytest from trying to parse as a TestCase
```
ktlim, line 67: Maybe a different name would be better, but I guess this works.

ktlim (approved):

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -336,7 +336,7 @@ def validate(self):
 
 ## \addtogroup LSST_task_documentation
 ## \{
-## \page CalibrateTask
+## \page page_CalibrateTask CalibrateTask
```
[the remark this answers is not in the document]
timj: And we know no-one outside of pipe_tasks is going to reference these since we are meant to be using sphinx.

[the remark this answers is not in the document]
kfindeisen: These blocks are used to generate [this listing](http://doxygen.lsst.codes/stack/doxygen/x_mainDoxyDoc/group___l_s_s_t__task__documentation.html). I'm fairly confident that nobody would try to link to the individual sections on the page (at least, not on purpose).

python/lsst/pipe/tasks/measurePsf.py
kfindeisen: As far as I can tell, the `meas_algorithms_measurement_Example` target hasn't existed in years, and I can't reconstruct what it was originally meant to explain.

kfindeisen: Restored just in case.

PR #662, opened 2022-04-07
DM-34364: Use the PSF-matched template for decorrelation

isullivan:
Decorrelation is supposed to correct the correlated noise caused by convolving the template with the PSF matching kernel, so we need to actually supply the PSF-matched template to the calculation. Currently we are passing in the warped, but not PSF-matched, template.


parejkoj (approved):
Assuming that the logic in the described ticket is correct, this looks fine.

PR #669, opened 2022-04-19
DM-34484: Update tests to remove dependency on obs_test.

erykoff:


tests/test_photoCal.py
```diff
@@ -36,12 +35,10 @@
 from lsst.pipe.tasks.photoCal import PhotoCalTask, PhotoCalConfig
 from lsst.pipe.tasks.colorterms import Colorterm, ColortermDict, ColortermLibrary
 from lsst.pipe.base.task_logging import TRACE
+from lsst.meas.algorithms.testUtils import MockLoadReferenceObjects
 
 RefCatDir = os.path.join(getPackageDir("pipe_tasks"), "tests", "data", "sdssrefcat")
```
timj, line 40: Interesting that we have the same refcats in pipe_tasks and meas_astrom...

timj (approved):
I assume there is no coverage loss in removing those gen2 Transform tests.

tests/test_photoCal.py
erykoff: It is interesting!  Something to consider, I suppose, but not on this ticket...

PR #670, opened 2022-04-25
DM-34391: Change default to assemble coadds with finalized PSF model.

erykoff:


python/lsst/pipe/tasks/makeCoaddTempExp.py
```diff
@@ -100,7 +100,7 @@ class MakeCoaddTempExpConfig(CoaddBaseTask.ConfigClass):
     doApplyFinalizedPsf = pexConfig.Field(
         doc="Whether to apply finalized psf models and aperture correction map.",
         dtype=bool,
-        default=False,
+        default=True,
```
laurenam, line 103: Ok, so with this default, the user is expected to have run the `isolatedStarAssociation` and `finalizeCharacterization` tasks, right?  In the case where they haven't, what error message do they get?  I'm wondering if it's possible for `doApplyFinalizedPsf = True`, but in reality `finalizedPsfApCorrCatalog = None` and the correction silently just doesn't get applied (because `finalizedPsfApCorrCatalog` being not `None` is the condition that is checked in `prepareCalibratedExposures()`, without regard to the value of `doApplyFinalizedPsf`).

erykoff: Interesting question.  If running via `pipetask`, then the pipetask quantum graph generation will fail and say that required input connection `finalizedPsfApCorrCatalog` is not available.  If running the task directly then it will not get applied (but this is how all the external calibrations are used in `runQuantum`/`run`).

So I _think_ this is okay.

python/lsst/pipe/tasks/propagateSourceFlags.py
```diff
@@ -212,7 +210,11 @@ def run(self, coadd_object_cat, ccd_inputs,
                                 return_indices=True
                             )
                         except IndexError:
-                            # No matches.  Workaround a bug in smatch.
+                            # No matches.  Workaround a bug in older version of smatch.
+                            continue
+
+                        if len(i1) == 0:
+                            # No matches (usually because detector does not overlap patch).
```
laurenam, line 217: Does finding no matches warrant a log message?

laurenam (approved):

PR #672, opened 2022-04-28
DM-34254: Validate psf configuration and output

parejkoj:


ebellm (changes requested):
I think you should do the check you have in `dcrAssembleCoadd.py` in `assembleCoadd.py` as well.

isullivan:
In this case, the new check in `dcrAssembleCoadd.py` is for an option that does not exist in `assembleCoadd.py`

ebellm (approved):

PR #673, opened 2022-04-28
DM-34603: Add trailed-source injection to insertFakes

langfzac:
- Add trailed sources as a `srcType` and add galsim object creation
- Include `srcType` in generator returned from `_generateGSObjectsFromCatalog` (needed to compute the trail image within `_add_fake_sources`)
- Add trail length and angle columns to config

### DIFF
--- a/python/lsst/pipe/tasks/insertFakes.py
+++ b/python/lsst/pipe/tasks/insertFakes.py
@@ -393,6 +393,18 @@ class InsertFakesConfig(PipelineTaskConfig,
         default="select",
     )
 
+    length_col = pexConfig.Field(
+        doc="Source catalog column name for trail length (in pixels).",
+        dtype=str,
+        default="trail_length",
+    )
+
+    angle_col = pexConfig.Field(
+        doc="Source catalog column name for trail angle (in radians).",
+        dtype=str,
+        default="trail_angle",
+    )
+
     # Deprecated config variables
 
     raColName = pexConfig.Field(
@@ -649,12 +661,15 @@ class InsertFakesTask(PipelineTask, CmdLineTask):
                 if isinstance(fakeCat[self.config.sourceType].iloc[0], str):
                     galCheckVal = "galaxy"
                     starCheckVal = "star"
+                    trailCheckVal = "trail"
                 elif isinstance(fakeCat[self.config.sourceType].iloc[0], bytes):
                     galCheckVal = b"galaxy"
                     starCheckVal = b"star"
+                    trailCheckVal = b"trail"
                 elif isinstance(fakeCat[self.config.sourceType].iloc[0], (int, float)):
                     galCheckVal = 1
                     starCheckVal = 0
+                    trailCheckVal = 2
                 else:
                     raise TypeError(
                         "sourceType column does not have required type, should be str, bytes or int"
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -835,11 +861,23 @@ def _generateGSObjectsFromCatalog(self, exposure, fakeCat, galCheckVal, starChec
                 gal = bulge*row['bulge_disk_flux_ratio'] + disk
                 gal = gal.withFlux(flux)
 
-                yield skyCoord, gal
+                yield skyCoord, gal, "gal"
```
sr525, line 864: What is the added string for?

langfzac: The galsim object for a trail needs to be updated inside of the `_add_fake_sources`, so this is just passing the source type to that function. There's nothing done if it's not `"trail"`, but I added strings for each just for consistency.

sr525: Can that not be done within the generateObjects function?

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -835,11 +861,23 @@ def _generateGSObjectsFromCatalog(self, exposure, fakeCat, galCheckVal, starChec
...
                 star = galsim.DeltaFunction()
                 star = star.withFlux(flux)
-                yield skyCoord, star
+                yield skyCoord, star, "star"
+            elif sourceType == trailCheckVal:
+                length = row['trail_length']
+                angle = row['trail_angle']
+
+                # Make a 'thin' box profile for a trailed source
+                thickness = 1e-6
```
sr525, line 874: How was thickness decided on? Should this be a config option? Please can you add a comment about the value.

python/lsst/pipe/tasks/insertFakes.py
langfzac: It could, but it would duplicate all the code to obtain the `gsWCS` on line 85. I suppose the relevant code could be put in a function and used both here and in generateObjects. Would that be preferable?

python/lsst/pipe/tasks/insertFakes.py
langfzac: The thickness choice was largely arbitrary, and seemed to work empirically. I'm just looking to mimic a 'line' surface brightness profile, which isn't implemented in galsim (as far as I know). I don't think it makes sense to make configurable, but I'll add a better comment.

sr525: @jmeyers314 once said something about this but I can't remember what it was.

python/lsst/pipe/tasks/insertFakes.py
sr525: I don't think it would be a huge amount of code duplication, I think it should just be lines 84 and 85. If we forsee needing it in more places as we develop the code further than we could pass it to the function. Otherwise I think duplicating those two lines here is preferable.

python/lsst/pipe/tasks/insertFakes.py
langfzac: If I recall, the box is what he recommended doing.

python/lsst/pipe/tasks/insertFakes.py
langfzac: Hmm, you're definitely right. I thought there was way more code to copy... I'll change that and remove the string arguments

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -840,6 +858,24 @@ def _generateGSObjectsFromCatalog(self, exposure, fakeCat, galCheckVal, starChec
                 star = galsim.DeltaFunction()
                 star = star.withFlux(flux)
                 yield skyCoord, star
+            elif sourceType == trailCheckVal:
+                length = row['trail_length']
+                angle = row['trail_angle']
+
+                # Make a 'thin' box to mimic a line surface brightness profile
+                thickness = 1e-6  # Make the box much thinner than a pixel
+                theta = galsim.Angle(angle * galsim.radians)
```
sr525, line 867: No space around *

PR #675, opened 2022-05-03
DM-34466-v23: Port functionality to reevaluate calibs in writeSourceTable to gen3

yalsayyad:
- Split functionality to reevaluate calibs by rerunning plugins
  base_LocalWcs and base_LocalPhotoCalib
  into a subclass WriteRecalibratedSourceTable
- move configs doApplyExternalPhotoCalib and doApplyExternalSkyWcs
  to subclass.
- Add ability to configurably apply global calibrations from e.g.
  jointcal and FGCM.

### DIFF
--- a/python/lsst/pipe/tasks/postprocess.py
+++ b/python/lsst/pipe/tasks/postprocess.py
@@ -22,6 +22,7 @@
 import functools
 import pandas as pd
 from collections import defaultdict
+import logging
 import numpy as np
 import numbers
 
@@ -29,16 +30,20 @@ import lsst.geom
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 import lsst.daf.base as dafBase
+from lsst.obs.base import ExposureIdInfo
 from lsst.pipe.base import connectionTypes
 import lsst.afw.table as afwTable
 from lsst.meas.base import SingleFrameMeasurementTask
 from lsst.pipe.base import CmdLineTask, ArgumentParser, DataIdContainer
 from lsst.coadd.utils.coaddDataIdContainer import CoaddDataIdContainer
 from lsst.daf.butler import DeferredDatasetHandle, DataCoordinate
+from lsst.skymap import BaseSkyMap
 
 from .parquetTable import ParquetTable
 from .multiBandUtils import makeMergeArgumentParser, MergeSourcesRunner
 from .functors import CompositeFunctor, Column
+
+log = logging.getLogger(__name__)
 
 
 def flattenFilters(df, noDupCols=['coord_ra', 'coord_dec'], camelCase=False, inputBands=None):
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -329,75 +333,359 @@ def run(self, catalog, ccdVisitId=None):
...
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        inputs['ccdVisitId'] = butlerQC.quantum.dataId.pack("visit_detector")
+        inputs['exposureIdInfo'] = ExposureIdInfo.fromDataId(butlerQC.quantum.dataId, "visit_detector")
+
+        if self.config.doReevaluatePhotoCalib or self.config.doReevaluateSkyWcs:
+            if self.config.doApplyExternalPhotoCalib or self.config.doApplyExternalSkyWcs:
+                inputs['exposure'] = self.attachCalibs(inputRefs, **inputs)
+
+            inputs['catalog'] = self.addCalibColumns(**inputs)
```
kherner, line 485: Will the call work with just **inputs given that the arguments are (self, catalog, exposure, exposureIdInfo, **kwargs)? I think it usually will but do you need any protections against a case where one of catalog, exposure, or exposureIdInfo is missing from inputs? Maybe the upstream stuff is robust enough that something else would have failed already if you found yourself in such a case.

kherner (changes requested):

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -329,75 +333,359 @@ def run(self, catalog, ccdVisitId=None):
...
+                tracts = [ref.dataId['tract'] for ref in inputRef]
+                if len(tracts) == 1:
+                    ind = 0
+                    self.log.info('Applying tract-level SkyWcs from tract %s', tracts[ind])
+                else:
+                    ind = self.getClosestTract(tracts, skyMap,
+                                               exposure.getBBox(), exposure.getWcs())
+                    self.log.info('Multiple overlapping externalSkyWcsTractCatalogs found (%s). '
+                                  'Applying closest to detector center: tract=%s', str(tracts), tracts[ind])
+            externalSkyWcsCatalog = externalSkyWcsTractCatalog[ind]
```
kherner, line 536: Same comment here about whether this line should be inside `if externalSkyWcsTractCatalog`. I think it has to be otherwise you could get a situation where ind is not defined (unless you can be 100% confident that externalSkyWcsTractCatalog will always be there).

python/lsst/pipe/tasks/postprocess.py
yalsayyad: `self.run(**inputs)` is our canonical `runQuantum`. Task methods are expected to have arguments that match the names of the input connection `Input` s.

https://pipelines.lsst.io/v/w_2022_14/modules/lsst.pipe.base/creating-a-pipelinetask.html#overriding-task-execution 

Following the pattern: data-prep methods should take  **inputs and write out new inputs. 

If `exposure` is missing for example you have an empty QG.

kherner (approved):

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -329,75 +333,359 @@ def run(self, catalog, ccdVisitId=None):
...
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        # Same connection boilerplate as all other applications of
+        # Global/Tract calibrations
+        if config.doApplyExternalSkyWcs and config.doReevaluateSkyWcs:
+            if config.useGlobalExternalSkyWcs:
+                self.inputs.remove("externalSkyWcsTractCatalog")
+            else:
+                self.inputs.remove("externalSkyWcsGlobalCatalog")
+        else:
```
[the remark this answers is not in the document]
yalsayyad: (if you're wondering, where the warning went: the `validate` method in the `Config `class is the best place for warnings about config combinations.)

PR #676, opened 2022-05-03
DM-34466: Port functionality to reevaluate calibs in writeSourceTable to gen3

yalsayyad:
- Split functionality to reevaluate calibs by rerunning plugins
  base_LocalWcs and base_LocalPhotoCalib
  into a subclass WriteRecalibratedSourceTable
- move configs doApplyExternalPhotoCalib and doApplyExternalSkyWcs
  to subclass.
- Add ability to configurably apply global calibrations from e.g.
  jointcal and FGCM.

### DIFF
--- a/python/lsst/pipe/tasks/postprocess.py
+++ b/python/lsst/pipe/tasks/postprocess.py
@@ -22,6 +22,7 @@
 import functools
 import pandas as pd
 from collections import defaultdict
+import logging
 import numpy as np
 import numbers
 import os
@@ -30,16 +31,20 @@ import lsst.geom
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 import lsst.daf.base as dafBase
+from lsst.obs.base import ExposureIdInfo
 from lsst.pipe.base import connectionTypes
 import lsst.afw.table as afwTable
 from lsst.meas.base import SingleFrameMeasurementTask
 from lsst.pipe.base import CmdLineTask, ArgumentParser, DataIdContainer
 from lsst.coadd.utils.coaddDataIdContainer import CoaddDataIdContainer
 from lsst.daf.butler import DeferredDatasetHandle, DataCoordinate
+from lsst.skymap import BaseSkyMap
 
 from .parquetTable import ParquetTable
 from .multiBandUtils import makeMergeArgumentParser, MergeSourcesRunner
 from .functors import CompositeFunctor, Column
+
+log = logging.getLogger(__name__)
 
 
 def flattenFilters(df, noDupCols=['coord_ra', 'coord_dec'], camelCase=False, inputBands=None):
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -278,18 +280,7 @@ class WriteSourceTableConnections(pipeBase.PipelineTaskConnections,
 
 class WriteSourceTableConfig(pipeBase.PipelineTaskConfig,
```
kherner, line 281: Is this just left here for compatibility reasons? Otherwise I don't see why it wouldn't be better to just remove it.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -330,75 +312,347 @@ def run(self, catalog, ccdVisitId=None):
...
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        # Same connection boilerplate as all other applications of
+        # Global/Tract calibrations
+        if config.doApplyExternalSkyWcs and config.doReevaluateSkyWcs:
+            if config.useGlobalExternalSkyWcs:
+                self.inputs.remove("externalSkyWcsTractCatalog")
+            else:
+                self.inputs.remove("externalSkyWcsGlobalCatalog")
+        else:
```
kherner, line 380: Same comment I made on the other PR about adding a warning.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -330,75 +312,347 @@ def run(self, catalog, ccdVisitId=None):
...
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+        inputs['ccdVisitId'] = butlerQC.quantum.dataId.pack("visit_detector")
+        inputs['exposureIdInfo'] = ExposureIdInfo.fromDataId(butlerQC.quantum.dataId, "visit_detector")
+
+        if self.config.doReevaluatePhotoCalib or self.config.doReevaluateSkyWcs:
+            if self.config.doApplyExternalPhotoCalib or self.config.doApplyExternalSkyWcs:
+                inputs['exposure'] = self.attachCalibs(inputRefs, **inputs)
+
+            inputs['catalog'] = self.addCalibColumns(**inputs)
```
kherner, line 452: Same comment on the other PR here, too.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -330,75 +312,347 @@ def run(self, catalog, ccdVisitId=None):
...
+                else:
+                    ind = self.getClosestTract(tracts, skyMap,
+                                               exposure.getBBox(), exposure.getWcs())
+                    self.log.info('Multiple overlapping externalPhotoCalibTractCatalogs found (%s). '
+                                  'Applying closest to detector center: tract=%s', str(tracts), tracts[ind])
 
-        # Just need the WCS or the PhotoCalib attached to an exposue
-        exposure = dataRef.get('calexp_sub',
-                               bbox=lsst.geom.Box2I(lsst.geom.Point2I(0, 0), lsst.geom.Point2I(0, 0)))
+            externalPhotoCalibCatalog = externalPhotoCalibTractCatalog[ind]
```
kherner, line 521: Should this line be inside the `if externalPhotoCalibTractCatalog:` statement? Looks like it is not.

kherner (changes requested):

python/lsst/pipe/tasks/postprocess.py
yalsayyad: Is the question why we need a `writeSourceTableConfig` that doesn't have any config fields?  Remember that the Task's config class holds the connections, which are configurable:
```  
writePreSourceTable:
    class: lsst.pipe.tasks.postprocess.WriteSourceTableTask
    config:
      connections.outputCatalog: preSource
```

python/lsst/pipe/tasks/postprocess.py
yalsayyad: I totally see why this was confusing. Making clearer.

yalsayyad:
Also, in general:  
-  "Changes requested" means "I must review this again after you implement my suggestions"
-   Writing comments and hitting "Approved" means "You must respond to all my comments, but I don't need to review it again, after you do"

kherner (approved):

PR #679, opened 2022-05-11
DM-34768: Deprecate characterizeImage refcat hooks

parejkoj:


python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -185,10 +185,12 @@ class CharacterizeImageConfig(pipeBase.PipelineTaskConfig,
     )
     refObjLoader = pexConfig.ConfigurableField(
         target=LoadIndexedReferenceObjectsTask,
+        deprecated="This field does nothing. Will be removed after v24 (see DM-34768).",
         doc="reference object loader",
     )
     ref_match = pexConfig.ConfigurableField(
         target=RefMatchTask,
+        deprecated="This field has never been properly tested. Will be removed after v24 (see DM-34768).",
```
TallJimbo, line 193: You can even replace "properly tested" with "usable" if you'd like.

TallJimbo (approved):
I think it'd be fine (and legal, under the principle of Gen2 removal being underway) to go further and actually remove everything but the config fields on this ticket, if you'd like.  No objection if you'd prefer to defer it just to avoid testing-work/waiting, but maybe add a comment here saying that it doesn't have to wait for a release boundary to happen, as long as the config options stay that long.

parejkoj:
I think I'm going to leave things as they are: I'm pretty sure none of this has ever been used by anyone, but I'd rather not discover it the hard way. Doing the usual deprecation process seems fine timeline wise.

PR #681, opened 2022-05-16
DM-34834: Fix crash in IsolatedStarAssociationTask when secondary band doesn't overlap primary.

erykoff:


python/lsst/pipe/tasks/isolatedStarAssociation.py
```diff
@@ -426,7 +426,8 @@ def _match_primary_stars(self, primary_bands, star_source_cat):
                                                self.config.match_radius/3600.)
                 # Any object with a match should be removed.
                 match_indices = np.array([i for i in range(len(idx)) if len(idx[i]) > 0])
-                band_cat = np.delete(band_cat, match_indices)
+                if len(match_indices) > 0:
+                    band_cat = np.delete(band_cat, match_indices)
```
laurenam, line 430: Worth adding a log message?

laurenam (approved):

python/lsst/pipe/tasks/isolatedStarAssociation.py
erykoff: This isn't a problem, it's fine for there to be no overlaps especially at the edges of things.

PR #682, opened 2022-05-17
DM-34635:     Add config to select columns to extract from DataId

yalsayyad:
- Previously, TransformBaseTableTasks populated a column for
    every key in the dataId. Now with the new config parameter,
    columnsFromDataId, users and subclasses choose which subset
    of keys to extract from the dataId and add to the table.
    Default is none.
    Removes e.g. visit_system, physical_filter, and skymap from tables.
    Tables now only extract:
      - Source: visit, detector, band
      - Object: tract, patch
      - ForcedSource: tract, patch (for seeds because TractPatchTable
        does not exist yet)


timj (approved):
Looks okay to me.

PR #684, opened 2022-05-19
DM-34687: Add HighResolutionHipsTask with custom quantum-graph generation.

erykoff:
The quantum-graph code was "pair-coded" with @TallJimbo (e.g. written by him and tested and lightly edited by me.)

### DIFF
--- a/bin.src/build-high-resolution-hips-qg
+++ b/bin.src/build-high-resolution-hips-qg
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+# This file is part of pipe_tasks.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""A custom script to generate a QuantumGraph for HighResolutionHipsTask.
+See ``HighResolutionHipsTask.build_quantum_graph`` for more information
+(including why this script exists), and use ``--help`` for documentation on
+arguments.
+"""
+
+import sys
+
+from lsst.pipe.tasks.hips import HighResolutionHipsTask
+
+
+if __name__ == "__main__":
+    HighResolutionHipsTask.build_quantum_graph_cli(sys.argv[1:])
--- a/python/lsst/pipe/tasks/hips.py
+++ b/python/lsst/pipe/tasks/hips.py
@@ -247,3 +247,172 @@ class HighResolutionHipsTask(pipeBase.PipelineTask):
             )
 
         return pipeBase.Struct(hips_exposures=exp_hpx_dict)
+
+    @classmethod
+    def build_quantum_graph(
+        cls,
+        task_def,
+        registry,
+        constraint_order,
+        constraint_ranges,
+        where=None,
+        collections=None,
+    ):
+        """Generate a `QuantumGraph` for running just this task.
+
+        This is a temporary workaround for incomplete butler query support for
+        HEALPix dimensions.
+
+        Parameters
+        ----------
+        task_def : `lsst.pipe.base.TaskDef`
+            Task definition.
+        registry : `lsst.daf.butler.Registry`
+            Client for the butler database.  May be read-only.
+        constraint_order : `int`
+            HEALPix order used to contrain which quanta are generated, via
+            ``constraint_indices``.  This should be a coarser grid (smaller
+            order) than the order used for the task's quantum and output data
+            IDs, and ideally something between the spatial scale of a patch or
+            the data repository's "common skypix" system (usually ``htm7``).
+        constraint_ranges : `lsst.sphgeom.RangeSet`
+            RangeSet which describes constraint pixels (HEALPix NEST, with order
+            constraint_order) to constrain generated quanta.
+        where : `str`, optional
+            A boolean `str` expression of the form accepted by
+            `Registry.queryDatasets` to constrain input datasets.  This may
+            contain a constraint on tracts, patches, or bands, but not HEALPix
+            indices.  Constraints on tracts and patches should usually be
+            unnecessary, however - existing coadds that overlap the given
+            HEALpix indices will be selected without such a constraint, and
+            providing one may reject some that should normally be included.
+        collections : `str` or `Iterable` [ `str` ], optional
+            Collection or collections to search for input datasets, in order.
+            If not provided, ``registry.defaults.collections`` will be
+            searched.
+        """
+        config = task_def.config
+
+        dataset_types = pipeBase.PipelineDatasetTypes.fromPipeline(pipeline=[task_def], registry=registry)
+        # Since we know this is the only task in the pipeline, we know there
+        # is only one overall input and one overall output.
+        (input_dataset_type,) = dataset_types.inputs
+
+        # Extract the main output dataset type (which needs multiple
+        # DatasetRefs, and tells us the output HPX level), and make a set of
+        # what remains for more mechanical handling later.
+        output_dataset_type = dataset_types.outputs[task_def.connections.hips_exposures.name]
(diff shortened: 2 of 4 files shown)


bin.src/build-high-resolution-hips-qg
```diff
@@ -0,0 +1,35 @@
...
+arguments.
+"""
+
+import sys
+
+from lsst.pipe.tasks.hips import HighResolutionHipsTask
+
+
+if __name__ == "__main__":
+    HighResolutionHipsTask.build_quantum_graph_cli(sys.argv[1:])
```
timj, line 35: I understand that this implementation is a temporary measure to get us through DP0.2. In the future it seems like it should work like `butler` and be a dynamic pluggable command for `pipetask` such that `pipetask hips segment` works. @erykoff can you make a ticket for that and maybe @n8pease can take a look. As things stand the current implementations can't auto-generate any sphinx documentation because there is no entry point for sphinx to examine.

bin.src/build-high-resolution-hips-qg
```diff
@@ -0,0 +1,35 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""A custom script to generate a QuantumGraph for GatherResourceUsageTask.
```
ktlim, line 23: Is this task name copypasta?

bin.src/build-high-resolution-hips-qg
erykoff: I'm not sure if the ticket should be to make this particular usage work in `pipetask` or should be to get the actual quantum graph generation code to be able to figure out healpix overlaps with tracts/patches?

timj: Treating it as a private API is fine for a while but do we know how long @TallJimbo thinks it's going to be until this code disappears?

erykoff: I've added a description of how to use `build-high-resolution-hips-qg` to the task documentation.  I assume that's a net good, but I'm not sure how to note that this is a temporary workaround.

python/lsst/pipe/tasks/hips.py
```diff
@@ -0,0 +1,545 @@
...
+        # is only one overall input and one overall output.
+        (input_dataset_type,) = dataset_types.inputs
+
+        # Extract the main output dataset type (which needs multiple
+        # DatasetRefs, and tells us the output HPX level), and make a set of
+        # what remains for more mechanical handling later.
+        output_dataset_type = dataset_types.outputs[task_def.connections.hips_exposures.name]
+        incidental_output_dataset_types = dataset_types.outputs.copy()
+        incidental_output_dataset_types.remove(output_dataset_type)
+        (hpx_output_dimension,) = (d for d in output_dataset_type.dimensions if d.name != "band")
```
natelust, line 434: I think this is a bit fragile, what if there are more dimensions?

python/lsst/pipe/tasks/hips.py
```diff
@@ -0,0 +1,545 @@
...
+            inputs_by_patch[input_ref.dataId.subset(patch_dimensions)].add(input_ref)
+        if not inputs_by_patch:
+            message_body = '\n'.join(input_refs.explain_no_results())
+            raise RuntimeError(f"No inputs found:\n{message_body}")
+
+        # Iterate over patches and compute the set of output healpix pixels
+        # that overlap each one.  Use that to associate inputs with output
+        # pixels, but only for the output pixels we've already identified.
+        inputs_by_hpx = defaultdict(set)
+        for patch_data_id, input_refs_for_patch in inputs_by_patch.items():
```
natelust, line 504: 3 layer for loop!

python/lsst/pipe/tasks/hips.py
erykoff: @TallJimbo ?

python/lsst/pipe/tasks/hips.py
erykoff: @TallJimbo ?

python/lsst/pipe/tasks/hips.py
```diff
@@ -0,0 +1,545 @@
...
+
+        (task_def,) = expanded_pipeline
+
+        butler = Butler(args.butler_config, collections=args.input)
+
+        if args.subparser_name == 'segment':
+            # Do the segmentation
+            hpix_pixelization = HealpixPixelization(level=args.hpix_build_order)
+            conn = task_def.config.connections
+            dataset = conn.coadd_exposure_handles.format(coaddName=conn.coaddName)
```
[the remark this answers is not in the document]
natelust: What I dont understand is why you are doing that line? cant you just get what you want from task_def.connections.coadd_exposure_handles?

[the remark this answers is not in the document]
erykoff: Because I didn't know that was possible.  Now I do, thanks!

python/lsst/pipe/tasks/hips.py
TallJimbo: It's either this or a loop with `isinstance(SkyPixDimension, ...)` checks; figured this was slightly better (it's far from the only thing that would break if the dimensions change, and I consider that unlikely).

natelust: what I mean is what if the dimensions are `("visit", "detector", "band", "healpix9")` then this will be too large to unpack, could it be switched to be a matching check instead of a not match check?

erykoff: I don't understand that, this is meant to be run on coadds.

natelust: I just used it as an example, it might be any set of dimensions, though I might not be following what Jim is doing here

python/lsst/pipe/tasks/hips.py
```diff
@@ -0,0 +1,545 @@
...
+            data_ids = set(butler.registry.queryDataIds("tract", datasets=dataset).expanded())
+            region_pixels = []
+            for data_id in data_ids:
+                region = data_id.region
+                pixel_range = hpix_pixelization.envelope(region)
+                for r in pixel_range.ranges():
+                    region_pixels.extend(range(r[0], r[1]))
```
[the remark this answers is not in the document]
erykoff: Well, it doesn't want the input to be sorted (it can handle that), but it returns sorted data (which we want).  So we could use a set and then sort, or use a list and unique ... this isn't going to be a big list anyway, so I'm just going to leave it.

python/lsst/pipe/tasks/hips.py
TallJimbo: My translation of what Nate is suggesting to concrete code is:
```
(hpx_output_dimension,) = (d for d in output_dataset_type.dimensions if isinstance(d, SkyPixDimension))
```
That does express what we're trying to do a bit more clearly, which I think is a better argument in this case than fragility (this whole function will be invalidated by those dimensions changing).  Whether it's better overall depends on how much you dislike `isinstance`.  This also gets a :man_shrugging: from me.

python/lsst/pipe/tasks/hips.py
```diff
@@ -0,0 +1,545 @@
...
+        multiple=True,
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+
+        quantum_order = None
+        for dim in self.dimensions:
+            if 'healpix' in dim:
+                quantum_order = int(dim.split('healpix')[1])
```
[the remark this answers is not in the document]
erykoff: I added a check if more than one healpix dimension is specified.

python/lsst/pipe/tasks/hips.py
```diff
@@ -0,0 +1,545 @@
...
+        if quantum_order is None:
+            raise ValueError("Must specify a healpix dimension in quantum dimensions.")
+
+        if quantum_order > config.hips_order:
+            raise ValueError("Quantum healpix dimension order must not be greater than hips_order")
+
+        order = None
+        for dim in self.hips_exposures.dimensions:
```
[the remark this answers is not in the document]
erykoff: I also added a check here to make sure not more than one healpix dimension is specified.  So I'm not using your clever code here, but thanks!

python/lsst/pipe/tasks/hips.py
```diff
@@ -0,0 +1,545 @@
...
+                stats_ctrl,
+                [1.0]*len(warp_dict[pixel]),
+                clipped=0,
+                maskMap=[]
+            )
+
+        return pipeBase.Struct(hips_exposures=exp_hpx_dict)
+
+    @classmethod
+    def build_quantum_graph_cli(cls, argv):
```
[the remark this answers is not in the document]
erykoff: I moved building the parser into its own helper function and I think that makes it significantly more readable without going too far down the rabbit hole of sub-functions.

PR #685, opened 2022-05-23
DM-33959: Calibrate Source Tables with Global Calibrations

yalsayyad:
Switch configs to reEval and doApplyExternal to True

Almost every time this task will be used after DP0.2
will be to apply jointcal/FGCM calibrations to the
source tables


cmsaunders (approved):
This looks good to me.

PR #686, opened 2022-05-25
DM-34943: Set psf before measurement and fix aperture correction config.

erykoff:


python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -169,7 +169,23 @@ def setDefaults(self):
                                      'slot_Centroid_flag',
                                      'base_GaussianFlux_flag']
 
-        self.measure_ap_corr.sourceSelector['flagged'].field = 'calib_psf_used'
+        # Configure aperture correction to select only high s/n sources (that
+        # were used in the psf modeling) to avoid background problems when
+        # computing the aperture correction map.
+        self.measure_ap_corr.sourceSelector = 'science'
```
laurenam, line 175: Is this missing a `.name`?

erykoff: `pex_config` 🪄 , it works either way.

laurenam: Ah...this line followed by the one below looks very odd to the muggles 😉

TallJimbo (approved):

PR #692, opened 2022-06-23
DM-35312 Support typing of ConfigurableAction(fields)

natelust:
Support supplying Field types through python class typing syntax.
Another minor fix includes setting an identity attribute on an
action if it is accessed via a Field.


python/lsst/pipe/tasks/configurableActions/_configurableAction.py
```diff
@@ -45,5 +48,17 @@ class ConfigurableAction(Config):
...
+    `ConfigurableActionStructField` the name of the field will be bound to this
+    variable when it is retrieved.
+    """
+
+    def __setattr__(self, attr, value, at=None, label="assignment"):
+        if attr == 'identity':
+            return object.__setattr__(self, attr, value)
+        return super().__setattr__(attr, value, at, label)
+
     def __call__(self, *args, **kwargs) -> Any:
```
TallJimbo, line 63: This is preexisting, and it seems like something we might have even discussed already, but doesn't this signature imply that all `ConfigurableAction` subclasses *must* accept arbitrary `*args, **kwargs`, rather than the concrete signature declared by some intermediate `ConfigurableAction` abstract subclass?  (This is exactly the same problem as `Task.run`, but without the history that makes that one harder to remove, I think).

In any case, it seems to me to make more sense to drop the `__call__` here and define it only in subclasses like `AnalysisAction` that have a concrete signature.  If the goal here is to be able to define some concrete `ConfigurableAction` subclasses here that have *generic* `__call__` signatures becaus they're just forwarders, I think there's a better way to type that with PEP-612 stuff, though I don't know more than that.

python/lsst/pipe/tasks/configurableActions/_configurableActionField.py
```diff
@@ -22,14 +22,16 @@
...
+
+from lsst.pex.config import ConfigField, FieldValidationError, FieldTypeVar, Config
 from lsst.pex.config.config import _typeStr, _joinNamePath
 from lsst.pex.config.callStack import getCallStack
 
 from . import ConfigurableAction
 
 
-class ConfigurableActionField(ConfigField):
+class ConfigurableActionField(ConfigField[FieldTypeVar]):
```
TallJimbo, line 34: Might want to use a typevar here that has `bound=ConfigurableAction` instead, like the `ActionTypeVar` you already use elsewhere.

python/lsst/pipe/tasks/configurableActions/_configurableActionField.py
```diff
@@ -42,7 +44,13 @@ class ConfigurableActionField(ConfigField):
     # classes
     name: str
 
-    def __set__(self, instance, value, at=None, label="assignment"):
+    def __set__(
+            self,
+            instance: Config,
+            value: FieldTypeVar | type[FieldTypeVar],
+            at: Any = None,
+            label: str = "assignment"
```
TallJimbo, line 52: Indentation.

python/lsst/pipe/tasks/configurableActions/_configurableActionField.py
```diff
@@ -63,6 +71,12 @@ def __set__(self, instance, value, at=None, label="assignment"):
         history = instance._history.setdefault(self.name, [])
         history.append(("config value set", at, label))
 
+    def __get__(self, instance, owner=None) -> Any:
+        result = super().__get__(instance, owner)
+        if instance is not None:
+            result.identity = self.name  # type: ignore
+        return result
```
TallJimbo, line 78: Should use overload.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -23,14 +23,16 @@
 __all__ = ("ConfigurableActionStructField", "ConfigurableActionStruct")
 
 from types import SimpleNamespace
-from typing import Iterable, Mapping, Optional, TypeVar, Union, Type, Tuple, List, Any, Dict
+from typing import (
+    Iterable, Mapping, Optional, TypeVar, Union, Type, Tuple, List, Any, Dict, Iterator, Generic)
```
TallJimbo, line 27: Maybe I'm just too used to `black` style, but I don't think this kind of indentation and close-parens are consistent with our style guide.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -178,13 +181,15 @@ def __setattr__(self, attr: str, value: Union[ConfigurableAction, Type[Configura
                 valueInst = type(value)(__name=name, __at=at, __label=label, **value._storage)
             else:
                 valueInst = value(__name=name, __at=at, __label=label)
-            self._attrs[attr] = valueInst
+            self._attrs[attr] = valueInst  # type: ignore
```
TallJimbo, line 184: Please provide a comment for the `type: ignore` (or consider `typing.cast`, as it's often self-explanatory in a way that `type: ignore` is not).

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -271,14 +283,18 @@ def __set__(self, instance: Config,
...
+        label='default'
+    ) -> ConfigurableActionStruct[ActionTypeVar]:
         if instance is None or not isinstance(instance, Config):
-            return self
+            return self  # type: ignore
         else:
-            field: Optional[ConfigurableActionStruct] = instance._storage[self.name]
-            return field
+            field: Optional[ConfigurableActionStruct] = instance._storage[self.name]  # type: ignore
+            return field  # type: ignore
```
TallJimbo, line 297: Looks like another case where you need overrides, and that might fix your `type: ignore`s.

TallJimbo (approved):
Edit: this overview blurb is mostly about the [pex_config|https://github.com/lsst/pex_config/pull/79] PR; I just put it here by accident, but now that it's out in the world, it's not worth moving.

Lots of minor comments (maybe about needing more `__get__` overloading and bad indentation - am I confused about whether `black` is being run on `pex_config` or something?).

My main concern from a big-picture standpoint is that  there's no `py.typed` file, no `MYPYPATH`-setting in the EUPS table file, and more importantly no `mypy.ini`, github action or even any expectation that running mypy here would yield anything other than a ton of errors.  So I think it's just people who have pylance configured a certain way who might benefit from the static type checking this enables, and I bet the cardinality of that set of people is of order unity.

So, that's not a blocker for merging, and I don't think expanding this ticket to include even really permissive mypy checking is a good idea either.  And you've got tests that the runtime side of this works, and that helps a lot.  But the static typing here is going to be super fragile as long as your editor is the only thing using it, and you should be mentally prepared for any typing that depends on it to be frequently broken (well, about as frequently as pex_config is modified at all, which of course isn't that frequent).

TallJimbo (approved):
Accidentally put my big-picture comments on the pipe_task PR; please see that for the overview.

python/lsst/pipe/tasks/configurableActions/_configurableAction.py
natelust: I will look more into PEP-612 stuff before merging this but AFAICT type checkers have no problem with this from a substitution principal, and I am not sure why. I dont know if they treat `__` methods as special, or if `__call__` is special because it is used in construction in so many parts of the language.

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
natelust: You are correct and I meant to fix this. I had turned off black while working on this for fear of hitting all of pipe_tasks, and it shows how much I was used to it doing things already.

PR #695, opened 2022-07-05
DM-35386 Fix Serialization of ConfigurableActionStructField

natelust:
Fields were not being properly initialized empty when serialized.
This caused state from a defaunt instatiation of a class to add
state back into a de-serialized object when it should not be
there.

### DIFF
--- a/python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
+++ b/python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
@@ -220,6 +220,9 @@ class ConfigurableActionStruct(Generic[ActionTypeVar]):
         for name in self.fieldNames:
             yield name, getattr(self, name)
 
+    def __bool__(self) -> bool:
+        return bool(self._attrs)
+
 
 T = TypeVar("T", bound="ConfigurableActionStructField")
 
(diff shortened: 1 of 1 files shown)


laurenam:
Very nit-picky, but the PR title should have a `:` after the ticket number.  Also, there are a few typos in your commit message:
* defaunt instatiation -> default instantiation

I don't pretend to fully understand the fix (e.g. I don't know why the `def __bool__` was necessary), but this has indeed fixed my problem as my Jenkins run on DM-34908 with this patch cherry picked has passed `ci_hsc`, so marking as reviewed 😄

python/lsst/pipe/tasks/configurableActions/_configurableActionStructField.py
```diff
@@ -355,8 +358,11 @@ def toDict(self, instance):
     def save(self, outfile, instance):
         actionStruct = self.__get__(instance)
         fullname = _joinNamePath(instance._name, self.name)
+
+        # ensure that a struct is always empty before assigning to it
```
laurenam, line 362: Another nit-pick: Comments are supposed to start capitalized.

laurenam (approved):

PR #698, opened 2022-07-12
DM-35496: Update coord columns per reevaluated WCS

yalsayyad:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -673,6 +673,9 @@ def addCalibColumns(self, catalog, exposure, exposureIdInfo, **kwargs):
         newCat = afwTable.SourceCatalog(schema)
         newCat.extend(catalog, mapper=mapper)
 
+        if self.config.doReevaluateSkyWcs:
```
cmsaunders, line 676: It took me a little while to track down the difference between `doApplyExternalSkyWcs` and `doReevaluateSkyWcs`--would it be worth clarifying in the docs that the former updates the wcs of the exposure while the latter updates the coordinates of the input catalog, if I am interpreting correctly?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -673,6 +673,9 @@ def addCalibColumns(self, catalog, exposure, exposureIdInfo, **kwargs):
...
+            afwTable.updateSourceCoords(exposure.wcs, newCat)
```
cmsaunders, line 677: Do you need to have a corresponding statement for `self.config.doReevaluatePhotoCalib`?

cmsaunders (approved):
I have a couple questions, but I'm not sure how valid they are.

python/lsst/pipe/tasks/postprocess.py
yalsayyad: *  `doReevaluateSkyWcs` tells it  whether or not to update the WCS in the source table at all.
*  IFF that's true, `doApplyExternalSkyWcs` tells it whether to use `calexp.wcs` or `jointcal`. 
* IFF that's that's true, `useGlobalExternalSkyWcs` tells it whether to use jointcal-global or jointcal-tract. 

 `doApplyExternalSkyWcs`+`useGlobalExternalSkyWcs` here means the thing as all the other Tasks that decide whether to get their WCS from the calexp, jointcal tract-mode, or jointcal global-mode (i.e. `makeWarp`, `ForcedPhotCcd` and faro).   I find the control structure that you find in every[ `__init__` of  Configs that have a  `doApplyExternalSkyWcs`](https://github.com/lsst/pipe_tasks/blob/0780a3ce30710804fe8592868aa4df8795e4bf09/python/lsst/pipe/tasks/makeCoaddTempExp.py#L718) super confusing, hate it, and hope that one day when you make `gbdes` an option that you find a good replacement for  `doApplyExternalSkyWcs`+`useGlobalExternalSkyWcs` everywhere. 
 
I opted for consistency. 

to help, I'll add a line to the docstring of `doApplyExternalSkyWcs` "Used if and only if doReevaluateSkyWcs"

PR #699, opened 2022-07-13
DM-35551: fix ConfigurableAction tests and make serialization deterministic

TallJimbo:


erykoff (approved):
Looks fine as far as I can tell, but I'm curious what was wrong with the old test configs?

PR #703, opened 2022-07-26
DM-35035: Remove gen2 curated calibration code

timj:
The core reader code has been moved to lsst/obs_base#424


erykoff (approved):
Looks good.  Sorry I missed that gen2 test previously!

erykoff:
One more question I just realized: this code used to be a public interface in pipe_tasks but is now private in obs_base.  Is there a need to do a deprecated forwarding here?

timj:
read_all was public because obs_base needed to call it. The real public interface was the gen2 ingestCuratedCalibs.py. I was pondering whether `read_all` should be public in obs_base to allow Instrument subclasses to call it directly but I decided not to unless we need it because the actual curated calibrations interface in `Instrument` is much better. If we want this code to hang around for a deprecation cycle I should bring it back rather than calling obs_base because the interface has changed in obs_base. Given there's a tiny chance that someone has used `read_all` outside of a butler context (Which is tiny since they also need the corresponding obs_*_data package layout) then I'll bring it back and add a deprecation.

timj:
I've added the deprecation message.

PR #706, opened 2022-07-28
DM-34689: Add GenerateHipsTask.

erykoff:


python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+        optional=False,
+    )
+    spectral_ranges = pexConfig.ConfigDictField(
+        doc=("Mapping from band to lambda_min, lamba_max (nm).  May be approximate."),
+        keytype=str,
+        itemtype=HipsPropertySpectralTerm,
+        default={},
+    )
+    initial_ra = pexConfig.Field(
+        doc="Initial RA (deg).  If not set will use a point in MOC.",
```
timj, line 667: Is this "initial" as in a hint to the position at which a HiPS viewer should center the display?

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+        dtype=float,
+        optional=True,
+    )
+    obs_ack = pexConfig.Field(
+        doc="Observation acknowledgements (free text).",
+        dtype=str,
+        optional=True,
+    )
+    t_min = pexConfig.Field(
+        doc="Start time (MJD) of observations included in HiPS",
```
timj, line 687: "Time of earliest observation included in HiPS" ?

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+        dtype=float,
+        optional=True,
+    )
+    t_max = pexConfig.Field(
+        doc="Stop time (MJD) of observations included in HiPS",
```
timj, line 692: "Time of most recent [latest] observation included in HiPS"?

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+    )
+    png_gray_asinh_softening = pexConfig.Field(
+        doc="AsinhMapping softening parameter (Q) for grayscale png scaling.",
+        dtype=float,
+        default=8.0,
+    )
+
+
+class GenerateHipsTask(pipeBase.PipelineTask):
+    """Task for making a HiPS tree with fits and grayscale pngs."""
```
timj, line 786: FITS and PNGs.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+            (hips_exposure_handle.dataId[healpix_dim],
+             hips_exposure_handle.dataId['band']): hips_exposure_handle
+            for hips_exposure_handle in inputs["hips_exposure_handles"]
+        }
+
+        band = butlerQC.quantum.dataId['band']
+
+        self.run(bands=[band], max_order=order, hips_exposure_handle_dict=hips_exposure_handle_dict)
+
+    def run(self, bands, max_order, hips_exposure_handle_dict, do_color=False):
```
timj, line 813: Should we have a separate timeMethod around the run?

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+    @timeMethod
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+
+        dims = inputRefs.hips_exposure_handles[0].dataId.names
+        order = None
+        for dim in dims:
+            if "healpix" in dim:
+                order = int(dim.split("healpix")[1])
+                healpix_dim = dim
```
timj, line 799: Do you mean for this loop to continue going after a match has been found? Are you expecting dims to have different orders? It doesn't seem like it's trying to find the highest or lowest order.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+        if order is None:
```
timj, line 800: If the loop does `break` this can be an `else`.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+            raise RuntimeError("Could not determine healpix order for input exposures.")
+
+        hips_exposure_handle_dict = {
+            (hips_exposure_handle.dataId[healpix_dim],
+             hips_exposure_handle.dataId['band']): hips_exposure_handle
+            for hips_exposure_handle in inputs["hips_exposure_handles"]
+        }
+
+        band = butlerQC.quantum.dataId['band']
```
timj, line 809: Please add a note here that you require a single band (but you do not actually check you only have one band) but subclasses will use multiple bands but share the `run()`. I was confused until I saw the subclass at the end.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+    def run(self, bands, max_order, hips_exposure_handle_dict, do_color=False):
+        """Run the GenerateHipsTask.
+
+        Parameters
+        ----------
+        bands : `list [ `str` ]
+            List of bands to be processed (or single band).
+        max_order : `int`
+            HEALPix order of the maximum (native) HPX exposures.
+        hips_exposure_handle_dict : `dict` {`int`: `lsst.daf.butler.DeferredDatasetHandle`}
```
timj, line 822: This type annotation is not right but you likely don't have space to make it correct.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+        for order in range(min_order, max_order + 1):
+            pixels_shifted[order][-1] = -1
+
+        # Read in the first pixel for determining image properties
+        exp0 = hips_exposure_handle_dict[list(hips_exposure_handle_dict.keys())[0]].get()
+        bbox = exp0.getBBox()
+        npix = bbox.getWidth()
+        shift_order = int(np.round(np.log2(npix)))
+
+        # Create blank exposures for each level, including the highest order
```
timj, line 874: Pedantically these comments should all have periods at the end of sentences... Some do, some don't.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+                Q=self.config.png_color_asinh_softening,
+            )
+
+            bcb = self.config.blue_channel_band
+            gcb = self.config.green_channel_band
+            rcb = self.config.red_channel_band
+            colorstr = f'{bcb}{gcb}{rcb}'
+
+        # The base path is the hips_base_uri with the band as a sub-directory.
+        hips_base_path = ResourcePath(self.config.hips_base_uri, forceDirectory=True)
```
timj, line 849: The comment says the base path should have the band included but then uses a variable called base_path that doesn't include the band.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+        pixels_shifted[max_order] = pixels
+        for order in range(max_order - 1, min_order - 1, -1):
+            pixels_shifted[order] = np.right_shift(pixels_shifted[order + 1], 2)
+
+        # And set the gutter to an illegal pixel value
+        for order in range(min_order, max_order + 1):
+            pixels_shifted[order][-1] = -1
+
+        # Read in the first pixel for determining image properties
+        exp0 = hips_exposure_handle_dict[list(hips_exposure_handle_dict.keys())[0]].get()
```
timj, line 869:
```python
exp0 = list(hips_exposure_handle_dict.values())[0].get()
```
or
```python
exp0 = next(iter(hips_exposure_handle_dict.values())).get()
```

are maybe shorter alternatives?

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+                exp.image.array[:, :] = np.nan
+                exposures[(band, order)] = exp
+
+        # Loop over all pixels, avoiding the gutter.
+        for pixel_counter, pixel in enumerate(pixels[:-1]):
+            self.log.debug("Working on high resolution pixel %d", pixel)
+            for band in bands:
+                # Read all the exposures here for the highest order.
+                if (pixel, band) in hips_exposure_handle_dict:
+                    exposures[(band, max_order)] = hips_exposure_handle_dict[(pixel, band)].get()
```
timj, line 890: `band, max_order` vs `pixel, band`  ordering?

Is this max_order by definition because the healpix generation code only generates images at the highest order? Therefore it's not possible `pixel != max_order`. Can we have a comment for that?

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+                        pixels_shifted[order][pixel_counter],
+                        exposures[(self.config.red_channel_band, order)].image,
+                        exposures[(self.config.green_channel_band, order)].image,
+                        exposures[(self.config.blue_channel_band, order)].image,
+                        png_color_mapping,
+                    )
+
+                if order == (max_order - 3):
+                    # This is a reasonable amount of INFO logging density.
+                    self.log.info(
```
timj, line 927: If you want to reduce duplication of log message content you can do

```python
log_level = self.log.INFO if order == (max_order -3) else self.log.DEBUG
self.log.log(log_level, "Completed HiPS generation...", ...)
```

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+                        exp.image.array[0: npix//2, npix//2:] = binned_image_arr
+                    else:
+                        # This should be impossible.
+                        raise ValueError("Illegal pixel sub index")
+
+                    # Erase the previous exposure.
+                    if order < max_order:
+                        exposures[(band, order)].image.array[:, :] = np.nan
+
+        # Write the property files and mocs
```
timj, line 983: "MOCs"

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+            Input for blue channel of output png.
+        png_mapping : `astropy.visualization.lupton_rgb.AsinhMapping`
+            Mapping to convert image to scaled png.
+        """
+        # WARNING: In general PipelineTasks are not allowed to do any outputs
+        # outside of the butler.  This task has been given (temporary)
+        # Special Dispensation because of the nature of HiPS outputs until
+        # a more controlled solution can be found.
+
+        dir_number = (pixel//10000)*10000
```
timj, line 1099: Consider factoring out this hips_dir, dir_number calculation since it is reused and is required to match.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+            multiband,
+            exposure,
+            max_order,
+            shift_order,
+            area,
+            initial_ra,
+            initial_dec,
+            initial_fov
+    ):
+        """Write HiPS properties file.
```
timj, line 1213: The method says "property file" but this says "properties file"

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+                Open for writing.
+            name : `str`
+                Name of property
+            value : `str`
+                Value of property
+            """
+            for char in name:
+                if char.isspace():
+                    raise ValueError(f"``{name}`` cannot contain any space characters.")
+            if '=' in name:
```
timj, line 1260: File seems mostly double quotes so this should be too (and black uses double quotes)

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+            Parameters
+            ----------
+            fh : file handle (blah)
+                Open for writing.
+            name : `str`
+                Name of property
+            value : `str`
+                Value of property
+            """
+            for char in name:
```
timj, line 1257: Checking for unicode spaces?

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+                        fh,
+                        "obs_description",
+                        property_config.obs_description_template.format(band=band),
+                    )
+                if len(property_config.prov_progenitor) > 0:
+                    for prov_progenitor in property_config.prov_progenitor:
+                        _write_property(fh, "prov_progenitor", prov_progenitor)
+                if property_config.obs_ack is not None:
+                    _write_property(fh, "obs_ack", property_config.obs_ack)
+                _write_property(fh, "obs_regime", "Optical")
```
timj, line 1301: Would this chunk of code look easier to read if you constructed a dict and then had `_write_property_dict()` ? Or did `for k, v in property_dict: _write_property(fh, k, v)`? That could also allow the dict to be constructed outside of the `with` block.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +614,907 @@ def build_quantum_graph(
...
+class GenerateColorHipsTask(GenerateHipsTask):
+    """Task for making a HiPS tree with color pngs."""
+    ConfigClass = GenerateColorHipsConfig
+    _DefaultName = "generateColorHips"
+
+    @timeMethod
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
+        inputs = butlerQC.get(inputRefs)
+
+        dims = inputRefs.hips_exposure_handles[0].dataId.names
```
timj, line 1474: I'd rather this code duplication was removed. It seems to be the same as the parent class runQuantum until the data_bands turns up and even that code could be shared with the parent class if the parent class runQuantum then checked that there was only one band in data_bands and it matched the quantum dataId.

timj (approved):
Looks good. I admit that I skipped over some of the specifics of how the hips images were constructed.

The lack of unit tests is concerning -- even a test for the property file creation (which might need code to parse it back to compare) would be something.

python/lsst/pipe/tasks/hips.py
erykoff: You're right, it should break.

python/lsst/pipe/tasks/hips.py
erykoff: I did not know that.  Huh!

erykoff: Wait, this is the opposite ... it is supposed to break.

python/lsst/pipe/tasks/hips.py
erykoff: I added this comment:
```
                # There will always be at least one band with a HiPS image available
                # at the highest order. However, for color images it is possible that
                # not all bands have coverage so we require this check.
```

python/lsst/pipe/tasks/hips.py
timj: That's my point. If you break out the else won't trigger. `else` only runs if the loop completes.

python/lsst/pipe/tasks/hips.py
erykoff: spaces, tabs, anything space-like is NOT ALLOWED by the HiPS standard.

timj: Can you say:

```python
import re

if re.search(r"\s", name):
    raise ValueError()
```
rather than looping around? I'd definitely do it as a regex.

python/lsst/pipe/tasks/hips.py
erykoff: The `prov_progenitor` property in particular is allowed to have multiple entries (and there are others) so that makes handling the dict more complicated.  I guess it could have a value that is a str or a list, but then the code simplification gain is much smaller.

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +615,1002 @@ def build_quantum_graph(
...
+        allsky_order_uri = hips_base_path.join(f'Norder{allsky_order}', forceDirectory=True)
+        png_uris = list(
+            ResourcePath.findFileResources(
+                candidates=[allsky_order_uri],
+                file_filter=re.compile(r'Npix.*\.png'),
+            )
+        )
+
+        for png_uri in png_uris:
+            matches = re.search("Npix([0-9]+).png", png_uri.geturl())
```
timj, line 1484: It seems like this regex could match the `file_filter` regex above to guarantee that you won't have a situation where `matches` doesn't match. In theory you should have a `$` at the end of this regex and if you use `png_uri.basename()` rather than `geturl` you can make it even more specific.

```python
matches = re.match("Npix([0-9]+).png$", png_uri.basename())
```

python/lsst/pipe/tasks/hips.py
```diff
@@ -604,3 +616,1002 @@ def build_quantum_graph(
...
+        allsky_order_uri = hips_base_path.join(f"Norder{allsky_order}", forceDirectory=True)
+        png_uris = list(
+            ResourcePath.findFileResources(
+                candidates=[allsky_order_uri],
+                file_filter=re.compile(r"Npix.*\.png"),
+            )
+        )
+
+        for png_uri in png_uris:
+            matches = re.match("Npix([0-9]+).png$", png_uri.basename())
```
timj, line 1485: I will reiterate that the `.` here matches any character when pedantically you want `\.` -- which happens to be in the regex 5 lines above.

PR #711, opened 2022-08-05
DM-971: fix errors now that tasks always validate on init

parejkoj:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -562,6 +562,10 @@ def addCalibColumns(self, catalog, exposure, exposureIdInfo, **kwargs):
         measureConfig = SingleFrameMeasurementTask.ConfigClass()
         measureConfig.doReplaceWithNoise = False
 
+        # Clear all slots, because we aren't running the relevant plugins.
+        for slot in measureConfig.slots:
+            setattr(measureConfig.slots, slot, None)
+
```
erykoff, line 568: Do we not need any of the slots at all?  I guess this is just the recalibration plugins, so no.  And if it runs in ci_hsc then it's fine, I suppose!

tests/test_calibrate.py
```diff
@@ -170,7 +170,9 @@ def _checkDoRefcats(self, doAstrometry, doPhotoCal, ids):
         config.doAstrometry = doAstrometry
         config.doPhotoCal = doPhotoCal
         config.connections.photoRefCat = "cal_ref_cat"
+        config.photoRefObjLoader.ref_dataset_name = "cal_ref_cat"
```
erykoff, line 173: Why is this necessary?  Isn't this usage deprecated?

tests/test_interpImageTask.py
```diff
@@ -122,6 +122,7 @@ def validateInterp(miInterp, useFallbackValueAtEdge, fallbackValue):
                         # Should raise if negative not allowed, but USER supplied negative value
                         if not negativeFallbackAllowed and fallbackValueType == "USER":
                             self.assertRaises(ValueError, config.validate)
+                            continue
```
erykoff, line 125: How did this work before?

erykoff (changes requested):
A question about setting deprecated configs...

python/lsst/pipe/tasks/postprocess.py
parejkoj: These slots are what the plugins use to define what gets assigned where, but if the input catalog already has slots, these aren't necessary. And yes, none of them are needed because none of the relevant plugins are being run.

tests/test_calibrate.py
parejkoj: I think until I merge [DM-35871](https://jira.lsstcorp.org/browse/DM-35871) we might need still need this for how colorterms are used. I'll deal with it on that ticket.

tests/test_interpImageTask.py
parejkoj: Since the validation wasn't being run before, it just continued on into the rest of that iteration with a non-valid config. I don't know what, if any, failure modes it might have had, but clearly it didn't fail.

erykoff (approved):

PR #715, opened 2022-08-12
DM-19374: More Gen2 cleanup and also eups table fixups

timj:
There are still some gen2 references in the code but I don't know enough to remove them.


erykoff (approved):
Looks great; thanks for the doc cleanup, I wasn't sure what to put for the boilerplate to replace command-line-task info (I've been asking for a template for ages).  I guess this is it?

And the import test is a great new addition!

PR #716, opened 2022-08-13
DM-35871: Add refcat name kwarg to ReferenceObjectLoader init

parejkoj:


python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -93,17 +93,18 @@ class LoadReferenceCatalogTask(pipeBase.Task):
     ConfigClass = LoadReferenceCatalogConfig
     _DefaultName = "loadReferenceCatalog"
 
-    def __init__(self, dataIds=None, refCats=None, **kwargs):
-        if dataIds is not None and refCats is not None:
+    def __init__(self, dataIds=None, refCats=None, name=None, **kwargs):
+        if dataIds is not None and refCats is not None and name is not None:
```
erykoff, line 97: I've waffled on this; I think that name can be done if `doApplyColorTerms` is `False`, but this would require capturing the `config` from the `kwargs`, and then instantiating the default `ConfigClass` if it is `None` which is perhaps more trouble than its worth.  So I think I'm okay with requiring the `name` for this task where it is going to be needed almost all the time.

But ... please update the docstring above to (a) remove Gen3 references, and (b) add the `name` keyword and (c) make it clear these are all required.  (E.g., they say `optional` but they really aren't).

erykoff (approved):
Some docstring and error message changes requested.

python/lsst/pipe/tasks/loadReferenceCatalog.py
parejkoj: I'll just make them all required named arguments and get rid of the `if: raise` block, since that was mostly for gen2/gen3 use, I think.

erykoff:
On DM-971 we noticed that the `ref_dataset_name` is still be used in the calibrate config validation, when it shouldn't be.  So please remove that validation check, as well as removing the use of the deprecated config in `test_calibrate.py`

erykoff (approved):

PR #722, opened 2022-09-15
DM-36198: Allow fakes propagation in postprocessing

kfindeisen:
This PR adds the `calexpType` template to the outputs of `MakeCcdVisitTableTask` and `MakeVisitTableTask` (they were added to inputs on 80aa66dc92c83c1e284b116ae07675d6559376d3). This makes it possible run both fakes and non-fakes versions of these tasks in the same pipeline, though the only difference in output should be a few extra columns in the fakes version.

### DIFF
--- a/python/lsst/pipe/tasks/postprocess.py
+++ b/python/lsst/pipe/tasks/postprocess.py
@@ -1456,7 +1456,7 @@ class MakeCcdVisitTableConnections(pipeBase.PipelineTaskConnections,
     )
     outputCatalog = connectionTypes.Output(
         doc="CCD and Visit metadata table",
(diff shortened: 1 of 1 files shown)


parejkoj (approved):
This is making me wish that the `defaultTemplate`s had docstrings: if I hadn't seen your commit message, I wouldn't have known this was for "fakes_". Not sure there's a good way to do that, though.

kfindeisen:
I think the original author wanted to specifically avoid using the word "fakes" because of the whole argument over what they should be called. As for how to do it, we could always start adding docstrings to `*Connections` classes...

PR #724, opened 2022-10-03
DM-36440: Add EpochPropertyMap to track observing epoch in coadds.

erykoff:
This also fixes the high coverage nside warnings from healsparse which were not fully suppressed.


arunkannawadi (approved):
LGTM, and seems to do what we discussed at the Monday Morning Meeting today.

PR #727, opened 2022-10-18
DM-36571: Remove applyColorTerms=None option from PhotoCalTask and default to False

parejkoj:


python/lsst/pipe/tasks/photoCal.py
```diff
@@ -52,13 +52,11 @@ class PhotoCalConfig(pexConf.Config):
...
-        doc=("Apply photometric color terms to reference stars? One of:\n"
-             "None: apply if colorterms and photoCatName are not None;\n"
-             "      fail if color term data is not available for the specified ref catalog and filter.\n"
-             "True: always apply colorterms; fail if color term data is not available for the\n"
-             "      specified reference catalog and filter.\n"
-             "False: do not apply."),
+        default=False,
+        doc=("Apply photometric color terms to reference stars?\n"
+             "True: always apply colorterms; fail if color term data is not available for the "
+             "specified reference catalog and filter.\n"
```
leeskelvin, line 58: This doc string line would be 120 characters long. Would it look nicer to maintain the indented look (indenting to match `True: ` above) as before?

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -52,13 +52,11 @@ class PhotoCalConfig(pexConf.Config):
...
-        default=None,
-        doc=("Apply photometric color terms to reference stars? One of:\n"
-             "None: apply if colorterms and photoCatName are not None;\n"
-             "      fail if color term data is not available for the specified ref catalog and filter.\n"
-             "True: always apply colorterms; fail if color term data is not available for the\n"
-             "      specified reference catalog and filter.\n"
-             "False: do not apply."),
+        default=False,
+        doc=("Apply photometric color terms to reference stars?\n"
+             "True: always apply colorterms; fail if color term data is not available for the "
```
leeskelvin, line 57: Perhaps this would read more true if written:
```
True: attempt to apply color terms; fail if color term data...
```

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -84,13 +82,13 @@ class PhotoCalConfig(pexConf.Config):
     )
     colorterms = pexConf.ConfigField(
         dtype=ColortermLibrary,
-        doc="Library of photometric reference catalog name: color term dict",
+        doc="Library of photometric reference catalog name: color term dict, see also applyColorTerms.",
```
leeskelvin, line 85: Maybe `(see also applyColorTerms)` instead?

python/lsst/pipe/tasks/photoCal.py
```diff
@@ -253,19 +251,9 @@ def extractMagArrays(self, matches, filterLabel, sourceKeys):
...
-            photoCatSpecified = self.config.photoCatName is not None
-            applyCTReason += " and data %s available" % ("is" if ctDataAvail else "is not")
-            applyCTReason += " and photoRefCat %s provided" % ("is" if photoCatSpecified else "is not")
-            applyColorTerms = ctDataAvail and photoCatSpecified
-
-        if applyColorTerms:
-            self.log.info("Applying color terms for filter=%r, config.photoCatName=%s because %s",
-                          filterLabel.physicalLabel, self.config.photoCatName, applyCTReason)
+        if self.config.applyColorTerms:
+            self.log.info("Applying color terms for filter=%r, config.photoCatName=%s,",
```
leeskelvin, line 255: Superfluous comma at the end.

python/lsst/pipe/tasks/photoCal.py
parejkoj: I've taken out the "always"; the "attempt" is implied by the failure description.

python/lsst/pipe/tasks/photoCal.py
parejkoj: This spawned a whole conversation in #dm-docs-support. The challenge is balancing the `config.save` output vs. the rendered HTML that goes in pipelines.lsst.io. I think I've found a good balance of those here.
https://lsstc.slack.com/archives/C2B6DQBAL/p1667327052273339

parejkoj:
@leeskelvin : what do you think about this reworked version? I think the rendered HTML and `config.save` versions both look ok.

python/lsst/pipe/tasks/photoCal.py
leeskelvin: Had you previously removed the `always` in the `True` section here? If so, it looks like it has reappeared. 

I still think it would read better if written as `attempt to apply color terms; fail if...`. If instead you simply write `apply color terms; fail if...`, that implies that you a) are applying color terms, but then b) are failing to apply color terms, because none are available. Logically, this doesn't make sense to me, hence the need for an admission of an `attempt` being made.

Follow up point: use of `colorterms` vs `color terms`. I think the former should be reserved for reference to the config parameter, whilst the latter reserved for discussion of "color terms". I think this sentence probably should make use of `color terms` (with a space), but happy to defer to you if you were instead meaning to refer to the config parameter.

python/lsst/pipe/tasks/photoCal.py
leeskelvin: Thanks - as mentioned in a reply above, I think your edit was undone by a subsequent edit. For the reasons elucidated above, I still prefer the use of `attempt to` here, but am happy to defer to your judgement on this.

leeskelvin:
Thanks John, this is looking good. I left a comment on the new doc string format, but otherwise I think this looks good to me, cheers!

python/lsst/pipe/tasks/photoCal.py
parejkoj: Ok, I think I see what you mean. I've taken that wording.

PR #734, opened 2022-12-07
Add a lookup function for brightObjectMasks

arunkannawadi:


python/lsst/pipe/tasks/assembleCoadd.py
```diff
@@ -51,6 +52,25 @@
 log = logging.getLogger(__name__)
 
 
+def _lookupDataset(datasetType, registry, quantumDataId, collections) -> list[DatasetRef]:
```
arunkannawadi, line 55: This function is generic enough, that I could see living in `lsst.utils` and importing from there.

TallJimbo: I think I'd prefer to leave it here and not advertise, because I'd like to get this fixed in pipe_base before it happens again.  I've created [DM-73286](https://jira.lsstcorp.org/browse/DM-37286) for that.

TallJimbo (approved):
If we're going to have a warning for this, it'd probably be better for it to be emitted by `run` or `runQuantum` rather than QG generation, but I don't have a terribly strong opinion about that.

PR #736, opened 2022-12-09
DM-37242: Refactor SkyCorrectionTask

leeskelvin:


python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -183,64 +106,100 @@ class SkyCorrectionConnections(pipeBase.PipelineTaskConnections, dimensions=("in
...
+    )
+    bgModel = ConfigField(
+        dtype=FocalPlaneBackgroundConfig,
+        doc="Initial background model, prior to sky frame subtraction",
+    )
+    doBgModel = Field(
+        dtype=bool,
+        default=True,
+        doc="Do initial background model subtraction (prior to sky frame subtraction)?",
+        deprecated="This field is deprecated and will be removed after v26. See RFC-898 for further details.",
```
arunkannawadi, line 178: removed after v26, since v25 candidate is already tagged. Also, if this is just part of the renaming, suggest the field that takes its place.

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -183,64 +106,100 @@ class SkyCorrectionConnections(pipeBase.PipelineTaskConnections, dimensions=("in
...
+    maskObjects = ConfigurableField(
+        target=MaskObjectsTask,
+        doc="Mask Objects",
+    )
+    doMaskObjects = Field(
+        dtype=bool,
+        default=True,
+        doc="Iteratively mask objects to find good sky?",
+    )
+    bgModel = ConfigField(
```
arunkannawadi, line 170: If you are going to introduce `bgModel1`, then this should be deprecated and set `optional=True`.

python/lsst/pipe/tasks/skyCorrection.py
leeskelvin: Here's a snippet from [the Dev Guide](https://developer.lsst.io/stack/deprecating-interfaces.html#code-removal):

> Removal of the deprecated code occurs, at the earliest, immediately after the next major release following the release with the first deprecation release note; at the latest, immediately before the major release following. In other words, if deprecation is first noted in release 17.2.3, the code cannot be removed until after 18.0 is released and must be removed before 19.0 is released.

To my mind, v25 hasn't been fully released yet (it's still only a release candidate), but assuming that still counts, that means that the next major release is v26. This means that this deprecation effort should take place in the v26 to v27 time period, no?

python/lsst/pipe/tasks/skyCorrection.py
leeskelvin: As discussed via Slack, there is no `optional` argument for a `ConfigField`. Instead, I have set the `dtype` for this `ConfigField` to the base `Config` class (to allow `validate` to pass successfully) and left comments informing the user that this field should not be used - please use `bgModel1` instead.

python/lsst/pipe/tasks/skyCorrection.py
arunkannawadi: Right, after v26 is correct then. Apologies.

arunkannawadi: Also, it looks like both `doBgModel` and `doBgModel1` are deprecated. That's surely not what you intended.

arunkannawadi (approved):
LGTM. Not really a show stopped, but this could have two commits - one for the actual changes, and one for formatting the rest of the code.

python/lsst/pipe/tasks/skyCorrection.py
leeskelvin: Following [RFC-898](https://jira.lsstcorp.org/browse/RFC-898), the first full focal plane background model will be made mandatory. That means that `doBgModel` will be removed on the implementation ticket. However, as we're also changing the name of the task that it targets (`bgModel`) that means that I also need to temporarily set up a new `bgModel1` config field for use from now until the deprecation is implemented. At the time of implementation, I'll remove both `doBgModel` and `doBgModel1` (and make the relevant code block mandatory).

I can't think of a better way to do this, but if you can, happy to switch it up!

leeskelvin:
Thanks Arun - Re: splitting commits, normally I'd agree, but in this case the act of refactoring is what led to any new features being changed. I'd find it difficult right now to unpick this effort in short order, so if it's okay with you, I'll leave it on one commit as-is.

python/lsst/pipe/tasks/skyCorrection.py
arunkannawadi: Either way, I can't see a way to not raise `DeprecatedWarning` messages. So this is fine.

PR #738, opened 2022-12-19
DM-37412: move and refactor some exposure summary stats code

TallJimbo:


python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -316,13 +252,87 @@ def _computePsfStats(self, sources):
...
+        summary.psfStarDeltaE1Median = float(psfStarDeltaE1Median)
+        summary.psfStarDeltaE2Median = float(psfStarDeltaE2Median)
+        summary.psfStarDeltaE1Scatter = float(psfStarDeltaE1Scatter)
+        summary.psfStarDeltaE2Scatter = float(psfStarDeltaE2Scatter)
+        summary.psfStarDeltaSizeMedian = float(psfStarDeltaSizeMedian)
+        summary.psfStarDeltaSizeScatter = float(psfStarDeltaSizeScatter)
+        summary.psfStarScaledDeltaSizeScatter = float(psfStarScaledDeltaSizeScatter)
 
-        return psfStats
+    def update_wcs(self, summary, wcs, bbox, visitInfo):
```
laurenam, line 264: Ditto on adding another descriptor to the function name (e.g. `update_wcs_stats`)?

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -121,77 +121,24 @@ def run(self, exposure, sources, background):
...
+        photoCalib = exposure.getPhotoCalib()
+        self.update_photo_calib(summary, photoCalib)
 
         if background is not None:
             bgStats = (bg[0].getStatsImage().getImage().array
                        for bg in background)
-            skyBg = sum(np.median(bg[np.isfinite(bg)]) for bg in bgStats)
-        else:
-            skyBg = np.nan
+            summary.skyBg = sum(np.median(bg[np.isfinite(bg)]) for bg in bgStats)
```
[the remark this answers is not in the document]
TallJimbo: The initial default-construction of `summary = ExposureSummaryStats()` now initializes all of these to `None`.  In the other methods I've set things to `nan` explicitly because they might be called on objects that weren't just initialized that way (indeed, that's exactly what I'm planning to do on DM-35207), but here that's now unnecessary.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -202,94 +149,83 @@ def run(self, exposure, sources, background):
...
-            psfStarDeltaE1Median=psfStats.psfStarDeltaE1Median,
-            psfStarDeltaE2Median=psfStats.psfStarDeltaE2Median,
-            psfStarDeltaE1Scatter=psfStats.psfStarDeltaE1Scatter,
-            psfStarDeltaE2Scatter=psfStats.psfStarDeltaE2Scatter,
-            psfStarDeltaSizeMedian=psfStats.psfStarDeltaSizeMedian,
-            psfStarDeltaSizeScatter=psfStats.psfStarDeltaSizeScatter,
-            psfStarScaledDeltaSizeScatter=psfStats.psfStarScaledDeltaSizeScatter
-        )
+            summary.astromOffsetMean = md['SFM_ASTROM_OFFSET_MEAN']
+            summary.astromOffsetStd = md['SFM_ASTROM_OFFSET_STD']
```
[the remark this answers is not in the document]
TallJimbo: See previous.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -202,94 +149,83 @@ def run(self, exposure, sources, background):
...
 
         return summary
 
-    def _computePsfStats(self, sources):
-        """Compute psf residual statistics.
-
-        All residuals are computed using median statistics on the difference
-        between the sources and the models.
+    def update_psf(self, summary, psf, bbox, sources=None):
```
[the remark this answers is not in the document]
TallJimbo: Will rename to `update_psf_stats` (etc).

I don't have a problem with some amount of case-convention inconsistency within a file, and I think the developer guide provides a lot of leeway on that.  These days I tend to write pretty much any new interface as snake_case, as it's the only way to push us along on that transition even a tiny amount.

Requiring an `exposure` here would indeed be a bit of a problem, as I was hoping not to have to open up the `calexp` at all in the final visit summary stage.  But there's nothing fundamental there; it's just more I/O.  Maybe we limit it to the mask plane (if image plane or variance plane usage in the snippet on DM-36930 adds anything, it means something is wrong with our masking).  Then we could take just the `mask` as an extra parameter to `update_psf_stats`; I could add that here (with it being unused so far) in order to set up a connection for it in the PipelineTask I'll be writing.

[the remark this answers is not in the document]
TallJimbo: I've added that `mask` parameter, and have now also factored out the background- and masked-image-based updates into separate methods.  Since I'll be loading the `calexp` mask anyway, it probably makes sense to load those and recompute their stats for consistency.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -339,3 +321,61 @@ def update_photo_calib_stats(self, summary, photo_calib):
...
         else:
             summary.zeroPoint = float("nan")
+
+    def update_background_stats(self, summary, background):
+        """Compute summary-statistic fields that depend on only on the
+        background model.
+
+        Parameters
+        ----------
+        summary : `ExposureSummaryStatistics`
```
laurenam, line 340: `lsst.afw.image.ExposureSummaryStats`?  (but at least remove the "istic").

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -339,3 +321,61 @@ def update_photo_calib_stats(self, summary, photo_calib):
...
+        else:
+            summary.skyBg = float("nan")
+
+    def update_masked_image_stats(self, summary, masked_image):
+        """Compute summary-statistic fields that depend on the masked image
+        itself.
+
+        Parameters
+        ----------
+        summary : `ExposureSummaryStatistics`
```
laurenam, line 366: `lsst.afw.image.ExposureSummaryStats`?  (but at least remove the "istic").

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -339,3 +321,61 @@ def update_photo_calib_stats(self, summary, photo_calib):
             summary.zeroPoint = float(2.5*np.log10(photo_calib.getInstFluxAtZeroMagnitude()))
         else:
             summary.zeroPoint = float("nan")
+
+    def update_background_stats(self, summary, background):
+        """Compute summary-statistic fields that depend on only on the
```
laurenam, line 335: There's an extra "on" here.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -316,13 +246,145 @@ def _computePsfStats(self, sources):
...
+        """
+        nan = float("nan")
+        if masked_image is None:
+            summary.skyNoise = nan
+            summary.meanVar = nan
+            return
+        statsCtrl = afwMath.StatisticsControl()
+        statsCtrl.setNumSigmaClip(self.config.sigmaClip)
+        statsCtrl.setNumIter(self.config.clipIter)
+        statsCtrl.setAndMask(afwImage.Mask.getPlaneBitMask(self.config.badMaskPlanes))
```
laurenam, line 380: Did you also want to update the entries in `self.config.badMaskPlanes` here, or is that another ticket?

laurenam (approved):

python/lsst/pipe/tasks/computeExposureSummaryStats.py
TallJimbo: I think it's another ticket.  I don't this one to rock any unnecessary boats.

PR #739, opened 2023-01-02
DM-37411: Add visit-level PSF model robustness metrics

laurenam:


python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -69,6 +69,24 @@ class ComputeExposureSummaryStatsConfig(pexConfig.Config):
         dtype=str,
         default="base_SdssShape_psf"
     )
+    psfSampling = pexConfig.Field(
+        dtype=int,
+        doc="Factor by which to subsample the masked image in each dimension for the maxDistToNearestPsf "
```
TallJimbo, line 74: "Factor by which to subsample" reads to me as the number of points in the grid in each dimension, not the step size.  How about "Sampling rate in pixels in each dimension for the maxDistToNearestPsf grid"?

TallJimbo (approved):

PR #749, opened 2023-01-19
DM-36408: Add Coadd PSF and multiband source fitting tasks

taranu:
These tasks operate on coadd images and are meant to be run sequentially, in contrast to the old monolithic fitting `MultibandFitTask` in `fit_multiband.py`, which did both PSF and source fitting in one task. Also, unlike the old task (which will be deprecated soon), the outputs are astropy tables rather than `SourceCatalog`s. The structure is otherwise similar, with the actual work done by a configurable subtask (not included).


python/lsst/pipe/tasks/fit_coadd_multiband.py
```diff
@@ -0,0 +1,321 @@
...
+class CatalogExposureInputs(CatalogExposure, PsfFitInputs):
+    def get_catalog(self):
+        return self.catalog
+
+
+class CoaddMultibandFitConnections(
+    pipeBase.PipelineTaskConnections,
+    dimensions=("tract", "patch", "skymap"),
+    defaultTemplates=CoaddMultibandFitBaseTemplates,
+):
```
fred3m, line 64: Assuming that you are working with deblender models, you'll also need the scarlet models. See
https://github.com/lsst/meas_base/blob/main/python/lsst/meas/base/forcedPhotCoadd.py#L75-L80 and https://github.com/lsst/meas_base/blob/main/python/lsst/meas/base/forcedPhotCoadd.py#L340-L355 for an example of how to do this as of w.2022.26.

python/lsst/pipe/tasks/fit_coadd_multiband.py
```diff
@@ -0,0 +1,321 @@
...
+        -------
+        retStruct : `lsst.pipe.base.Struct`
+            A struct with a cat_output attribute containing the output
+            measurement catalog.
+
+        Notes
+        -----
+        Subtasks may have further requirements; see `CoaddMultibandFitSubTask.run`.
+        """
+        cat_output = self.fit_coadd_multiband.run(catalog_multi=cat_ref, catexps=catexps).output
```
fred3m, line 319: You'll need code here to add the footprints from the scarlet models, otherwise your individual sources won't have footprints.

python/lsst/pipe/tasks/fit_coadd_psf.py
```diff
@@ -0,0 +1,207 @@
...
+        name="{name_coadd}Coadd_calexp",
+        storageClass="ExposureF",
+        dimensions=("tract", "patch", "band", "skymap"),
+    )
+    cat_meas = cT.Input(
+        doc="Deblended single-band source catalog",
+        name="{name_coadd}Coadd_meas",
+        storageClass="SourceCatalog",
+        dimensions=("tract", "patch", "band", "skymap"),
+    )
```
fred3m, line 69: You'll also need the scarlet models

fred3m (changes requested):
Just a few comments and minor changes requested. Although there seems to be a lot of duplicate code between the different tasks, and I wonder if you could abstract more of it out?

python/lsst/pipe/tasks/fit_coadd_psf.py
taranu: For PSF fitting? I can't see why. If anything, I'm considering dropping the full exposure and only retrieving the PSF models (I think that's possible).

fred3m: If it's just for the PSF models, why do you need the meas catalog then? Are you just using the columns to identify the un-blended point sources?

taranu: No, I need source centroids at the minimum, and I also read the standard `detect_isPrimary`/`merge_peak_sky` flags to skip fitting PSF models for those sources since they won't be processed in the source fitting step anyway.

taranu:
> Just a few comments and minor changes requested. Although there seems to be a lot of duplicate code between the different tasks, and I wonder if you could abstract more of it out?

About the duplication - I accidentally left in a couple of unnecessary methods in the PSF fitting task. Other than the base `CatalogExposure` dataclass that I left in `fit_multiband.py`, I don't think there's anything beneficial to be shared with the multiband fitting task.

`fit_coadd_multiband.py` was indeed adapted from `fit_multiband.py` and some of the code is identical, but I plan to just axe everything but `CatalogExposure` in `fit_multiband.py`. I'm not sure if I need to formally deprecate it considering that the only two implementations of the subtasks are in `lsst-dm` repos, but I will if necessary.

fred3m:
No, I don't think that you need to deprecate lsst-dm code.

taranu:
I meant deprecating the `MultibandFitTask` in `fit_multiband.py`, here in `pipe_tasks`, which is not part of any pipelines in part because there are no implementations of `MultibandFitSub[Config/Task]` in `lsst`, only in `lsst-dm`.

fred3m:
Right, sorry, that's what I meant too. Even though the task is in `pipe_tasks`, it requires code from lsst-dm, so I feel like it would be silly to deprecate (effectively) unused code.

fred3m (approved):

PR #752, opened 2023-02-06
DM-37865: Remove unused parameters from deferred dataset handle get calls

timj:
`datasetType` and `immediate` parameters were for gen2 but will imminently not be allowed in gen3 (they are currently ignored).

### DIFF
--- a/python/lsst/pipe/tasks/multiBand.py
+++ b/python/lsst/pipe/tasks/multiBand.py
@@ -539,7 +539,7 @@ class MeasureMergedCoaddSourcesTask(PipelineTask):
             schema = initInputs['inputSchema'].schema
         if schema is None:
             assert butler is not None, "Neither butler nor schema is defined"
-            schema = butler.get(self.config.coaddName + self.inputCatalog + "_schema", immediate=True).schema
+            schema = butler.get(self.config.coaddName + self.inputCatalog + "_schema").schema
         self.schemaMapper = afwTable.SchemaMapper(schema)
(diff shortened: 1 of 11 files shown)


python/lsst/pipe/tasks/multiBand.py
```diff
@@ -539,7 +539,7 @@ def __init__(self, butler=None, schema=None, peakSchema=None, refObjLoader=None,
             schema = initInputs['inputSchema'].schema
         if schema is None:
             assert butler is not None, "Neither butler nor schema is defined"
-            schema = butler.get(self.config.coaddName + self.inputCatalog + "_schema", immediate=True).schema
+            schema = butler.get(self.config.coaddName + self.inputCatalog + "_schema").schema
```
erykoff, line 542: Wait, are tasks ever initialized with a butler in gen3?

erykoff (approved):
Apologies for not catching all these before in the big removal.  I do note that the "fakes" code is all going away; and the bright star stamp code is completely bit-rotted.  One comment on task initialization with a butler.

python/lsst/pipe/tasks/multiBand.py
timj: I was wondering. The code is clearly never called because if it was called it would break... I wasn't brave enough to remove the butler parameter completely from the 8 places that use it but maybe I should.

timj: I don't think I want to get into that on this ticket though.

PR #757, opened 2023-02-17
DM-38062: Fix dataId property in deferred dataset handle usage

timj:
@erykoff The actual driver for this fix is that the tests fail with debug logging because the mocked deferred dataset handle didn't define the dataId as a property.

Fixing DeferredDatasetHandle usage is now in #758.

### DIFF
--- a/python/lsst/pipe/tasks/drpAssociationPipe.py
+++ b/python/lsst/pipe/tasks/drpAssociationPipe.py
@@ -135,7 +135,7 @@ class DrpAssociationPipeTask(pipeBase.PipelineTask):
 
         Parameters
         ----------
-        diaSourceTables : `list` of `lst.daf.butler.DeferredDatasetHandle`
+        diaSourceTables : `list` of `lsst.daf.butler.DeferredDatasetHandle`
             Set of DiaSource catalogs potentially covering this patch/tract.
         skyMap : `lsst.skymap.BaseSkyMap`
             SkyMap defining the patch/tract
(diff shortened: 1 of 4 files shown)


tests/assembleCoaddTestUtils.py
```diff
@@ -46,38 +46,20 @@
...
 
 
-class MockWarpReference(lsst.daf.butler.DeferredDatasetHandle):
+class MockWarpReference(pipeBase.InMemoryDatasetHandle):
     """Very simple object that looks like a Gen 3 data reference to a warped
     exposure.
 
     Parameters
     ----------
     exposure : `lsst.afw.image.Exposure`
```
ktlim, line 55: I think this should at least be documented as mapping to the `inMemoryDataset` attribute of the dataclass.

tests/surveyPropertyMapsTestUtils.py
```diff
@@ -180,7 +180,7 @@ def makeMockVisitSummary(visit,
     return visit_summary
 
 
-class MockVisitSummaryReference(lsst.daf.butler.DeferredDatasetHandle):
+class MockVisitSummaryReference(lsst.pipe.base.InMemoryDatasetHandle):
```
ktlim, line 183: I'm not sure I understand why these are not using the `inMemoryDataset` attribute as was done in `assembleCoaddTestUtils.py`.

timj: I've fixed on #758 -- there was a bit of subtlety in the tests reusing the ref since InMemoryDatasetHandle is a frozen data class.

ktlim (approved):

PR #758, opened 2023-02-21
DM-38091: Use InMemoryDatasetHandle

timj:
Many of the get() implementations here might Just Work using the default ExposureF delegate although the cloning going on makes me worried that the returned item is modified and so it won't work because we don't clone in the default delegate. We might also need to think about a way for an InMemoryDatasetHandle to specify a default dataId constructor and also whether additional kwargs can be stored and then retrieved later on.

Requires lsst/pipe_base#318

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -44,6 +44,7 @@ import astropy.units as u
 from astropy.coordinates import SkyCoord
 
 from lsst.utils import doImport
+from lsst.utils.introspection import get_full_type_name
 from lsst.daf.butler import DeferredDatasetHandle
 import lsst.geom as geom
 import lsst.sphgeom as sphgeom
@@ -339,6 +340,8 @@ class Functor(object):
         elif isinstance(data, DeferredDatasetHandle):
             # Load in-memory dataframe with appropriate columns the gen3 way
             df = data.get(parameters={"columns": columns})
+        else:
+            raise RuntimeError(f"Unexpected type provided for data. Got {get_full_type_name(data)}.")
 
         # Drop unnecessary column levels
         if is_multiLevel:
@@ -355,8 +358,8 @@ class Functor(object):
         return vals.dropna()
 
     def __call__(self, data, dropna=False):
+        df = self._get_data(data)
         try:
-            df = self._get_data(data)
             vals = self._func(df)
         except Exception as e:
             self.log.error("Exception in %s call: %s: %s", self.name, type(e).__name__, e)
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -293,6 +293,7 @@ class Functor(object):
                 return self._colsFromDict(columnDict, columnIndex=columnIndex)
             else:
                 return columnDict
+        raise RuntimeError(f"Unexpected data type. Got {get_full_type_name}.")
 
     def _func(self, df, dropna=True):
         raise NotImplementedError('Must define calculation on dataframe')
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -46,6 +46,7 @@ from astropy.coordinates import SkyCoord
 from lsst.utils import doImport
 from lsst.utils.introspection import get_full_type_name
 from lsst.daf.butler import DeferredDatasetHandle
+from lsst.pipe.base import InMemoryDatasetHandle
 import lsst.geom as geom
 import lsst.sphgeom as sphgeom
 
@@ -95,7 +96,7 @@ class Functor(object):
     """Define and execute a calculation on a ParquetTable
 
     The `__call__` method accepts either a `ParquetTable` object or a
-    `DeferredDatasetHandle`, and returns the
+    `DeferredDatasetHandle` or `InMemoryDatasetHandle`, and returns the
     result of the calculation as a single column.  Each functor defines what
     columns are needed for the calculation, and only these columns are read
     from the `ParquetTable`.
@@ -185,12 +186,14 @@ class Functor(object):
 
         Parameters
         ----------
-        data : `MultilevelParquetTable` or `DeferredDatasetHandle`
-
+        data : various
+            The data to be read, can be a `MultilevelParquetTable`,
+            `DeferredDatasetHandle`, or `InMemoryDatasetHandle`.
         columnnIndex (optional): pandas `Index` object
-            if not passed, then it is read from the `DeferredDatasetHandle`
+            If not passed, then it is read from the `DeferredDatasetHandle`
+            for `InMemoryDatasetHandle`.
         """
-        if isinstance(data, DeferredDatasetHandle):
+        if isinstance(data, (DeferredDatasetHandle, InMemoryDatasetHandle)):
             if columnIndex is None:
                 columnIndex = data.get(component="columns")
         if columnIndex is not None:
@@ -203,11 +206,13 @@ class Functor(object):
     def _get_data_columnLevelNames(self, data, columnIndex=None):
         """Gets the content of each of the column levels for a multilevel table
 
-        Similar to `_get_data_columnLevels`, this enables backward compatibility with gen2.
+        Similar to `_get_data_columnLevels`, this enables backward
+        compatibility with gen2.
 
-        Mirrors original gen2 implementation within `pipe.tasks.parquetTable.MultilevelParquetTable`
+        Mirrors original gen2 implementation within
+        `pipe.tasks.parquetTable.MultilevelParquetTable`
         """
-        if isinstance(data, DeferredDatasetHandle):
+        if isinstance(data, (DeferredDatasetHandle, InMemoryDatasetHandle)):
             if columnIndex is None:
                 columnIndex = data.get(component="columns")
         if columnIndex is not None:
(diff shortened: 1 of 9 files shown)


tests/assembleCoaddTestUtils.py
```diff
@@ -479,7 +443,11 @@ def makeDataRefList(exposures, matchedExposures, warpType, tract=0, patch=42, co
...
             else:
                 raise ValueError("warpType must be one of 'direct' or 'psfMatched'")
-            dataRef = MockWarpReference(exposure, coaddName=coaddName,
-                                        tract=tract, patch=patch, visit=expId)
+            dataId = lsst.daf.butler.DataCoordinate.standardize(
+                tract=tract, patch=patch, visit=expId,
+                instrument="DummyCam", skymap="Skymap",
+                universe=lsst.daf.butler.DimensionUniverse(),
+            )
+            dataRef = MockWarpReference(exposure, dataId=dataId)
```
erykoff, line 451: Is the preferred pattern this (hand the dataId to the mock initializer) or what is done below (override init and construct the dataId in the mock initializer)?  I think we should pick a pattern and stick with it for clarity now and in the future.

tests/test_makeSurveyPropertyMaps.py
```diff
@@ -220,9 +220,9 @@ def testPropertyMapCreationEmptyInputMap(self):
         # Replace the input map with an empty one.
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
-            self.input_map_dict[0].input_map = hsp.HealSparseMap.make_empty_like(
-                self.input_map_dict[0].input_map
-            )
+            self.input_map_dict[0] = MockInputMapReference(hsp.HealSparseMap.make_empty_like(
+                self.input_map_dict[0].inMemoryDataset
+            ), **self.input_map_dict[0].dataId)
```
erykoff, line 225: So you didn't like the way I was just hacking in the value? 😄

erykoff (commented):
Looks basically good, but I'm a bit worried that we have two different patterns here about how to subclass the `InMemoryDatasetHandle`, to generate a dataId or not.  I'd like to settle on one or the other.

tests/test_makeSurveyPropertyMaps.py
timj: It's not a question of "like" it's a problem with InMemoryDatasetHandle being read-only.

tests/assembleCoaddTestUtils.py
erykoff: If the `InMemoryDataset` had something like `mock_dataid=dict()` that then turned some arbitrary dict into a generic dataid, wouldn't that work generally for these sorts of purposes, and also be clear that this was just for mocking in tests?

timj: The `dataId=` for an `InMemoryDatasetHandle` is meant to be the mock dataId already -- there is no need to add an alternative mock dataId. Allowing it to be a `dict` but turning it into some kind of mocked DataCoordinate internally should be okay (we actually have a mocked DataCoordinate thing in butler testing).

python/lsst/pipe/tasks/functors.py
```diff
@@ -475,10 +484,12 @@ def __call__(self, data, **kwargs):
...
-        data : `lsst.daf.butler.DeferredDatasetHandle`,
-               `lsst.pipe.tasks.parquetTable.MultilevelParquetTable`,
-               `lsst.pipe.tasks.parquetTable.ParquetTable`,
-               or `pandas.DataFrame`.
+        data : various
+            The data represented as `lsst.daf.butler.DeferredDatasetHandle`,
+            `lsst.pipe.tasks.parquetTable.MultilevelParquetTable`,
+            `lsst.pipe.tasks.parquetTable.ParquetTable`,
+            `lsst.pipe.base.InMemoryDatasetHandle`,
+            or `pandas.DataFrame`.
```
erykoff, line 492: Does this actually accept a DataFrame?  Is that tested?

tests/assembleCoaddTestUtils.py
```diff
@@ -91,36 +73,18 @@ def get(self, bbox=None, component=None, parameters=None):
...
         elif component == 'visitInfo':
-            return self.exposure.getInfo().getVisitInfo()
+            return self.inMemoryDataset.getInfo().getVisitInfo()
+        bbox = None
         if parameters is not None:
-            if "bbox" in parameters:
-                bbox = parameters["bbox"]
-        exp = self.exposure.clone()
+            bbox = parameters.get("bbox")
+        exp = self.inMemoryDataset.clone()
```
timj, line 82: If this `clone()` is removed everything breaks because assembleMetadata gets no data sets but doesn't check it has none.

```python
tests/test_assembleCoadd.py:87: in runQuantum
    retStruct = self.run(mockSkyInfo, inputs.tempExpRefList, inputs.imageScalerList,
../../stack/lsst-scipipe-5.1.0/Darwin/utils/g8963c4024b+f0a1800e64/python/lsst/utils/timer.py:347: in timeMethod_wrapper
    res = func(self, *args, **keyArgs)
python/lsst/pipe/tasks/assembleCoadd.py:667: in run
    self.assembleMetadata(coaddExposure, tempExpRefList, weightList)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <test_assembleCoadd.MockAssembleCoaddTask object at 0x16c3de920>, coaddExposure = <lsst.afw.image.exposure.ExposureF object at 0x16af492f0>, tempExpRefList = [], weightList = []

    def assembleMetadata(self, coaddExposure, tempExpRefList, weightList):
        """Set the metadata for the coadd.
    
        This basic implementation sets the filter from the first input.
    
        Parameters
        ----------
        coaddExposure : `lsst.afw.image.Exposure`
            The target exposure for the coadd.
        tempExpRefList : `list`
            List of data references to tempExp.
        weightList : `list`
            List of weights.
    
        Raises
        ------
        AssertionError
            Raised if there is a length mismatch.
        """
        assert len(tempExpRefList) == len(weightList), "Length mismatch"
    
        # We load a single pixel of each coaddTempExp, because we just want to get at the metadata
        # (and we need more than just the PropertySet that contains the header), which is not possible
        # with the current butler (see #2777).
        bbox = geom.Box2I(coaddExposure.getBBox().getMin(), geom.Extent2I(1, 1))
    
        tempExpList = [tempExpRef.get(parameters={'bbox': bbox}) for tempExpRef in tempExpRefList]
    
        numCcds = sum(len(tempExp.getInfo().getCoaddInputs().ccds) for tempExp in tempExpList)
    
        # Set the coadd FilterLabel to the band of the first input exposure:
        # Coadds are calibrated, so the physical label is now meaningless.
>       coaddExposure.setFilter(afwImage.FilterLabel(tempExpList[0].getFilter().bandLabel))
E       IndexError: list index out of range

python/lsst/pipe/tasks/assembleCoadd.py:748: IndexError
```

erykoff, line 82: Is this override necessary (rather than simply using the straight `InMemoryDatasetHandle`) because of this clone?  I'm wondering if we could avoid this subclass and do the clone below in `makeDataRefList`? https://github.com/lsst/pipe_tasks/blob/6a5135d49b1060558619b0b02ca348b6040e185a/tests/assembleCoaddTestUtils.py#L451

tests/assembleCoaddTestUtils.py
erykoff: Is this needed now, or can it accept a dict with recent updates?

tests/assembleCoaddTestUtils.py
timj: Changing that line to:
```python
dataRef = MockWarpReference(exposure.clone(), dataId=dataId)
```
does not fix the problem. It's almost like assembleCoaddTask is modifying the data it retrieves and so subsequent calls are retrieving something different each time.

tests/test_makeSurveyPropertyMaps.py
```diff
@@ -156,7 +154,7 @@ def setUp(self):
         inputs.ccds = ccds
         coadd.getInfo().setCoaddInputs(inputs)
 
-        coadd_ref = MockCoaddReference(coadd, patch=patch, tract=tract)
+        coadd_ref = InMemoryDatasetHandle(coadd, patch=patch, tract=tract, storageClass="ExposureF")
```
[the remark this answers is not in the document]
TallJimbo: Seems to me like all criteria of https://developer.lsst.io/team/empowerment.html are in play.  Change away.

python/lsst/pipe/tasks/functors.py
timj: No idea. I was simply adjusting the text.

tests/assembleCoaddTestUtils.py
timj: There are three places inside `assembleCoaddTask` that are assuming that each time `get()` is called it is getting a brand new exposure that it can modify. This suggests that we are required to keep a special subclass of InMemoryDatasetHandle for now until there is some way that we can indicate that a clone/copy is needed. We could have a new parameter to the InMemoryDatasetHandle constructor that indicates whether a copy/clone is required or not. This would then have to be something that delegate has to understand since I don't think we can assume that `copy.deepcopy` would always work (a parameter to `get()` seems painful since DeferredDatasetHandle for file datastore will always give you a new copy -- it can't do anything else)

erykoff: I was just thinking the same thing about InMemoryDatasetHandle.

timj: Will not solve this on the current ticket.

tests/assembleCoaddTestUtils.py
```diff
@@ -91,36 +73,18 @@ def get(self, bbox=None, component=None, parameters=None):
...
+            return self.inMemoryDataset.getPsf()
         elif component == 'visitInfo':
-            return self.exposure.getInfo().getVisitInfo()
+            return self.inMemoryDataset.getInfo().getVisitInfo()
+        bbox = None
         if parameters is not None:
-            if "bbox" in parameters:
-                bbox = parameters["bbox"]
-        exp = self.exposure.clone()
+            bbox = parameters.get("bbox")
```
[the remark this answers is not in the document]
timj: I removed all this code and now use the standard Exposure delegate in the base class.

tests/assembleCoaddTestUtils.py
timj: Changed it, and now also retain the coaddName since it no longer has to be a real DataCoordinate.

erykoff (approved):

tests/test_makeSurveyPropertyMaps.py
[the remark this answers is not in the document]
timj: See lsst/afw#685

PR #760, opened 2023-02-28
DM-37955: Update tasks and tests with new robust aperture correction.

erykoff:


python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -417,11 +417,12 @@ def run(self, exposure, exposureIdInfo=None, background=None):
         # if wanted
         self.measurement.run(measCat=dmeRes.sourceCat, exposure=dmeRes.exposure,
                              exposureId=exposureIdInfo.expId)
+        # Run the catalog calculations before aperture correction.
+        self.catalogCalculation.run(dmeRes.sourceCat)
```
TallJimbo, line 421: This is a potential problem.  We actually invented `CatalogCalculations` as a separate class of plugin from the "measurement" ones in order to have a round that runs after aperture corrections have been applied, since the archetypal catalog calculation is Extendedness, and that only works if the PSF fluxes and galaxy-model fluxes have been put on the same system.

I'm guessing you're moving this earlier so the star selector can _use_ Extendedness, but it's not going to be any good before aperture corrections are applied - or at least it's not going to be as good then as it ought to be, even if it's being fed the Gaussian fluxes instead of CModel fluxes and that puts a ceiling on how good it can be.

Easiest potential way out of this would be to not use extendedness to feed aperture correction measurement, if we can get away with that.  We might want to add a different S/G classification cut that just uses source to PSF size ratios somewhere else, like the science source selector, as a replacement.

TallJimbo (changes requested):

TallJimbo (approved):

PR #766, opened 2023-04-05
DM-38575: Change filterData to floats and add seed

cmsaunders:


python/lsst/pipe/tasks/maskStreaks.py
```diff
@@ -624,7 +624,7 @@ def _cannyFilter(self, image):
         cannyData : `np.ndarray`
             2-d image of edges found in input image.
         """
-        filterData = image.astype(int)
+        filterData = (image > 0).astype(np.float64)
```
erykoff, line 627: We should probably add a comment here explicitly that we are setting pixels to 0 or 1 and making the map a float to be compatible with the canny filter routine.

stefanv:
> Two things before a full review ... first, I don't see the seed change, was that not committed? Second, the whole discussion at [scikit-image/scikit-image#6871](https://github.com/scikit-image/scikit-image/issues/6871) makes me really, really, really, **really**, _really_ worried that we are relying on this code. Because it does not look reliable at all.

@erykoff I understand your concern (although I may only have used one or two really's :wink:); I would feel the same if I were building a pipeline that needed to produce highly reliable and consistent results over a very long period of time. There are some obvious ways to address that:

1. Pin the skimage version
2. Vendor the code
3. Implement a canny edge detector yourselves

From the skimage perspective, let me address the principle violated:

We strive to *never* have code change introduce changes in *numerical* results, unless it is a bug-fix that gets us closer to the "true" answer. This is the reason behind the skimage v2 redesign, so we have the freedom to make some such changes.

The change you witnessed was introduced inadvertently. We consider this of a breach of contract with our users, and for that I apologize. When I saw your bug report, I spent most of the day debugging it; my primary goal here is to ensure that whatever results you see, they are *correct*.

I cannot speak as to the wisdom of relying on skimage; but I want to clarify the intent of the team working on it. Also thank you for flagging this bug; while we do our best to test our own code, we are very grateful for extensive and careful testing done by the community.

erykoff (approved):

PR #771, opened 2023-04-17
DM-38689: Remove gen2 compatibility code from functors and deprecate pipe.tasks.parquetTable

erykoff:


timj (approved):
Looks great.

Do the test_functors.py tests run even faster now without the temporary files?

PR #772, opened 2023-04-18
DM-31924: update to new system for catalog ID generation

TallJimbo:


python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -336,13 +342,13 @@ def __init__(self, butler=None, refObjLoader=None, schema=None, **kwargs):
 
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
-        if 'exposureIdInfo' not in inputs.keys():
-            inputs['exposureIdInfo'] = ExposureIdInfo.fromDataId(butlerQC.quantum.dataId, "visit_detector")
+        if 'idGenerator' not in inputs.keys():
+            inputs['idGenerator'] = self.config.idGenerator.apply(butlerQC.quantum.dataId)
```
parejkoj, line 346: When would `idGenerator` be in `inputs.keys()`? I don't see how that could happen.

python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
```diff
@@ -78,17 +78,18 @@ class DeblendCoaddSourceSingleConnections(PipelineTaskConnections,
         storageClass="SourceCatalog"
     )
 
-    def setDefaults(self):
-        super().setDefaults()
-        self.singleBandDeblend.propagateAllPeaks = True
```
parejkoj, line 83: Good catch that this is in the wrong place!

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1448,11 +1461,14 @@ def run(self, visitSummaryRefs):
...
-                       summaryTable['id']]
-            packer = visitSummaryRef.dataId.universe.makePacker('visit_detector', visitSummaryRef.dataId)
-            ccdVisitIds = [packer.pack(dataId) for dataId in dataIds]
-            ccdEntry['ccdVisitId'] = ccdVisitIds
+            ccdEntry['ccdVisitId'] = [
+                self.config.idGenerator.apply(
+                    visitSummaryRef.dataId,
+                    detector=detector_id,
+                    is_exposure=False,
+                ).catalog_id
```
parejkoj, line 1469: Assigning a "catalog_id" to a "ccdVisitId" feels a bit odd to me. Should we rename the "ccdVisitId" database key (which is why we have this in the first place), or just live with this slightly confusing assignment?

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -393,51 +394,70 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
-                    Photometric calibration to be used to calibrate the fake sources
-        exposureIdInfo : `lsst.obs.base.ExposureIdInfo`
-        icSourceCat : `lsst.afw.table.SourceCatalog`
-                    Default : None
-                    Catalog to take the information about which sources were used for calibration from.
-        sfdSourceCat : `lsst.afw.table.SourceCatalog`
-                    Default : None
-                    Catalog produced by singleFrameDriver, needed to copy some calibration flags from.
+        wcs : `lsst.afw.geom.SkyWcs`, optional
+            WCS to use to add fake sources
```
parejkoj, line 411: Thanks for the formatting cleanups. Mind adding punctuation to these while you're at it?

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -393,51 +394,70 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+        exposureIdInfo : `lsst.obs.base.ExposureIdInfo`, optional
+            Object that carries ID information for this image/catalog.
+            Deprecated in favor of ``idGenerator``.
+        icSourceCat : `lsst.afw.table.SourceCatalog`, optional
+            Catalog to take the information about which sources were used for
+            calibration from.
+        sfdSourceCat : `lsst.afw.table.SourceCatalog`, optional
+            Catalog produced by singleFrameDriver, needed to copy some
+            calibration flags from.
+        externalSkyWcsGlobalCatalog : `lsst.afw.table.ExposureCatalog`, \
```
parejkoj, line 423: I did not know that was the correct way to extend a long argument docstring.

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -713,7 +753,7 @@ def run(self, fakeCats, exposure, skyMap, wcs=None, photoCalib=None, exposureIdI
                                        ccdVisitFakeMagnitudes=ccdVisitMagnitudes)
         return resultStruct
 
-    def addVariablity(self, fakeCat, band, exposure, photoCalib, exposureIdInfo):
+    def addVariability(self, fakeCat, band, exposure, photoCalib, rngSeed):
```
parejkoj, line 756: The method name change counts as an API change, but I'm definitely not going to ask you to RFC it or add deprecations!

parejkoj (commented):
Several comments about the relationship between `runQuantum`, `inputs` and `run`, which might not be worth digging into too far on this ticket.

Seeing the new packer in use here, it feels mostly like a side-grade in terms of code clarity, but a definite upgrade in terms of flexibility.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -478,7 +480,7 @@ def __init__(self, butler=None, astromRefObjLoader=None,
 
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
-        inputs['exposureIdInfo'] = ExposureIdInfo.fromDataId(butlerQC.quantum.dataId, "visit_detector")
+        inputs['idGenerator'] = self.config.idGenerator.apply(butlerQC.quantum.dataId)
```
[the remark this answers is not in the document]
TallJimbo: I agree with you on both points, especially in terms of recommendations for new code rather than something to try to fix in existing tasks.  But I definitely don't want to attempt that on this ticket.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -615,27 +621,29 @@ def run(self, exposure, exposureIdInfo=None, background=None,
             matchMeta = astromRes.matchMeta
             if exposure.getWcs() is None:
                 if self.config.requireAstrometry:
-                    raise RuntimeError(f"WCS fit failed for {exposureIdInfo.expId} and requireAstrometry "
+                    raise RuntimeError(f"WCS fit failed for {idGenerator} and requireAstrometry "
```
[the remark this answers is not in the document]
TallJimbo: Yeah, I agree that it's sort of weird to combine the roles of "ID generator" and "data ID formatting" abstraction, especially in a class that's 95% about the former only, but I think it's actually quite convenient in a way that passing another argument for the latter would not be.  The real advantage is that in both cases `IdGenerator` abstracts over whether there really is a data ID or not, so the `run` method doesn't have to care.

Maybe what we need is to put the `__str__` implementation in a better-named method?  I liked `__str__` because it meant we could use the deferred-stringification stuff in `logging` naturally, but that's not going to be a meaningful optimization.

python/lsst/pipe/tasks/characterizeImage.py
TallJimbo: Me neither, just replicating the existing logic.

python/lsst/pipe/tasks/postprocess.py
TallJimbo: I'll add a comment; this is a slightly unusual usage of `IdGenerator` and that's why it seems strange.  It's a valid usage but a rare enough one that I don't that to inspire any name changing in `IdGenerator` itself.

python/lsst/pipe/tasks/processCcdWithFakes.py
TallJimbo: Yeah, I figured the odds were in my favor that nothing else was calling this.

This whole ticket really reminded me of the importance of making your methods private by default and public only when needed, which is something I don't think we've directly encouraged enough and definitely weren't doing when most of pipe_base was written.

python/lsst/pipe/tasks/calibrate.py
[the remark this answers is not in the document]
parejkoj: Would you mind filing a ticket about this? Something to encompass both implementing the change to "Our convention is to call `run` from `runQuantum` with `**inputs`, as extracted from the quantum graph, plus separate named kwargs." in existing code, and also documenting it somewhere (the `PipelineTask` docs?).

python/lsst/pipe/tasks/processCcdWithFakes.py
parejkoj: Yeah, we absolutely should do that more.

python/lsst/pipe/tasks/calibrate.py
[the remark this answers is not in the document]
TallJimbo: I'm sure I actually agree it's worth making a convention out of.  I agree I generally prefer what you've said, but it's not like the alternative is in any way problematic (even in terms of readability) and I'm wary of over-prescription in conventions.

python/lsst/pipe/tasks/characterizeImage.py
parejkoj: Digging into the history, this is a holdover from the gen2 `adaptArgsAndRun`, and only `processCcdWithFakes` has similar logic (added at a similar time, likely one cargo culted from the other). That `if` looks entirely unnecessary now. Don't need to change it here, since I plan to make `characterizeImage` go away with the new `CalibrateImage`.

python/lsst/pipe/tasks/calibrate.py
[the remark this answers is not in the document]
parejkoj: I think either we can clearly document the `IdGenerator`s to say that they are used as standins for dataIds in logging in task `run` methods and keep the `str` behavior as-is, or add a `print_data_id` method (or something like that) and then we'd be able to make `str` more relevant to the idgenerator itself.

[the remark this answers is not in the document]
TallJimbo: I've done that documenting in the `meas_base` PR now (it's at the top of the `IdGenerator` class docs.

PR #774, opened 2023-04-19
DM-38751: Allow MeasureApCorrTask to fail and filter when making warps.

erykoff:


python/lsst/pipe/tasks/characterizeImage.py
```diff
@@ -426,9 +431,22 @@ def run(self, exposure, exposureIdInfo=None, background=None):
...
+                    catalog=dmeRes.sourceCat,
+                ).apCorrMap
+                apCorrSuccess = True
+            except MeasureApCorrError:
+                # We have failed to get a valid aperture correction map.
+                # Proceed with processing, and image will be filtered
+                # downstream.
+                dmeRes.exposure.info.setApCorrMap(None)
+
+            if apCorrSuccess:
```
TallJimbo, line 447: You've discovered the exact reason `try/else` exists, and this is possibly the best opportunity you'll ever have to use it!
```
            try:
                apCorrMap = self.measureApCorr.run(
                    exposure=dmeRes.exposure,
                    catalog=dmeRes.sourceCat,
                ).apCorrMap
            except MeasureApCorrError:
                # We have failed to get a valid aperture correction map.
                # Proceed with processing, and image will be filtered
                # downstream.
                dmeRes.exposure.info.setApCorrMap(None)
            else:
                dmeRes.exposure.info.setApCorrMap(apCorrMap)
                self.applyApCorr.run(catalog=dmeRes.sourceCat, apCorrMap=exposure.getInfo().getApCorrMap())
```
No `apCorrSuccess` variable necessary!

But feel free to ignore this suggestion if you find the original more readable; just know that by doing so you'd practically be declaring that `try/else` shouldn't exist in Python.

TallJimbo (approved):

PR #776, opened 2023-04-21
DM-38667: Putting a condition on the minimum number of valid pixels within the normalisation annulus

bazkiaei:


bazkiaei:
The corrections in this PR are merged to the main branch through PR #786

PR #780, opened 2023-05-01
DM-38955: Fix composite functor error log and fix bad EBV units.

erykoff:
This also adds associated tests.

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -1910,7 +1910,7 @@ class Ebv(Functor):
         super().__init__(**kwargs)
 
     def _func(self, df):
-        coords = SkyCoord(df['coord_ra']*u.rad, df['coord_dec']*u.rad)
+        coords = SkyCoord(df['coord_ra'].values * u.rad, df['coord_dec'].values * u.rad)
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/functors.py
```diff
@@ -479,7 +480,13 @@ def __call__(self, data, **kwargs):
                     )
                     valDict[k] = f._func(subdf)
                 except Exception as e:
-                    self.log.error("Exception in %s call: %s: %s", self.name, type(e).__name__, e)
+                    self.log.error(
```
timj, line 483: Are you sure you don't want to use `self.log.exception` here and automatically report the stack trace? If someone sees an exception like this inside the functor code will they be able to work out where their code triggered it?

erykoff: I am not sure about that, I just changed this to work better than it was working.  How does one use `self.log.exception`?

timj: Same as `self.log.error` except it automatically includes the exception information.

timj (approved):
I can see the name is now defined and the logging is better.

I have no idea about the functors themselves and the tests but they look fine.

PR #782, opened 2023-05-03
DM-38694: Use InMemoryDatasetHandle in assemble coadd test

timj:
Depends on lsst/pipe_base#328

### DIFF
--- a/tests/test_skySources.py
+++ b/tests/test_skySources.py
@@ -24,6 +24,7 @@ import unittest
 import logging
 
 import lsst.afw.image as afwImage
+import lsst.meas.extensions.piff.piffPsfDeterminer
 import lsst.utils.tests
 from lsst.utils import getPackageDir
 from lsst.pipe.tasks.characterizeImage import CharacterizeImageTask, CharacterizeImageConfig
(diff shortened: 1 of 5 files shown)


tests/test_skySources.py
```diff
@@ -24,6 +24,7 @@
 import logging
 
 import lsst.afw.image as afwImage
+import lsst.meas.extensions.piff.piffPsfDeterminer
```
erykoff, line 27: Why is this now needed?

erykoff (approved):
I think you're correct that this is not used anywhere.  But searching for other misspelled uses of "seperation" is amusing.  Both rubin sim and drp_pipe have had somebody misspell the word (but only in docs, not methods or this functor usage).

tests/test_skySources.py
timj: Because I tried to run the test file on its own (it's the slowest test now) and it didn't work. The test code was relying on other test code doing that import.

timj: Is that telling me there is a file in pipe_tasks itself that is not doing this import? I'm not really sure how the meas_extensions import stuff works.

PR #784, opened 2023-05-09
DM-39031: Use resolved references for HiPS graph

andy-slac:
This commit adds two command line options to the graph builder script which define output collection names, similarly to `pipetask qgraph` command. Output run is used to resolve references when creating quantum graph for HiPS task. At least one of the `--output` or `--output-runs` options must be provided when invoking the script.


python/lsst/pipe/tasks/hips.py
```diff
@@ -429,6 +447,24 @@ def _make_cli_parser(cls):
                 help="Data ID expression used when querying for input coadd datasets.",
             )
 
+        parser_build.add_argument(
+            "--output",
+            type=str,
+            help="Name of the output CHAINED collection.",
```
timj, line 453: Mention that if this is given without output_run that a default timestamped output run will be used?

python/lsst/pipe/tasks/hips.py
```diff
@@ -454,8 +490,10 @@ def build_quantum_graph(
         registry,
         constraint_order,
         constraint_ranges,
+        output_run,
         where=None,
         collections=None,
+        metadata=None,
```
timj, line 496: Same comment as I made in lsst/drp_analysis#62 about the possibility for output_run parameter to disagree with value in metadata.

python/lsst/pipe/tasks/hips.py
```diff
@@ -477,6 +515,10 @@ def build_quantum_graph(
         constraint_ranges : `lsst.sphgeom.RangeSet`
             RangeSet which describes constraint pixels (HEALPix NEST, with order
             constraint_order) to constrain generated quanta.
+        output : `str`, optional
```
timj, line 518: This isn't a parameter to this method.

python/lsst/pipe/tasks/hips.py
```diff
@@ -489,6 +531,8 @@ def build_quantum_graph(
             Collection or collections to search for input datasets, in order.
             If not provided, ``registry.defaults.collections`` will be
             searched.
+        butler_config : `str`, optional
```
timj, line 534: This is not a parameter. Should be `metadata`?

timj (approved):

python/lsst/pipe/tasks/hips.py
andy-slac: Right, I tried first to instantiate metadata in this method, but it required a bunch of extra parameters, so I decided just pass metadata here as an argument (but forgot to clean the docstring). Now I'm also dropping `output_run` and using one from metadata.

PR #785, opened 2023-05-09
DM-37196: Update fields to reflect new APDB schema

parejkoj:


arunkannawadi:
The fixup commits here were never fixed up and has landed in `main` !!!

PR #789, opened 2023-05-22
DM-39141: Ensure source selectors use detect_isPrimary in single exposure processing.

erykoff:


tests/test_photoCal.py
```diff
@@ -95,6 +95,7 @@ def setUp(self):
         self.config.match.referenceSelection.flags.good = ['photometric']
         self.config.match.referenceSelection.flags.bad = ['resolved']
         self.config.match.sourceSelection.doUnresolved = False  # Don't have star/galaxy in the srcCat
+        self.config.match.sourceSelection.doRequirePrimary = False  # Don't have detect_isPrimary in srcCat
```
laurenam, line 98: No `sky_source`s either?!

laurenam (approved):

tests/test_photoCal.py
erykoff: This is an ancient source catalog that came from who knows where and we just keep carrying forward.

PR #804, opened 2023-06-29
DM-34496: Remove already-deprecated fake-source insertion interfaces

jtmccann:


leeskelvin (approved):

leeskelvin:
LGTM!

leeskelvin (commented):
PS - I was concerned that AP may still be using some of these tools, but on closer inspection, I think we're safe. So, in summary, this still LGTM!

PR #806, opened 2023-06-30
DM-39840: Deprecate doPsfMatch field in coaddBase

arunkannawadi:


python/lsst/pipe/tasks/coaddBase.py
```diff
@@ -57,7 +57,8 @@ class CoaddBaseConfig(pexConfig.Config):
     )
     doPsfMatch = pexConfig.Field(
         dtype=bool,
-        doc="Match to modelPsf? Deprecated. Sets makePsfMatched=True, makeDirect=False",
+        doc="Match to modelPsf? Sets makePsfMatched=True, makeDirect=False",
+        deprecated="This field is no longer used. Will be removed after v26.",  # TODO: DM-39841
```
TallJimbo, line 61:
```suggestion
        deprecated="This field is no longer used. Will be removed after v27.",  # TODO: DM-39841
```

TallJimbo (approved):
Version for removal needs to be changed, otherwise looks good.

python/lsst/pipe/tasks/coaddBase.py
arunkannawadi: Ah, I always get this one wrong. Fixed.

PR #811, opened 2023-07-05
DM-36182: Cleanup and document functors.py.

leeskelvin:


python/lsst/pipe/tasks/functors.py
```diff
@@ -778,7 +785,7 @@ def __init__(self, col, calib=None, **kwargs):
         if calib is not None:
             self.fluxMag0 = calib.getFluxMag0()[0]
```
arunkannawadi, line 786: If we want to remove `calib`, we should first document that this is an unused argument in the docstring, remove this line and instead raise a deprecation warning here. I'd follow a pattern similar to these lines: https://github.com/lsst/pex_config/blob/aa9e307f6b0b52071272d5b1404e053a96c8fdb0/python/lsst/pex/config/config.py#L1543-L1546

leeskelvin: How does this revision (just pushed) look to you? Future deprecation ticket: [DM-39914](https://jira.lsstcorp.org/browse/DM-39914).

python/lsst/pipe/tasks/functors.py
```diff
@@ -735,50 +746,56 @@ def computePixel(row):
...
     _defaultDataset = 'meas'
 
     def __init__(self, col, calib=None, **kwargs):
         self.col = fluxName(col)
         self.calib = calib
         if calib is not None:
+            # TO DO: DM-39914 Remove deprecated calib argument in Mag functor.
             self.fluxMag0 = calib.getFluxMag0()[0]
+            warnings.warn(
+                "Config field 'calib' is deprecated, and will be removed after v27.",
```
arunkannawadi, line 793: I don't think `Config field` is applicable here. It should perhaps be "`calib` argument".

python/lsst/pipe/tasks/functors.py
arunkannawadi: This looks fine, as long as you're sure `getFluxMag0` exists. My GitHub search returns no implementation of it. If that's the case, I'd remove it right away.

arunkannawadi: And either way, the warning should come first before it executes the code in the `if` block.

leeskelvin: As this hasn't been deprecated officially yet, people may still want to use this to access older data. I realize that this may not be something we need to support here, but regardless, my feeling was that it costs us nothing to keep it in for now just in case.

python/lsst/pipe/tasks/functors.py
leeskelvin: Changed  to `The 'calib' argument is deprecated, and will be removed after v27.`.

python/lsst/pipe/tasks/functors.py
```diff
@@ -92,62 +92,64 @@ def init_fromDict(initDict, basePath='lsst.pipe.tasks.functors',
...
+    If not specified, the dataset will fall back on the `_defaultDataset`
+    attribute.
+    If band is not specified and ``dataset`` is anything other than ``'ref'``,
+    then an error will be raised when trying to perform the calculation.
+
+    Originally, `Functor` was set up to expect datasets formatted like the
+    ``deepCoadd_obj`` dataset; that is, a DataFrame with a multi-level column
+    index, with the levels of the column index being ``band``, ``dataset``, and
+    ``column``.
+    It has since been generalized to apply to DataFrames without mutli-level
```
taranu, line 132: multi-level

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -798,64 +800,51 @@ class TransformCatalogBaseConfig(pipeBase.PipelineTaskConfig,
...
-                    - base_PsfFlux
-                filt: HSC-G
-            gauss_magDiff:
-                functor: MagDiff
-                args:
-                    - base_GaussianFlux
-                    - base_PsfFlux
-                filt: HSC-G
-            count:
+            sourceId:
```
taranu, line 817: What was the motivation for this change? The previous example was out of date, or non-functional?

taranu (approved):
Looks fine. I have nitpicky suggestions/questions but nothing particularly important.

python/lsst/pipe/tasks/functors.py
```diff
@@ -954,16 +979,22 @@ def _func(self, df):
...
             raise RuntimeError('No psf shape parameter found in catalog')
 
         return hsm.where(np.isfinite(hsm), sdss) - psf
 
 
 class SdssTraceSize(Functor):
-    """Functor to calculate SDSS trace radius size for sources"""
+    """Functor to calculate SDSS trace radius size for sources.
+
+    SDSS trace radius size is a measure of size equal to the square root of
```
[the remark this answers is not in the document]
leeskelvin: I think renaming of functors is beyond the scope of this ticket. If you feel strongly about this though, please do set up a follow-up ticket to make this singular change.

python/lsst/pipe/tasks/functors.py
```diff
@@ -1067,6 +1143,15 @@ def _func(self, df):
 
 
 class RadiusFromQuadrupole(Functor):
```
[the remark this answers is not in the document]
leeskelvin: As above, if you feel strongly, please do put this on a follow-up ticket.

python/lsst/pipe/tasks/postprocess.py
leeskelvin: The initial motivation was to replace the mis-named flags section and replace the MagDiff examples, as MagDiff was going away. On review, MagDiff remains, but the example edits here were preserved to showcase more typical examples.

PR #813, opened 2023-07-06
DM-35901: Use the new lsst.utils.tests.ImportTestCase class for import test

timj:
Requires lsst/utils#165

### DIFF
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -19,42 +19,23 @@
 # the GNU General Public License along with this program.  If not,
 # see <https://www.lsstcorp.org/LegalNotices/>.
 
-import importlib.resources
-import os.path
 import unittest
 
-from lsst.utils import doImport
+from lsst.utils.tests import ImportTestCase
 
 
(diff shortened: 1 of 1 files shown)


tests/test_import.py
```diff
@@ -19,42 +19,23 @@
...
-            with self.subTest(module=module_name):
-                try:
-                    doImport(module_name)
-                except ImportError as e:
-                    raise AssertionError(f"Error importing module {module_name}: {e}") from e
+    PACKAGES = {
+        "lsst.pipe.tasks",
+        "lsst.pipe.tasks.script",
+        "lsst.pipe.tasks.dataFrameActions",
+    }
```
erykoff, line 38: Any reason this is a set instead of a tuple like elsewhere?

timj: I specifically defined `PACKAGES` to be an `Iterable` in the interface. set() has the advantage that duplicates will not be included. I can change it to a tuple if you prefer it.

erykoff (approved):

PR #823, opened 2023-08-02
DM-40285: Remove unused butler parameter from task constructors

timj:


erykoff (approved):
Looks fine, but reminds me that the bright stars code doesn't actually run anymore, and when can we delete processCcdWithFakes.py?

PR #827, opened 2023-08-28
DM-40525: Fix numpy invalid warnings with numpy 1.24

erykoff:


python/lsst/pipe/tasks/hips.py
```diff
@@ -1152,7 +1152,9 @@ def _write_hips_image(self, hips_base_path, order, pixel, image, png_mapping, sh
 
         # And make a grayscale png as well
 
-        vals = 255 - png_mapping.map_intensity_to_uint8(image.array).astype(np.uint8)
+        with np.errstate(invalid="ignore"):
+            vals = 255 - png_mapping.map_intensity_to_uint8(image.array).astype(np.uint8)
```
parejkoj, line 1156: Looking into the astropy code (which I contributed from our afw version), it has exactly this context manager, except it doesn't do the `astype(uint8)` cast (intentionally, per the docstring). It's unfortunate that we need to wrap it twice like this, but I guess it's fine (since the astropy method can't do the cast, due to how it's used internally).

parejkoj (approved):
Unfortunate, but it's fine.

python/lsst/pipe/tasks/hips.py
erykoff: Well, the unfortunate part is that the astropy wrap is unnecessary; or at least I couldn't figure out how to get a warning to trigger with just the clip.  But maybe I wasn't giving it the "correct" inputs?

parejkoj: Hmmm... those lines go back to RHL's initial implementation circa 2016, so who knows:

https://github.com/lsst/afw/blame/d3e8bac1ddff3274abec121fc06602073268ac50/python/lsst/afw/display/rgb.py

PR #828, opened 2023-09-13
DM-40451: Update multiband measurement task for deblending with partial coverage

fred3m:
Now that meas_extensions_scarlet will deblend footprints that only have partial coverage (eg. in one or more band there is missing data that prevents building a PSF), new columns were added to the catalog to mark sources that were not deblended in a given band.


cmsaunders:
Do you want the new columns to go into the `objectTable_tract`? If so, I think you need to at least add them to `schemas/Object.yaml`.

cmsaunders (commented):
Do you want the new columns to go into the objectTable_tract? If so, I think you need to at least add them to schemas/Object.yaml.

PR #829, opened 2023-09-13
DM-30895: Move AssembleCoaddTask and its children to drp_tasks

arunkannawadi:


fred3m (changes requested):
Don't forget to remove `assembleCoaddTestUtils.py`.

Other than that, and also moving `assembleChi2Coadd.py` this LGTM.

arunkannawadi:
Sorry, did you mean I should move `assembleChi2Coadd.py` as well in this ticket?

fred3m:
Yes. There aren't any branches in development, so it makes sense to move it now as opposed to making a new ticket just to move one file.

erykoff:
Drive-by review here: There need to be stubs here that forward to the new location and raise a deprecation warning if you use the old location.  And a ticket to remove those deprecated usages after whichever release we decided was the correct number.

arunkannawadi:
I thought about that, but it doesn't work in this case, does it? Given that `drp_tasks` is downstream and is not setup if you setup `pipe_tasks`, importing `drp_tasks` here would add a circular dependency. The redirection with a warning works only if you move the code upstream.

That reminds me that I wanted to add this as a note to the dev guide.

arunkannawadi:
- [ ] Move assembleCoaddUtils.py
- [ ] Move assembleChi2Coadd.py
- [ ] Add stubs for import redirects

fred3m (approved):

PR #831, opened 2023-09-13
DM-40746: Fix size and shape of HiPS Allsky file.

erykoff:
This also adds the ability to specify the Allsky file to be a jpeg or png.

### DIFF
--- a/python/lsst/pipe/tasks/hips.py
+++ b/python/lsst/pipe/tasks/hips.py
@@ -1535,7 +1535,20 @@ class GenerateHipsTask(pipeBase.PipelineTask):
             HEALPix order of the minimum order to make allsky file.
         """
         tile_size = self.config.allsky_tilesize
-        n_tiles_per_side = int(np.sqrt(hpg.nside_to_npixel(hpg.order_to_nside(allsky_order))))
+
+        # The Allsky file format is described in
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/hips.py
```diff
@@ -1549,19 +1575,19 @@ def _write_allsky_file(self, hips_base_path, allsky_order):
             matches = re.match(pixel_regex, png_uri.basename())
             pix_num = int(matches.group(1))
             tile_image = Image.open(io.BytesIO(png_uri.read()))
-            row = math.floor(pix_num//n_tiles_per_side)
-            column = pix_num % n_tiles_per_side
+            row = math.floor(pix_num//n_tiles_wide)
+            column = pix_num % n_tiles_wide
```
timj, line 1579: Row and column both use the width?

erykoff: Yeah, it's div and mod to get x/y from an ordered list.

timj (approved):
Looks ok to me.

erykoff:
I have removed the commit which allows you to specify the Allsky filetype since it cannot be specified independently of the rest of the hips images.

erykoff:
Yes indeed, and that's what DM-40747 can do, but it has to be done for everything at once.

PR #835, opened 2023-09-22
DM-39842: Configurable exception handling in CalibrateImageTask

parejkoj:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -413,12 +414,32 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+        assert not inputs, "runQuantum got more inputs than expected"
+
+        result = pipeBase.Struct(exposure=None,
+                                 stars=None,
+                                 background=None,
+                                 psf_stars=None,
+                                 applied_photo_calib=None,
+                                 astrometry_matches=None,
+                                 photometry_matches=None,
+                                 )
```
TallJimbo, line 427: I'm torn as to whether it's better to spell out the possible fields here with `None` values, vs. letting `run` be the sole source of truth on those.  I think the `QuantumContext` changes in `pipe_base` should allow either to work.

parejkoj: To simplify the call to `annotate` below (so it's not a mess of `getattr`), I need to at least specify the items that are passed to `annotate`. I don't like specifying a `result` list in both places (runQuantum and the top of run); I've changed it so `run` just makes an empty Struct (this caught a bug!) which is a bit better.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -424,12 +426,33 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             config=self.config.photometry_ref_loader, log=self.log)
         self.photometry.match.setRefObjLoader(photometry_loader)
 
-        outputs = self.run(id_generator=id_generator, **inputs)
-
-        butlerQC.put(outputs, outputRefs)
+        # This should not happen with a properly configured execution context.
+        assert not inputs, "runQuantum got more inputs than expected"
```
mrawls, line 430: If/when this falls over, will the error message specify that the `runQuantum` being talked about is the one from `calibrateImage`?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -424,12 +426,33 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+                self,
+                result.exposure,
+                result.psf_stars_footprints,
+                result.stars_footprints,
+                log=self.log
+            )
+            butlerQC.put(result, outputRefs)
+            raise error from e
+
+        butlerQC.put(result, outputRefs)
```
mrawls, line 452: Should this be in a `finally` block?

python/lsst/pipe/tasks/calibrateImage.py
mrawls: This feels like it could be fragile if we ever change this task to output more things in the result Struct. Is it not possible to construct a list of arguments to feed the error annotation by iterating over all the existing things in the Struct?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -477,40 +500,43 @@ def run(self, *, exposures, id_generator=None):
...
+            photometry_meta, result.applied_photo_calib = self._fit_photometry(result.exposure,
+                                                                               result.stars_footprints)
+        # fit_photometry returns a new catalog, so we need a new astropy table view.
+        result.stars = result.stars_footprints.asAstropy()
+        if self.config.optional_outputs:
+            result.photometry_matches = lsst.meas.astrom.denormalizeMatches(photometry_matches,
+                                                                            photometry_meta)
 
-        self._summarize(exposure, stars, background)
+        self._summarize(result.exposure, result.stars_footprints, result.background)
```
mrawls, line 537: Why is this using `result.stars_footprints` and not `result.stars` ?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -448,7 +471,7 @@ def run(self, *, exposures, id_generator=None):
         result : `lsst.pipe.base.Struct`
             Results as a struct with attributes:
```
mrawls, line 472: Is the results Struct guaranteed to have all of the documented attributes, now?

tests/test_calibrateImage.py
```diff
@@ -530,14 +540,88 @@ def test_runQuantum_no_optional_outputs(self):
...
+
+        # A generic exception should raise directly.
+        msg = "mocked run exception"
+        with (
+            mock.patch.object(task, "run", side_effect=ValueError(msg)),
+            self.assertRaisesRegex(ValueError, "mocked run exception")
+        ):
+            lsst.pipe.base.testUtils.runTestQuantum(task, self.butler, quantum, mockRun=False)
+
+        # A AlgormithError should write annotated partial outputs.
```
mrawls, line 584: Typo: "An AlgorithmError"

mrawls (approved):

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: Not without using a whole bunch of `getattr` and tests of whether those things have a `metadata` property. In this design, the `PipelineTask` author has to specifically choose which output datasets could get annotated on failure and pass them to `annotate` individually. It might be nice to just pass the `result` struct and let `annotate` figure it out, but that's probably more complicated than it's worth (not all datasets support `metadata`, e.g. `AstropyArrow`).

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: Yes, the assert will give where it happened. This is a "this should absolutely never happen" error, hence the bare assert.

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: No. We only `butler.put` if the task succeeded, or if `AlgorithmError` is caught, all other errors can't be annotated, so writing partial outputs is unhelpful.

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: `ComputeExposureSummaryStatsTask` takes a SourceCatalog. `result.stars` is just for the output. Eventually we'll start converting downstream tasks (like ipdiffim, maybe?) to take `AstropyArrow` input catalogs, but that's a long ways off.

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: I'm not sure what you mean?

If the task completes, it returns `result`. If an exception is raised, the input parameter has (hopefully!) been modified, but nothing is returned.

parejkoj: Oh, but apparently I never added docs for the `result` parameter: I'll do that now.

parejkoj: Here's what I added:

```
        result : `lsst.pipe.base.Struct`, optional
            Result struct that is modified to allow saving of partial outputs
            for some failure conditions. If the task completes successfully,
            this is also returned.
```

python/lsst/pipe/tasks/calibrateImage.py
mrawls: Ah, so `stars_footprints` and `stars` has the same information but in different formats? Cute, OK, carry on.

parejkoj: See the Connections here: https://github.com/lsst/pipe_tasks/blob/main/python/lsst/pipe/tasks/calibrateImage.py#L87

Eventually, the goal will be fore the `stars_footprints` to only contain the footprints and none of the catalog information, but that's a *long* ways off.

PR #836, opened 2023-09-27
DM-40917: Add description and units to healsparse property map metadata

erykoff:


tests/test_makeSurveyPropertyMaps.py
```diff
@@ -50,6 +51,8 @@
 class DistTimesPsfAreaPropertyMap(BasePropertyMap):
     """Property map to compute the distance from the boresight center
     by the psf area. Do not try this at home."""
+    description = "Distance times psf area"
+    units = "deg*pixel**2"
```
enourbakhsh, line 55: This overwrites the `units` module imported from `astropy`. I really think you intended to use `unit` here.

enourbakhsh: Also, would you be against presenting it in LaTeX format, `deg~pixel^2`? Otherwise, users would need to use tools like sympy to automatically get the LaTeX expression for plotting purposes.

enourbakhsh (changes requested):

enourbakhsh (approved):

PR #845, opened 2023-10-18
DM-32769: Implement streak detection and masking in single frame processing

mrawls:
This PR adds maskStreaksTask (with no bells or whistles) as a default step to single frame processing.

It supports both main "modes" of single frame: the original `isr -> characterizeImage -> calibrate` AS WELL AS the newfangled `isr -> calibrateImage`.

The result is that linear DETECTED features are also masked as STREAK.


bsmartradio (approved):
Everything looks good

PR #848, opened 2023-10-20
DM-40387: CalibrateImage improvements

parejkoj:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -463,10 +465,23 @@ def _compute_psf(self, exposure, guess_psf=True):
         cell_set : `lsst.afw.math.SpatialCellSet`
             PSF candidates returned by the psf determiner.
         """
-        self.log.info("First pass detection with Guassian PSF FWHM=%s", self.config.install_simple_psf.fwhm)
+        def log_psf(msg):
```
isullivan, line 468: Just a preference, but I would like to see `exposure` (or `psf`) passed in explicitly, along with `background`.

isullivan (approved):

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: It's a closure, so it already has access to all local variables.

PR #855, opened 2023-10-31
DM-40955: LATISS psf stamp size is incorrect for the large PSFs

laurenam:


erykoff (approved):
I'm surprised this is necessary; a question is whether this was always wrong and we should be doing the import in the config override file?

laurenam:
Excellent point!  I just asked at standup and all agreed it should go in the config.  I’ll do that and check all is working before merging (and then this PR will just get closed).

PR #856, opened 2023-11-01
DM-41512: Add check for no good sources in finalizeCharacterization

laurenam:


python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -598,6 +609,10 @@ def compute_psf_and_ap_corr_map(self, visit, detector, exposure, src, isolated_s
         """
         # Apply source selector (s/n, flags, etc.)
         good_src = self.source_selector.selectSources(src)
+        if sum(good_src.selected) < 1:
```
erykoff, line 612: I would do a check `== 0` here as well.

erykoff (approved):

PR #858, opened 2023-11-03
DM-34769: Remove characterizeImage refObjLoader

parejkoj:
The config fields were removed because they had deprecations in place for over a year, and were deprecated because they never did anything in the first place.


bsmartradio (approved):
Looks good to me, same comment about description as the other ones. I also tested building this locally and everything passed.

PR #861, opened 2023-11-08
DM-41646: Fix logging when all visits rejected in BestSeeingSelectVisitsTask.

erykoff:


python/lsst/pipe/tasks/selectImages.py
```diff
@@ -524,8 +524,16 @@ def run(self, visitSummaries, skyMap, dataId):
...
+
+        if len(output) == 0:
+            self.log.info("All images rejected in BestSeeingSelectVisitsTask.")
+        else:
+            self.log.info(
+                "%d images selected with FWHM range of %f--%f arcseconds",
+                len(output),
+                fwhmSizes[visits.index(output[0])],
+                fwhmSizes[visits.index(output[-1])],
+            )
```
laurenam, line 536: Ok, so this works because `goodVisits` ends up as an empty `dict`, so `butlerQC.put()` is effectively a no-op.  When I encountered a similar situation on DM-41512, I asked about this behavior and whether it's better to explicitly add a `NoWorkFound` (in my case I was sending `None` to `butlerQC.put()`...the consensus was yes, hence: https://github.com/lsst/pipe_tasks/blob/main/python/lsst/pipe/tasks/finalizeCharacterization.py#L375-L376

python/lsst/pipe/tasks/selectImages.py
```diff
@@ -524,8 +524,17 @@ def run(self, visitSummaries, skyMap, dataId):
             output = sortedVisits
         else:
             output = sortedVisits[:self.config.nVisitsMax]
-        self.log.info("%d images selected with FWHM range of %f--%f arcseconds",
-                      len(output), fwhmSizes[visits.index(output[0])], fwhmSizes[visits.index(output[-1])])
+
+        if len(output) == 0:
+            self.log.info("All images rejected in BestSeeingSelectVisitsTask.")
+            raise pipeBase.NoWorkFound(f'No good images found.')
```
laurenam, line 530: Don't think you need the `f` if nothing is getting evaluated (and I wouldn't mind seeing the dataId added 😉).

laurenam (approved):

PR #869, opened 2023-12-12
DM-42126: Remove ignored detectors exposures from being processed in SubtractBrightStarsTask

bazkiaei:


python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -364,6 +372,32 @@ def run(
...
 
+    def detInRegions(self, inputs):
+        """Determine whether the input exposure's detector is in the region(s)
+        where the extended PSF model(s) is(are) available.
+
+        Parameters
+        ----------
+        inputs : `dict`
+            Dictionary containing the inputs to the task, including
+            `inputExposure` and `inputExtendedPsf`.
```
leeskelvin, line 383: Rather than a single global `inputs` dict here, I think it would make more sense to split this into two required inputs: `inputExposure` and `inputExtendedPsf`. This will make it clear exactly what inputs from the input dict are required, what their purposes are, and prevent accidental usage of any other input.

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -364,6 +372,32 @@ def run(
 
         return subtractorExp, invImages, badStamps
 
+    def detInRegions(self, inputs):
```
leeskelvin, line 375: Is this intended to be used globally, outside this file? If not, we might consider renaming this to `_detInRegions`.

leeskelvin (approved):

python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -364,6 +372,32 @@ def run(
...
+        `bool`
+            True if the detector is in the region(s) where the extended PSF
+            model(s) is(are) available, False otherwise.
+        """
+        availableDets = [
+            det
+            for detList in inputs["inputExtendedPsf"].detectors_focal_plane_regions.values()
+            for det in detList.detectors
+        ]
+        if inputs["inputExposure"].detector.getId() in availableDets:
```
leeskelvin, line 396: I'm a fan of full words where possible, and not using storage types in names. I think that helps readability in the future for other developers contributing to this package and maintaining it.

So here for example, `availableDetectors` for `availableDets`, `detector` for `det` and `detectors` for `detList`. However, this is not a major comment, and I'll defer to you if you feel the way you have it now is more legible in general.

PR #871, opened 2023-12-12
DM-41648: Propagate higher order moments columns to objectTable

arunkannawadi:


schemas/Object.yaml
```diff
@@ -309,10 +309,6 @@ funcs:
         functor: Column
         args: slot_ShapeRound_xy
         dataset: meas
-    iPSF_flag:
```
jmeyers314, line 312: Why are you removing this?

jmeyers314 (changes requested):
Generally looks good.  I'd like to understand why you're removing `iPSF_flag` before hitting approve though.

schemas/Object.yaml
arunkannawadi: This is defined earlier here: https://github.com/lsst/pipe_tasks/blob/fb4685b71f45d7d9c43d08f860a877ec8270831b/schemas/Object.yaml#L292-L295

I'm removing the duplicate entry in this commit.

jmeyers314 (approved):

PR #872, opened 2023-12-18
DM-41900: Catch Cholesky failures

cmsaunders:


mrawls (approved):
This is a fine way to catch this specific error; please consider adding a log message to trace it later if this becomes more widespread. Perhaps the first of many informative logs from this task ;)

python/lsst/pipe/tasks/maskStreaks.py
```diff
@@ -378,6 +378,7 @@ def fit(self, dChi2Tol=0.1, maxIter=100):
             Maximum number of fit iterations allowed. The fit should converge in
             ~10 iterations, depending on the value of dChi2Tol, but this
             maximum provides a backup.
+        log : TODO
```
mrawls, line 381: Please complete the TODO 😄

python/lsst/pipe/tasks/maskStreaks.py
```diff
@@ -407,11 +408,18 @@ def line_search(c, dx):
             if chi2 == 0:
                 break
             if not np.isfinite(A).all():
-                # TODO: DM-30797 Add warning here.
                 fitFailure = True
+                if log is not None:
```
mrawls, line 412: Ah so THIS is the canonical way to log inside a function like this? Fascinating. (In what weird situation would log be None??)

mrawls (approved):
Even better with logging, thank you! Please just finish the docstring and good to go.

python/lsst/pipe/tasks/maskStreaks.py
cmsaunders: I think as I typed TODO I subconsciously knew I would forget it....

python/lsst/pipe/tasks/maskStreaks.py
cmsaunders: I'm not sure, but it looks like that there are a bunch of other places in the stack where the logger is passed around like this. I made it optional because it's possible we would want to use the line profile fitter outside the context of task, in which case you wouldn't have a logger set up.

PR #878, opened 2024-01-12
DM-42461: Implement CloughTocher2D Interpolation

arunkannawadi:


python/lsst/pipe/tasks/interpImage.py
```diff
@@ -262,3 +272,172 @@ def interpolateImage(self, maskedImage, psf, defectList, fallbackValue):
...
+            # Restrict to the bounding box of the image.
+            window.clip(bbox)
+
+            for xx, yy in product(range(window.minX, window.maxX + 1), range(window.minY, window.maxY + 1)):
+                if not (image.mask[xx, yy] & afwImage.Mask.getPlaneBitMask(maskPlanes)):
+                    goodpix[(xx, yy)] = image.image[xx, yy]
+        elif (x, y) in badpix:
+            # If (x, y) is in badpix, but did not get flagged as bad,
+            # raise an exception.
+            raise ValueError(f"Pixel ({x}, {y}) is not bad as specified by maskPlanes {maskPlanes}")
```
czwa, line 441: This function ignores entries in `goodpix` (it silently overwrites them with the true image value), but raises if the image doesn't match the `badpix`.  Shouldn't these cases be treated the same way?
While thinking through this case, I realized that `goodpix` is also susceptible to poisoning: an entry passed in `goodpix` will be used even if added to `badpix`.  That definitely needs fixed, even if the rest of the existing logic stays as is (which I'm fine with).

tests/test_interpImageTask.py
```diff
@@ -166,6 +166,133 @@ def testTranspose(self):
...
+        for n in range(n_runs):
+            task.run(self.maskedimage)
+
+        # Assert that the mask and the variance planes remain unchanged.
+        self.assertImagesEqual(self.maskedimage.variance, self.reference.variance)
+        self.assertMasksEqual(self.maskedimage.mask, self.reference.mask)
+
+        # Check that the long streak of bad pixels have been replaced with the
+        # fillValue, but not the short streak.
+        np.testing.assert_array_equal(self.maskedimage.image[0:1, :50].array, config.fillValue)
```
czwa, line 243: Won't this only check half the column?  Same question for the other test as well.

czwa (approved):
A couple typos, a few questions, and one potential bug that should be easy to resolve.

python/lsst/pipe/tasks/interpImage.py
```diff
@@ -262,3 +272,172 @@ def interpolateImage(self, maskedImage, psf, defectList, fallbackValue):
...
+class CloughTocher2DInterpolateConfig(pexConfig.Config):
+    """Config for CloughTocher2DInterpolateTask."""
+
+    badMaskPlanes = pexConfig.ListField[str](
+        doc="List of mask planes to interpolate over.",
+        default=["BAD", "SAT", "CR"],
+    )
+    fillValue = pexConfig.Field[float](
+        doc="Constant value to fill outside of the convex hull of the good "
+        "pixels. A long streak of bad pixels at an edge maybe set to this "
```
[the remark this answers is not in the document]
arunkannawadi: I used "may be", because I haven't defined what long means.  Perhaps I could define it in terms of `buffer` variable and change it to "will be".

python/lsst/pipe/tasks/interpImage.py
arunkannawadi: The use case here is we have a collection of images that have identical mask planes, but different image planes. So we shouldn't be treating `badpix` (that comes from mask) in the same way as `goodpix` (that comes from both the mask and image planes) the same way. But yes, agree that we should check that `goodpix` and `badpix` are mutually exclusive.

tests/test_interpImageTask.py
arunkannawadi: I thought I had a reason, but have no idea now why I did that. The test passes if I just use the full column.

PR #881, opened 2024-01-23
DM-42529: Use the accelerated utility functions in CloughTocher2DInterpolatorTask

arunkannawadi:


arunkannawadi:
Given that most of the heavy-lifting is done in `meas_algorithms`, I'm inclined to move this entire task there. If we'd rather have this in this repo, I can put it in its own module.

TallJimbo (approved):
Looks good, aside from the linter errors.

:+1: to moving this to meas_algorithms.  Probably best to make the new module _not_ lifted to package scope in `__init__.py`, just because we probably should have done that (to reduce unnecessary import-time logic) for most things already in meas_algorithms; better to start late than never.

PR #883, opened 2024-01-27
DM-42080: Add effTime to computeExposureSummaryStats

kadrlica:


python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -92,6 +92,29 @@ class ComputeExposureSummaryStatsConfig(pexConfig.Config):
         "robutsness metric calculations (namely, maxDistToNearestPsf and psfTraceRadiusDelta).",
         default=("BAD", "CR", "EDGE", "INTRP", "NO_DATA", "SAT", "SUSPECT"),
     )
+    fiducialSkyBackground = pexConfig.DictField(
+        keytype=str,
+        itemtype=float,
+        doc="Fiducial sky background level (cts/s) assumed when calculating effective exposure time.",
+        default={'u': 1.0, 'g': 1.0, 'r': 1.0, 'i': 1.0, 'z': 1.0, 'y': 1.0},
```
erykoff, line 99: It's worth putting in the docstring that the key is `band`.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -151,6 +174,12 @@ class ComputeExposureSummaryStatsTask(pipeBase.Task):
     (against, e.g., extrapolation instability):
     - maxDistToNearestPsf
     - psfTraceRadiusDelta
+
+    These quantities are computed to assess depth:
+    - effTime
+    - effTimeScalePsfSigma
```
erykoff, line 180: And similar renaming as for afw to `effTimePsfSigmaScale` (and below).

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -471,6 +502,97 @@ def update_masked_image_stats(self, summary, masked_image):
...
+        exposureTime = exposure.getInfo().getVisitInfo().getExposureTime()
+        filterLabel = exposure.getFilter()
+        if (filterLabel is None) or (not filterLabel.hasBandLabel):
+            band = None
+        else:
+            band = filterLabel.bandLabel
+
+        # PSF component
+        if np.isnan(summary.psfSigma):
+            self.log.warn("PSF sigma is NaN")
```
erykoff, line 552: I don't love having a warning here.  Presumably we'll have warnings upstream for bad psfs already, and doubling them up is not usually helpful.  I'd even make this a `debug`.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -471,6 +502,97 @@ def update_masked_image_stats(self, summary, masked_image):
...
+            f_eff = nan
+        elif band not in self.config.fiducialPsfSigma:
+            self.log.warn(f"Fiducial PSF value not found for {band}")
```
erykoff, line 555: Same here.  It _may_ be worth doing a warn above if band is None ... wouldn't that mean that none of this could be computed?  If so, then a warning of "no band associated with this filter, and effective exposure time statistics won't be available" may be okay (and would point to some terrible problem somewhere).

erykoff (changes requested):

python/lsst/pipe/tasks/computeExposureSummaryStats.py
kadrlica: Good point. I think that if the band is None then it is probably worth warning and returning with the effTime values set to NaN. On the other hand, if the band exists but is not in (all of) the fiducial dictionaries, it was less clear to me what should be done. I guess the safest thing is to return NaN unless fiducial values exist for *all three* components for the band of interest.

erykoff (approved):

PR #886, opened 2024-02-04
DM-42776: Add star count number to the metadata of bright star stamp and extended PSF models

bazkiaei:


python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -554,10 +554,15 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
                 "No detector groups were provided to MeasureExtendedPsfTask; computing a single, "
                 "constant extended PSF model over all available observations."
             )
+            self.metadata["psfStarCount"] = 0
+            for brightStarStamps in bss_ref_list:
+                stamps = brightStarStamps.get()
+                self.metadata["psfStarCount"] += len(stamps)
```
leeskelvin, line 560: This produces a change in behavior dependent upon whether the user has requested a region-based ePSF, or a per-detector/per-region based ePSF. That might cause confusion down the line when interpreting this metadata value, requiring an `if/else` or a `try/except`, or somehow parsing the `detectors_focal_plane_regions` parameter further downstream.

My suggestion instead is to also construct a `dict` here with a single entry. The key can be something simple like `full` or `all` (or something else you feel appropriate), and the value can be this summed number. Then, downstream, we can simply iterate over all `psfStarCountDict.items()` or `psfStarCountDict.values()` to get the metadata we need. Does that sound like a good idea here?

python/lsst/pipe/tasks/extended_psf.py
```diff
@@ -566,6 +571,10 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
                         region_name,
                     )
                     continue
+                self.metadata["psfStarCount"][region_name] = 0
+                for brightStarStamps in ref_list:
+                    stamps = brightStarStamps.get()
+                    self.metadata["psfStarCount"][region_name] += len(stamps)
```
leeskelvin, line 577: As you did for `validStarCount` and `allStarCount`, I'd suggest also adding a small in-line comment here to describe what this `psfStarCount` code block is doing. (Both here, and replicated above for the non-regions approach).

leeskelvin (approved):

python/lsst/pipe/tasks/extended_psf.py
bazkiaei: Yes, I agree with you and it sounds like a good idea! I made modifications based on your comment.

leeskelvin: Great, thanks Amir, this looks great!

PR #890, opened 2024-02-14
DM-42663: Use moments based classifier in isolatedStarAssociationTask.

erykoff:
The selection/threshold is updated in meas_algorithms, but we also need to update the transformed column name for the isolated star association task.

### DIFF
--- a/tests/test_calibrateImage.py
+++ b/tests/test_calibrateImage.py
@@ -116,8 +116,8 @@ class CalibrateImageTaskTests(lsst.utils.tests.TestCase):
         # find_stars needs an id generator.
         self.id_generator = lsst.meas.base.IdGenerator()
 
-        # Something about this test dataset prefers the older fluxRatio here.
-        self.config.star_catalog_calculation.plugins['base_ClassificationExtendedness'].fluxRatio = 0.925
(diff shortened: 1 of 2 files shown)


tests/test_calibrateImage.py
```diff
@@ -111,8 +111,8 @@ def setUp(self):
         # will use those fluxes here, and hopefully can remove this.
         self.config.astrometry.magnitudeOutlierRejectionNSigma = 9.0
 
-        # Something about this test dataset prefers the older fluxRatio here.
-        self.config.star_catalog_calculation.plugins['base_ClassificationExtendedness'].fluxRatio = 0.925
+        # Something about this test dataset prefers a larger threshold here.
+        self.config.star_selector["science"].unresolved.maximum = 0.2
```
parejkoj, line 115: Did you confirm that this was necessary, or is this just an extrapolation from how the previous value was different from the default?

erykoff: Sadly, it is necessary, and presumably for similar reasons.

arunkannawadi (approved):

TallJimbo (approved):

PR #891, opened 2024-02-15
DM-40061: Add parquet output to new CalibrateImageTask

parejkoj:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -134,6 +141,7 @@ def __init__(self, *, config=None):
         super().__init__(config=config)
         if not config.optional_outputs:
             self.outputs.remove("psf_stars")
+            self.outputs.remove("psf_stars_footprints")
```
TallJimbo, line 144: FWIW, you can also say
```
del self.psf_stars_footprints
```
if you find that more expressive.  This is moderately new so it's not in common use, but I think I find it a bit more intuitive, given that the connections generally appear as attributes.

TallJimbo (approved):

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: Oh, that's nice! Yes, I'll use that.

PR #892, opened 2024-02-16
DM-41600: Write task topic page for CalibrateImageTask

parejkoj:
I also found and fixed a bug: I'd somehow rebased out the use of the code that matched the psf_stars to set `calib_psf*` on the main output catalog. In the interest of not having yet another branch to be careful about rebasing, I'm doing the fix here.


erinleighh (approved):
Everything looks good, the writeup was well structured and readable. All the links worked, too! If you're feeling generous, there's a missing period at the end of a doc string on line 216 of `calibrateImage.py` but I didn't see any issue with any of the changes you made in this ticket.

PR #896, opened 2024-02-27
DM-43042: Fix indexing bug in _match_psf_stars

parejkoj:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -717,13 +717,14 @@ def _match_psf_stars(self, psf_stars, stars):
         # Closest matches is a dict of psf_stars source ID to Match record
         # (psf_stars source, sourceCat source, distance in pixels).
         best = {}
-        for match0, match1, d in matches:
-            id0 = match0.getId()
+        for match_psf, match_stars, d in matches:
+            id0 = match_psf.getId()
```
isullivan, line 721: With the renaming, and since there is no `id1`, I'd suggest just calling this `id` here.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -717,13 +717,14 @@ def _match_psf_stars(self, psf_stars, stars):
...
             match = best.get(id0)
             if match is None or d <= match[2]:
-                best[id0] = (match0, match1, d)
+                best[id0] = (match_psf, match_stars, d)
         matches = list(best.values())
-        ids = np.array([(match0.getId(), match1.getId()) for match0, match1, d in matches]).T
+        # We'll use this to construct index arrays into each catalog.
+        ids = np.array([(match_psf.getId(), match_stars.getId()) for match_psf, match_stars, d in matches]).T
```
isullivan, line 727: This looks overly complicated, especially since you use `ids[0]` and `ids[1]` separately. Could this just be:
`psf_ids = matches[0].getId()`
`star_ids = matches[1].getId()`

isullivan (approved):
Looks good. Please consider simplifying the line that sets `ids` by splitting it to two lines.

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: Unfortunately, no. `matches` is a list of tuples, and it's a pain to work with. This has always been a difficulty with the 2014-era `matchXy` API.

I agree it's complicated, but this is the only way I know of to do this without iterating through the match list twice.

PR #897, opened 2024-02-29
DM-43080: Remove maskStreaks from CalibrateImageTask

parejkoj:
This includes a temporary workaround to not break downstream tasks while [DM-43083](https://jira.lsstcorp.org/browse/DM-43083) is in progress, to be removed once masking happens in diffim.

[DM-43083]: https://rubinobs.atlassian.net/browse/DM-43083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrawls (approved):
Looks OK, thanks for adding the mask plane so downstream places expecting it won't break for now!

PR #899, opened 2024-03-12
DM-43247: Raise if no psf_stars match to selected stars

parejkoj:
We may eventually want to turn this into a warning, but for now we need to know when this happens, as we try to understand how to tune star selection.


hsinfang (approved):
LGTM. Good to have a more useful error message!

PR #900, opened 2024-03-12
DM-42811: Log unexpected measurement exceptions as WARNING

parejkoj:


tests/test_calibrate.py
```diff
@@ -216,7 +216,9 @@ def testNoAperCorrMap(self):
         calibTask = CalibrateTask(config=calibConfig)
         with self.assertLogs(level=logging.WARNING) as cm:
             _ = calibTask.run(charImResults.exposure)
-        self.assertIn("Image does not have valid aperture correction map", cm.output[0])
+        # Other warnings may also be emitted.
+        warnings = '\n'.join(cm.output)
+        self.assertIn("Image does not have valid aperture correction map", warnings)
```
TallJimbo, line 221: I'm surprised to see this particular message in a warning at this stage.  Is this because we're trying to continue even in the absence of aperture corrections, and we can only represent that by setting it to `None` on the `Exposure`?  If not, it's a pipeline configuration error and that should be an exception.

I realize this is preexisting, but we should probably do something about it, even if it's "create a ticket".

TallJimbo (approved):

tests/test_calibrate.py
parejkoj: `CalibrateTask` is a whole mess of "do this when that but this except that catch and log debug" nonsense. I'd rather spend as little time on it as possible, and just focus on `CalibrateImageTask` from here on.

PR #906, opened 2024-03-17
DM-43231: Create Analysis Tool to send visitSummary info to sasquatch

jrmullaney:
Added the ability to create and write metrics from the calexp summary stats. The task to calculate the metrics must be retargeted at runtime to avoid circular imports in calibrate. This is the reason for the private _EmptyTargetTask. This gets retargeted to CalexpSummaryTask in analysis tools, which is what creates the MetricMeasurementBundle that gets written to sasquatch.

### DIFF
--- a/python/lsst/pipe/tasks/calibrate.py
+++ b/python/lsst/pipe/tasks/calibrate.py
@@ -47,12 +47,19 @@ from .computeExposureSummaryStats import ComputeExposureSummaryStatsTask
 
 
 class _EmptyTargetTask(pipeBase.PipelineTask):
+    """
+    This is a placeholder target for CreateSummaryMetrics and must be retargeted at runtime.
+    CreateSummaryMetrics should target an analysis tool task, but that would, at the time
+    of writing, result in a circular import.
+
+    As a result, this class should not be used for anything else.
+    """
     ConfigClass = pipeBase.PipelineTaskConfig
 
     def __init__(self, **kwargs) -> None:
         raise NotImplementedError(
-            "doCreateSummaryMetrics is set to True, in which case"
-            "CreateSummaryMetrics must be retargeted."
+            "doCreateSummaryMetrics is set to True, in which case "
+            "createSummaryMetrics must be retargeted."
         )
 
 
--- a/python/lsst/pipe/tasks/calibrate.py
+++ b/python/lsst/pipe/tasks/calibrate.py
@@ -50,7 +50,10 @@ class _EmptyTargetTask(pipeBase.PipelineTask):
     ConfigClass = pipeBase.PipelineTaskConfig
 
     def __init__(self, **kwargs) -> None:
-        raise NotImplementedError("CreateSummaryMetrics must be retargeted")
+        raise NotImplementedError(
+            "doCreateSummaryMetrics is set to True, in which case"
+            "CreateSummaryMetrics must be retargeted."
+        )
 
 
 class CalibrateConnections(pipeBase.PipelineTaskConnections, dimensions=("instrument", "visit", "detector"),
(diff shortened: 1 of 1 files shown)


ctslater (commented):
Minor comments

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -46,6 +46,13 @@
 from .computeExposureSummaryStats import ComputeExposureSummaryStatsTask
 
 
+class _EmptyTargetTask(pipeBase.PipelineTask):
+    ConfigClass = pipeBase.PipelineTaskConfig
+
+    def __init__(self, **kwargs) -> None:
+        raise NotImplementedError("CreateSummaryMetrics must be retargeted")
```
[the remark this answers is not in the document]
jrmullaney: No, it doesn't get raised when `doCreateSummaryMetrics` is false and when CreateSummaryMetrics hasn't been retargets. The `CreateSummaryMetrics` task is only made if `doCreateSummaryMetrics` is set to True (see [this line](https://github.com/lsst/pipe_tasks/blob/5563584ad84321bfec6c15eecfcc41e50e7f946a/python/lsst/pipe/tasks/calibrate.py#L476)).

I agree with your sentiments about telling the user why that message has been flagged-up and will edit accordingly.

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -46,6 +46,16 @@
 from .computeExposureSummaryStats import ComputeExposureSummaryStatsTask
 
 
+class _EmptyTargetTask(pipeBase.PipelineTask):
```
natelust, line 49: I would still add a doctoring to this to indicate it is only a placeholder, why, and that no one should use it (I know the _ is supposed to mean that, but more documentation is always better)

python/lsst/pipe/tasks/calibrate.py
```diff
@@ -121,6 +131,13 @@ class CalibrateConnections(pipeBase.PipelineTaskConnections, dimensions=("instru
         dimensions=("instrument", "visit", "detector"),
     )
 
+    outputSummaryMetrics = cT.Output(
+        doc="Summary metrics created by the calibration task",
+        name="calexpSummary_metrics",
+        storageClass="MetricMeasurementBundle",
+        dimensions=("instrument", "visit", "detector"),
+    )
+
```
natelust, line 140: I would write out the SummaryStats too if someone wants to create the metrics later. I don't know why we already don't, maybe @erykoff knows. Is it just because we ARE writing it out with the exposure?

natelust (approved):

python/lsst/pipe/tasks/calibrate.py
erykoff: I'm actually confused about this whole thing.  You can get the summary stats as a component of a calexp with `calexp.summaryStats` in a connection without reading the full calexp.  I'm not sure why this code is in `calibrate.py` at all (beyond the circular import problems).  Furthermore, we are moving from `calibrate.py` to `calibrateImage.py` (which is already the default in AP) but I don't think this belongs there either.

jrmullaney: It's so that the SummaryStats are written to metrics during the calibrate task, which I understand is something that both @ctslater and @mfisherlevine would like. I agree that grabbing the stats is easy and quick post-calibrate, but it still involves read overhead, which is avoided by having it in-task like this. AFAIA, it's always been a feature of analysis_tools that they can be used in-task like this.

The reason it's in calibrate.py rather than calibrateImage.py is simply because this is where I started developing it. It can be easily added to calibrateImage.py, and I think @ctslater is planning to make a ticket to do that.

PR #909, opened 2024-03-21
DM-41210: Generalize MatchTractCatalogTask

jtmccann:


python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -37,6 +37,7 @@
 
 from abc import ABCMeta, abstractmethod
 from astropy.stats import mad_std
+import astropy.table as apTab
```
leeskelvin, line 40: This doesn't seem like standard practice. It looks like you only need `Table` and `vstack` anyway, so I'd recommend changing to:
```
from astropy.table import Table, vstack
```
and modifying the relevant code that calls it below.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -107,18 +108,23 @@ class DiffMatchedTractCatalogConnections(
         dimensions=("tract", "skymap"),
     )
     cat_matched = cT.Output(
-        doc="Catalog with reference and target columns for matched sources only",
+        doc="Catalog with reference and target columns for joined sources",
```
leeskelvin, line 111: Why the switch from "matched" to "joined" terminology?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -664,247 +667,288 @@ def run(
...
-                            select_target_sub = select_target_bin.copy()
-                            if is_extended is not None:
-                                is_extended_ref = (extended_ref == is_extended)
-                                select_ref_sub &= is_extended_ref
+        if config.include_unmatched:
+            # Create an unmatched table with the same schema as the matched one
+            # ... but only for objects with no matches (for completeness/purity)
+            # and that were selected for matching (or inclusion via config)
+            cat_right = apTab.Table.from_pandas(cat_ref[~matched_ref & select_ref].reset_index(drop=False))
+            cat_right.rename_columns(cat_right.colnames, [f"refcat_{col}" for col in cat_right.colnames])
```
leeskelvin, line 686: There's a mix of `'` and `"` in this script. I wouldn't be averse to you running the whole thing through `black` prior to a merge to resolve this (and other formatting choices).

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -664,247 +667,288 @@ def run(
...
+            match_row_target = catalog_match_target['match_row'].values
+            cat_left = cat_target[~(match_row_target >= 0) & select_target].reset_index(
+                drop=not has_index_left)
+            # This may be slower than pandas but will, for example, create
+            # masked columns for booleans, which pandas does not support.
+            # See https://github.com/pandas-dev/pandas/issues/46662
+            cat_unmatched = apTab.vstack([apTab.Table.from_pandas(cat_left), cat_right])
+
+        if config.output_statistics:
+            # Slightly smelly hack for when a column (like distance) is already relative to truth
```
leeskelvin, line 696: If this is smelly, what's the better way to do this? I know this isn't really a change on this PR, as it existed prior, but keen to hear thoughts on how we can make this operation less smelly in the future.

leeskelvin: @taranu :point_up:

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -107,18 +108,23 @@ class DiffMatchedTractCatalogConnections(
...
         name="matched_{name_input_cat_ref}_{name_input_cat_target}",
-        storageClass="DataFrame",
+        storageClass="ArrowAstropy",
         dimensions=("tract", "skymap"),
     )
     diff_matched = cT.Output(
         doc="Table with aggregated counts, difference and chi statistics",
         name="diff_matched_{name_input_cat_ref}_{name_input_cat_target}",
-        storageClass="DataFrame",
+        storageClass="ArrowAstropy",
```
leeskelvin, line 119: After having spoken with @taranu, it seems like these should be rolled back to `DataFrame` storage classes for the time being.

leeskelvin: PS - please do run some test data through post-change to make sure that everything still works as expected, e.g., no missing `.iloc`s anywhere!

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: After adding the option to output unmatched rows, the table no longer necessarily contains only matched objects.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -664,247 +667,288 @@ def run(
...
+            cat_right.rename_columns(cat_right.colnames, [f"refcat_{col}" for col in cat_right.colnames])
+            match_row_target = catalog_match_target['match_row'].values
+            cat_left = cat_target[~(match_row_target >= 0) & select_target].reset_index(
+                drop=not has_index_left)
+            # This may be slower than pandas but will, for example, create
+            # masked columns for booleans, which pandas does not support.
+            # See https://github.com/pandas-dev/pandas/issues/46662
+            cat_unmatched = apTab.vstack([apTab.Table.from_pandas(cat_left), cat_right])
+
+        if config.output_statistics:
```
leeskelvin, line 695: After having chatted with @taranu about this, it seems like everything in this `if` block will be deprecated at some future date. In that case, I'm less interested in addressing any concerns within at this point.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
leeskelvin: Update: it seems as though all of this `if` block will be removed at some future point, so I'm not so interested any longer in addressing my question above.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: Hmm... I'll consider it depending on how many changes there are. Sometimes I wish older pipelines packages would adopt middleware/atools-like pre-commit and do a run through.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: Yeah, I'd rather deprecate the whole thing than spend more energy on fixes like this.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: This is a moot point now but I really prefer to have a comprehensible package prefix when importing a name as generic as `Table` (though fair enough if you think `apTab` is too concise, but it is in line with how we shorten stack package names).

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: See https://rubinobs.atlassian.net/browse/DM-43831.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
leeskelvin: Nevertheless, `from astropy.table import Table` is a fairly standard practice across the stack, and we do seem to be moving away from `import X as Y` imports. It also helps a) keep line lengths shorter, which b) improves readability (imo), and c) makes it clear if an extra name is now required on a PR following a subsequent edit.

python/lsst/pipe/tasks/match_tract_catalog.py
```diff
@@ -32,6 +32,7 @@
 
 from abc import ABC, abstractmethod
 
+import astropy.table as apTab
```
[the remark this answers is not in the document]
leeskelvin: I still think this looks cleaner as I've suggested, for the reasons outlined above, but will not insist if you don't want to change it. This is how astropy have documented this import in their docs, e.g.: https://docs.astropy.org/en/stable/table/io.html.

leeskelvin (approved):

PR #910, opened 2024-03-28
DM-42870: Refactor multiband fit tasks to allow fit reconstruction

taranu:


python/lsst/pipe/tasks/fit_coadd_psf.py
```diff
@@ -38,7 +39,8 @@ class CatalogExposurePsf(CatalogExposure):
     def get_catalog(self):
         return self.catalog
 
-    def get_psf_image(self, source):
+    def get_psf_image(self, source, **kwargs):
```
fred3m, line 42: You added `kwargs` but never seem to use them. Was this just to make this a uniform API with something else?

fred3m (approved):
One minor comment

python/lsst/pipe/tasks/fit_coadd_psf.py
taranu: Whoops... yes, that was the original intent but it turned out to not be necessary.

PR #911, opened 2024-03-29
DM-43545: Ensure that images without psfs/aperture corrections are not used in warps.

erykoff:
This also partly covers the deprecation cleanup from DM-39854, at least as far as makeWarp.py is concerned.


TallJimbo (approved):
Looks great; thanks for including the deprecation-removals, too.

PR #915, opened 2024-04-11
DM-39841: Remove deprecated doPsfMatch field

arunkannawadi:


erykoff (approved):

arunkannawadi:
It's scary that Jenkins/unit tests don't catch this. I shouldn't have removed `modelPsf` field because that's still being used if `makePsfMatched` is set to `True`.

arunkannawadi:
It should be moved to `makeWarp.py` instead where it may be used if `makePsfMatched` is set to `True`.

PR #917, opened 2024-04-12
DM-43771: Disable warping kernel caches and require consistency with PSF warping.

TallJimbo:


python/lsst/pipe/tasks/makeWarp.py
```diff
@@ -176,9 +176,17 @@ def validate(self):
             log.warning("Config doPsfMatch deprecated. Setting makePsfMatched=True and makeDirect=False")
             self.makePsfMatched = True
             self.makeDirect = False
+        if self.warpAndPsfMatch.warp.cacheSize != self.coaddPsf.cacheSize:
+            # This is an incomplete check: usually the CoaddPsf cache size
+            # configured here in MakeWarpTask is superseded by the one in
+            # AssembleCoaddTask.  A pipeline contract in the drp_pipe is
+            # present to check that.
+            raise RuntimeError("Image warping cache size and CoaddPSf warping cache size do not agree.")
```
arunkannawadi, line 184: Raising `FieldValidationError` would be the right thing to do (that class was implemented specifically for this, AFAIK) but given that `RuntimeError` is raised above for another check, I wouldn't be strongly opposed to raising it here as well.

arunkannawadi (approved):

python/lsst/pipe/tasks/makeWarp.py
TallJimbo: `FieldValidationError` is a little too specific to errors in individual fields (rather than consistency across multiple fields) to be used here cleanly.  But I see that it inherits from `ValueError`, so that's probably a better built-in to raise than `RuntimeError`.  I'll also fix the one above the same way.

PR #921, opened 2024-04-23
DM-43935: Enable higher order moments plugins

arunkannawadi:
in finalizeCharacterization.


Alex-Broughton (approved):
See https://rubinobs.atlassian.net/browse/DM-43935

PR #922, opened 2024-04-24
DM-43973: Remove ClassificationSizeExtendedness from plugin

arunkannawadi:
list to be run before having a PSF model.


parejkoj (approved):
I've filed DM-44061 to deal with the other cause of this (`slot_PsfShape` isn't checked for field validity when it is set). We can deal with the the unsetting of that slot in `self.psf_source_measurement` here once that ticket is merged.

arunkannawadi:
Quick question - why do we need to unset `slot_PsfShape` necessarily?

parejkoj:
If it's pointing to an invalid field, it shouldn't be set at all (hence DM-44061, to make sure it can't be set to something invalid).

PR #928, opened 2024-05-08
DM-44246: Use full set of src footprints for noise replacement in FinalizeCharacterizationTask.

erykoff:
This also adds additional comments about why `includeThresholdMultiplier` is an important config.

### DIFF
--- a/python/lsst/pipe/tasks/calibrateImage.py
+++ b/python/lsst/pipe/tasks/calibrateImage.py
@@ -265,6 +265,11 @@ class CalibrateImageConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Cali
         self.install_simple_psf.fwhm = 4
 
         # S/N>=50 sources for PSF determination, but detection to S/N=5.
+        # The thresholdValue sets the minimum flux in a pixel to be included in the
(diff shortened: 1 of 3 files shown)


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -265,6 +265,11 @@ def setDefaults(self):
         self.install_simple_psf.fwhm = 4
 
         # S/N>=50 sources for PSF determination, but detection to S/N=5.
+        # These thresholds are set such that the effective
+        # detection threshold is 50sigma (5*10) but the footprints are very large
+        # (out to 5 sigma).  This ensures that the noise replacer does a reasonable
+        # job during measurement in replacing neighbors with noise, even if those
+        # neighbors are not sources to be measured.
```
fred3m, line 272: Maybe re-word this and the other comment slightly by being a little more explicit. Something like "The `thresholdValue` sets the minimum flux in a pixel to be included in the footprint, while peaks are only detected when they are above `thresholdValue * includeThresholdMultiplier`. The low `thresholdValue` ensures that the footprints are large enough for the noise replacer to mask out faint undetected neighbors."

fred3m (approved):

PR #929, opened 2024-05-09
DM-43332: Add support for spherical coord distances in DiffMatchedTractCatalogTask

taranu:


python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -44,7 +44,8 @@
 import numpy as np
 import pandas as pd
 from scipy.stats import iqr
-from typing import Dict, Sequence
+from smatch.matcher import sphdist
+from typing import Sequence
```
enourbakhsh, line 48: Now that we're getting rid of the `Dict` import from `typing` on this ticket, we should tidy up the docstrings in a couple of places: `compute_stats()` and `_get_columns()`.

enourbakhsh (commented):
Looks good - see the minor comments below.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -685,13 +685,24 @@ def run(
...
+
         dist_err = np.full(n_target, np.nan)
-        dist_err[matched_row] = np.hypot(cat_target.iloc[matched_row][coord1_target_err].values,
-                                         cat_target.iloc[matched_row][coord2_target_err].values)
+        dist[matched_row] = sphdist(
+            target_match_c1, target_match_c2, target_ref_c1, target_ref_c2
+        ) if config.coord_format.coords_spherical else np.hypot(
+            target_match_c1 - target_ref_c1, target_match_c2 - target_ref_c2,
+        )
+        # Should probably explicitly add cosine terms if ref has errors too
```
enourbakhsh, line 697: Are we ignoring the errors in the ref catalog because they're so tiny they don't really matter, or is it just overkill? Or are you thinking we should start paying attention to them sometime soon?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -685,13 +685,24 @@ def run(
...
+        dist_err[matched_row] = sphdist(
+            target_match_c1, target_match_c2,
+            target_match_c1 + cat_target.iloc[matched_row][coord1_target_err].values,
+            target_match_c2 + cat_target.iloc[matched_row][coord2_target_err].values,
+        ) if config.coord_format.coords_spherical else np.hypot(
```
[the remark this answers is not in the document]
taranu: It is, but you raise a good point that I have to update the reference data because the spherical difference are different from the old hypot.

Just to confirm, only the distances change, i.e.:
`{k: v for k, v in dict(result.diff_matched.iloc[0] - self.diff_matched).items() if 'dist' not in k and (v != 0)}` prints nothing.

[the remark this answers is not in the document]
taranu: I will add a second part to the test to set that config value to the opposite of the default and run again. Having said that, we don't generally test every single config permutation in tasks; it's only viable here since it's a boolean.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: That part of the code is slated for deprecation in RFC-1008 and so I'm not updating any of it.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: Right now this is only used for DC2 and source injection, where the reference catalogs are truth with no errors. It will take a lot more work on a separate ticket to incorporate reference catalog errors.

PR #930, opened 2024-05-09
DM-44268: Add ra/dec error columns to characterizeImage output

mfisherlevine:


natelust:
Be sure to run all the CI packages as schema changes can cause weird breakages. For instance that other fix someone did in T/S might now break because the columns will now already be there.

natelust (approved):

mfisherlevine:
> Be sure to run all the CI packages as schema changes can cause weird breakages. For instance that other fix someone did in T/S might now break because the columns will now already be there.

Yup, great point, thanks. I'll run the works 👍

PR #932, opened 2024-05-16
DM-44382: Update GAIA refcat to DR3 in subtractBrightStars

leeskelvin:


python/lsst/pipe/tasks/subtractBrightStars.py
```diff
@@ -94,7 +94,7 @@ class SubtractBrightStarsConnections(
     )
     refCat = PrerequisiteInput(
         doc="Reference catalog that contains bright star positions",
-        name="gaia_dr2_20200414",
+        name="gaia_dr3_20230707",
```
bazkiaei, line 97: This is the only change in this PR and I see no problem with that.

bazkiaei (approved):
Looks good to me.

PR #933, opened 2024-05-17
DM-43490: Increase the kernelSize if PIFF is modeling PSF in sky coordinates

arunkannawadi:


PFLeget (approved):
This looks good to me!

PR #934, opened 2024-05-21
DM-43064: Make PSF-matched warp from DirectWarp

arunkannawadi:


arunkannawadi:
I need to put the new unit tests in its own file and fix some minor linting errors, but this branch is ready for a review.

python/lsst/pipe/tasks/coaddInputRecorder.py
```diff
@@ -95,6 +95,11 @@ def addCalExp(self, calExp, ccdId, nGoodPix):
             Number of good pixels this image will contribute to the CoaddTempExp.
             If saveEmptyCcds is not set and this value is zero, no record will be
             added.
+
+        Returns
+        -------
+        record : `lsst.afw.image.ExposureRecord`
+            The record corresponding to this CCD.
```
arunkannawadi, line 102: Oh, gotta document that `None` may be returned in some corner cases.

python/lsst/pipe/tasks/make_psf_matched_warp.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+
+    # matchingKernelSize is there just for backward compatibility.
+    @property
+    def matchingKernelSize(self):
+        return self.psfMatch.kernel.active.kernelSize
+
+    @matchingKernelSize.setter
+    def matchingKernelSize(self, value):
+        self.psfMatch.kernel.active.kernelSize = value
```
arunkannawadi, line 94: I think support for `matchingKernelSize` can be dropped altogether in this pipeline. We are supposed to directly set the `kernelSize` in the subtask instead of wrapping it as another config field. If we continue to support, at least we should warn that setting `matchingKernelSize` is not the recommended method.

python/lsst/pipe/tasks/make_psf_matched_warp.py
```diff
@@ -0,0 +1,215 @@
...
+                )
+                destination_polygon = src_polygon.transform(transform).intersectionSingle(
+                    geom.Box2D(direct_warp.getBBox())
+                )
+                self.log.trace("Polygon for detector=%d is calculated as %s",
+                               row["ccd"],
+                               destination_polygon
+                               )
+            else:
+                self.log.trace("Polygon for detector=%d is read from the input calexp as %s",
```
TallJimbo, line 165: I'd just make these logs `debug` instead of `trace`.  There wouldn't been all that many of them if they were on, and `trace` level is a little obscure.

python/lsst/pipe/tasks/make_psf_matched_warp.py
```diff
@@ -0,0 +1,215 @@
...
+                exposure_psf_matched.maskedImage[bbox],
+                temp_psf_matched.maskedImage[bbox],
+                bit_mask,
+            )
+
+            del temp_psf_matched
+
+            self.log.trace(
+                "Copied %d pixels from temp_warp to exposure_psf_matched", num_good_pixels
+            )
```
TallJimbo, line 210: Might even be worth making this INFO-level if you can put a data ID or something similar in; there'd still only be <= 4 of them per quantum.

TallJimbo (approved):

python/lsst/pipe/tasks/make_direct_warp.py
```diff
@@ -408,15 +410,22 @@ def run(self, inputs, sky_info, visit_summary):
...
             nGood = copyGoodPixels(
                 final_warp.maskedImage,
                 warpedExposure.maskedImage,
                 final_warp.mask.getPlaneBitMask(["NO_DATA"]),
             )
             ccdId = self.config.idGenerator.apply(dataId).catalog_id
-            inputRecorder.addCalExp(calexp, ccdId, nGood)
+            record = inputRecorder.addCalExp(calexp, ccdId, nGood)
             totalGoodPixels += nGood
+            record.setValidPolygon(polygon)
```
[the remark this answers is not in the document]
arunkannawadi: It just seemed appropriate that a warp should be able to separate out the detectors. I do this calculation in the PSF-matching task if this information is not available, but I thought it'd be nice to persist this info with the warp itself.

python/lsst/pipe/tasks/make_psf_matched_warp.py
```diff
@@ -0,0 +1,215 @@
...
+        Each individual detector from the direct warp is isolated, one at a
+        time, and PSF-matched to the same model PSF. The PSF-matched images are
+        then added back together to form the final PSF-matched warp. The bulk
+        of the work is done by the `psfMatchTask`.
+
+        Notes
+        -----
+        Pixels that receive no inputs are set to NaN, for e.g, chip gaps. This
+        violates LSST algorithms group policy, but will be fixed during the
+        coaddition stage.
```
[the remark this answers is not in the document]
arunkannawadi: The docstring is indeed taken from the old warping task. I can remove the second half of that sentence and leave the note that it is a policy violation.

python/lsst/pipe/tasks/make_psf_matched_warp.py
```diff
@@ -0,0 +1,215 @@
...
+            # outside the detector.
+            temp_warp = direct_warp[bbox].clone()
+            temp_warp.setPsf(warp_psf)
+            temp_warp.image.array *= ccd_mask_array
+            temp_warp.mask.array |= (~ccd_mask_array) * bit_mask
+            # We intend to divide by zero outside the detector to set the
+            # per-pixel variance values to infinity. Suppress the warning.
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", message="divide by zero", category=RuntimeWarning)
+                temp_warp.variance.array /= ccd_mask_array
```
arunkannawadi, line 192: Wanted to double check if this is an anti-pattern or fine as is.

TallJimbo: If you're sure it makes `inf` rather than `nan` it doesn't bother me too much; I'm annoyed by the fact that numpy decided this should be a warning.  But it is worth checking that the results are indeed `inf`.

python/lsst/pipe/tasks/make_psf_matched_warp.py
arunkannawadi: Do we need access to data ID explicitly to make this log (at INFO level) useful. This is how it appears to me anyway

```
INFO 2024-05-22T13:04:54.489-07:00 lsst.makePsfMatchedWarp (makePsfMatchedWarp:{instrument: 'LSSTCam-imSim', skymap: 'DC2', tract: 3828, patch: 42, visit: 214464, band: 'i', physical_filter: 'i_sim_1.4'})(make_psf_matched_warp.py:208) - Copied 3249155 pixels from temp_warp to exposure_psf_matched
```

This method knows nothing about the data ID and it won't unless we override the `runQuantum` method.

arunkannawadi: Oh, I suppose you meant the `ccd` id of the input `calexp`.

python/lsst/pipe/tasks/make_direct_warp.py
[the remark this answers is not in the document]
arunkannawadi: This actually seems to be causing harm. `CoaddPsf` is unable to find the average position with this being set.

python/lsst/pipe/tasks/make_psf_matched_warp.py
TallJimbo: Yes, the visit ID isn't needed because one can get that from task start/stop logs, but he detector would make the logging a bit more useful.

python/lsst/pipe/tasks/make_direct_warp.py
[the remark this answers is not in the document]
TallJimbo: I'm not sure why that is, but it sounds like the safe thing to do is just remove it.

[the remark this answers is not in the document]
arunkannawadi: Well, to be fair, it's causing issues in unit tests and it wasn't evident when I ran it on some DC2 data. The unit tests aren't robust, so it could actually be an issue with the tests. In any case, punting it to a different ticket since I don't _need_ this.

PR #936, opened 2024-05-27
DM-37952: Add psfApCorrDelta & psfApFluxDelta metrics to exposure summaries

laurenam:


erykoff (changes requested):
Some small comments.  The bigger one is that I don't love that the quantity being recorded in the summary is *not* scaled by the psf sigma, but the quantity being cut on *is* scaled.  I think this may lead to confusion downstream.  In fact, looking back, I think the `analysis_tool` plot is unscaled, but the cut is scaled?  So I think this needs to be fixed all the way through (sorry).

laurenam:
No worries…I actually waffled on this as there is precedence (here’s looking at you `medianE`!)  But I do agree that it is confusing and a bad precedent, so I’ll change it to be the scaled version throughout.

(Also, I didn’t actually add it to any `analysis_tools` plots, but I would've used the scaled version, à la: https://github.com/lsst/analysis_tools/blob/c7e645f379c4a59b9dcf54a2dcbd930e88e74cbf/python/lsst/analysis/tools/actions/plot/multiVisitCoveragePlot.py#L300-L303).

erykoff (approved):

PR #939, opened 2024-05-31
DM-44535: Express dipoleAngle, trailAngle in position angle on sky instead of detector angle

wmwv:
Create new functor `ConvertDetectorAngleToPositionAngle` to calculate position angle on sky (degrees E of N in sky coordinates), instead of orientation (radians +y from +x in detector coordinates).  Uses `SpherePoint.bearingTo` for the spherical trig.

I was a little perhaps overeager with the tests, but I have this fear, based on experience, of messing up spherical trig.

@isullivan Do you agree this is doing the right thing?

@leeskelvin Could you take a particular look at the `tests/test_functors.py`.  Your the last person to look at this part of the code.

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -26,7 +26,9 @@ __all__ = ["init_fromDict", "Functor", "CompositeFunctor", "mag_aware_eval",
            "PsfSdssTraceSizeDiff", "HsmTraceSize", "PsfHsmTraceSizeDiff",
            "HsmFwhm", "E1", "E2", "RadiusFromQuadrupole", "LocalWcs",
            "ComputePixelScale", "ConvertPixelToArcseconds",
-           "ConvertPixelSqToArcsecondsSq", "ReferenceBand", "Photometry",
+           "ConvertPixelSqToArcsecondsSq",
+           "ConvertDetectorAngleToPositionAngle",
+           "ReferenceBand", "Photometry",
            "NanoJansky", "NanoJanskyErr", "LocalPhotometry", "LocalNanojansky",
            "LocalNanojanskyErr", "LocalDipoleMeanFlux",
            "LocalDipoleMeanFluxErr", "LocalDipoleDiffFlux",
@@ -1203,7 +1205,7 @@ class LocalWcs(Functor):
         super().__init__(**kwargs)
 
     def computeDeltaRaDec(self, x, y, cd11, cd12, cd21, cd22):
-        """Compute the distance on the sphere from x2, y1 to x1, y1.
+        """Compute the dRA, dDec from dx, dy.
 
         Parameters
         ----------
@@ -1211,8 +1213,6 @@ class LocalWcs(Functor):
             X pixel coordinate.
         y : `~pandas.Series`
             Y pixel coordinate.
-        cd11 : `~pandas.Series`
-            [1, 1] element of the local Wcs affine transform.
         cd11 : `~pandas.Series`
             [1, 1] element of the local Wcs affine transform.
         cd12 : `~pandas.Series`
@@ -1225,9 +1225,13 @@ class LocalWcs(Functor):
         Returns
         -------
         raDecTuple : tuple
-            RA and dec conversion of x and y given the local Wcs.
+            RA and Dec conversion of x and y given the local Wcs.
             Returned units are in radians.
 
+        Notes
+        -----
+        If x and y are with respect to the CRVAL1, CRVAL2
+        then this will return the RA, Dec for that WCS.
         """
         return (x * cd11 + y * cd12, x * cd21 + y * cd22)
 
@@ -1272,8 +1276,6 @@ class LocalWcs(Functor):
             Y pixel coordinate.
         cd11 : `~pandas.Series`
             [1, 1] element of the local Wcs affine transform.
-        cd11 : `~pandas.Series`
-            [1, 1] element of the local Wcs affine transform.
         cd12 : `~pandas.Series`
             [1, 2] element of the local Wcs affine transform.
         cd21 : `~pandas.Series`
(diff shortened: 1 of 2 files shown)


leeskelvin:
I don't think I've ever touched the particular file `tests/test_functors.py` - perhaps you're thinking of @arunkannawadi. In any case, I'm not sure I'm best placed to review it quickly. My immediate comment however is: I would strongly recommend collapsing your commits into one/a few, in place of 23!

wmwv:
DM-36182 is what got your name attached to `git blame`.
Chris Morrison is otherwise the relevant person, but I understand is no longer working for Rubin DM.

tests/test_functors.py
```diff
@@ -678,8 +827,13 @@ def testConvertPixelToArcseconds(self):
                                             atol=1e-16,
                                             rtol=1e-16))
 
-    def _makeWcs(self, dec=53.1595451514076):
-        """Create a wcs from real CFHT values.
+    def _makeWcs(self, dec=53.1595451514076, theta=0):
+        """Create a wcs from real CFHT values, rotated by an optional theta.
+
+        dec : float
```
arunkannawadi, line 833:
```suggestion
        dec : `float`
```

tests/test_functors.py
```diff
@@ -678,8 +827,13 @@ def testConvertPixelToArcseconds(self):
...
+            Set reference declination of CRVAL2 [degrees]
+        theta : float
```
arunkannawadi, line 835:
```suggestion
        theta : `float`
```

tests/test_functors.py
```diff
@@ -581,14 +584,159 @@ def testDipPhotometry(self):
...
+        local_wcs = LocalWcs(*cd_matrix.flatten())
+        pa = local_wcs.computePositionAngle(df["ra1"], df["dec1"], df["ra2"], df["dec2"])
+        expected = df["expected"]
+        tolerance_deg = 0.05  # degrees
+        tolerance_rad = np.deg2rad(tolerance_deg)
+
+        # Use SphereGeom to handle wrap-around separations correctly.
+        diff = []
+        for o, e in zip(pa, expected):
+            diff.append(geom.Angle(o, geom.radians).separation(geom.Angle(e, geom.radians)).asRadians())
```
arunkannawadi, line 655: This would work better with list comprehension as

```python
diff = [geom.Angle(o, geom.radians).separation(geom.Angle(e, geom.radians)).asRadians() for o, e in zip(pa, expected)]
```

tests/test_functors.py
```diff
@@ -581,14 +584,159 @@ def testDipPhotometry(self):
                                     atol=1e-13,
                                     rtol=0))
 
+    def testComputePositionAngle(self):
+        """Test computation of position angle from (RA1, Dec1) to (RA2, Dec2)
+        """
+
+        d = 0.0001  # [rad] offset amount, length of our vector
```
arunkannawadi, line 591: I'd recommend 1) making this an optional argument to `testComputePositionAngle` and 2) use a longer variable name, at least in the function signature, that can be referred to as `d` inside this test later on.

arunkannawadi (commented):
I'm not sure if a review is requested from me, but some comments from a quick look so far.

python/lsst/pipe/tasks/functors.py
```diff
@@ -1291,6 +1293,76 @@ def getSkySeparationFromPixel(self, x1, y1, x2, y2, cd11, cd12, cd21, cd22):
         # Great circle distance for small separations.
         return self.computeSkySeparation(ra1, dec1, ra2, dec2)
 
+    def computePositionAngle(self, ra1, dec1, ra2, dec2):
+        """Compute position angle (E of N) from (ra1, dec1) to (ra2, dec2).
+
+        Parameters
+        ----------
+        ra1 : `~pandas.Series`
```
isullivan, line 1301: I believe each of these could be any `iterable`, not just a pandas Series

isullivan (approved):
Thank you, this looks like a very helpful addition. I see Arun already left some formatting cleanups, but in addition to those I note that the new functions you added state in their docstring that they accept `pandas.Series` as inputs, but I believe any python iterable would work for most of them (a list would be problematic for `computeDeltaRaDec`, though).

tests/test_functors.py
arunkannawadi: The best way to format it with line breaks would be to run `black` on it with the line-length specified. But this repo is not fully black compliant yet, so you'd need to put this in a separate file to see what `black` would do, and that'd probably be the best solution for now. But yeah, I don't think there are any specific guidelines as to where to break otherwise.

wmwv: To be clear, I've changed this as suggested, but I would be interested in any formal style guidance.

wmwv: I have re-read the DM Python Style Guide now, and that prompted me to remove the spaces I had put around `*` and `/` operators.

leeskelvin: For reference, `black` here gives this:
```
        diff = [
            geom.Angle(o, geom.radians).separation(geom.Angle(e, geom.radians)).asRadians()
            for o, e in zip(pa, expected)
        ]
```

wmwv: Oh, thanks.  What would `black` do (with line length 110 characters) is excellent guidance.

wmwv: > I have re-read the DM Python Style Guide now, and that prompted me to remove the spaces I had put around * and / operators.

Ah, and I realize `black` is why I do put spaces around `*` and `/`.  But we're not `black` formatting this repo, so I should stick with the DM Python Style Guide and have no spaces.

wmwv:
Thanks @arunkannawadi @leeskelvin @isullivan for the helpful PR comments.  I feel welcome and supported as I get back up to speed in making commits to the Science Pipelines.

tests/test_functors.py
arunkannawadi: The suggestion to put the lines to format in a new file was an editor-agnostic one. If you use VSCode, then you could select those lines and format just those as well. I don't know which other editors have that feature.

python/lsst/pipe/tasks/functors.py
isullivan: I would suggest:
```
ra1 : `iterable` of `float`
    RA of the first coordinate [radians].
    Typically a `pandas.Series`.

PR #940, opened 2024-06-03
DM-44000: Support matching reference catalogs with no spatial sharding

taranu:


jtmccann:
Looks good to me! The changes here are simple and easy to follow.

jtmccann (approved):

PR #942, opened 2024-06-04
DM-44678: Validate matchingKernelSize for consistency.

arunkannawadi:


python/lsst/pipe/tasks/makeWarp.py
```diff
@@ -177,6 +177,15 @@ def validate(self):
             # present to check that.
             raise ValueError("Image warping cache size and CoaddPSf warping cache size do not agree.")
 
+        if self.matchingKernelSize:
+            if not self.matchingKernelSize == self.warpAndPsfMatch.psfMatch.kernel.active.kernelSize:
```
TallJimbo, line 181:
```suggestion
            if self.matchingKernelSize != self.warpAndPsfMatch.psfMatch.kernel.active.kernelSize:
```

TallJimbo (approved):

python/lsst/pipe/tasks/makeWarp.py
arunkannawadi: Thanks.

Oh, I forgot to mention - Upon deprecating `matchingKernelSize`, I'd like to make it optional and have `None` as its default value. Hence the conditioning on `self.matchingKernelSize` in anticipation of that.

PR #944, opened 2024-06-07
DM-38632: Add normalized compensated tophat fluxes as default calibration flux.

erykoff:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -164,6 +164,12 @@ class CalibrateImageConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Cali
         doc="Task to combine two snaps to make one exposure.",
     )
 
+    do_normalized_calibration = pexConfig.Field(
```
parejkoj, line 167: Is there a reason we wouldn't want to use compensated tophats here? I've avoided `do_x` configs in CalibrateImage, opting to instead just implement the algorithm that we want for each step. If this is something for which there's no real reason to turn it off, there's no need to have a `do_x` for it. That would remove a lot of the branches in the code.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -190,6 +196,11 @@ class CalibrateImageConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Cali
         target=measurePsf.MeasurePsfTask,
         doc="Task to measure the psf on bright sources."
     )
+    psf_normalized_calibration_flux = pexConfig.ConfigurableField(
+        target=lsst.meas.algorithms.NormalizedCalibrationFluxTask,
+        doc="Task to normalize the calibration flux (e.g. compensated tophats) "
+            "for the psf measurement bright stars.",
```
parejkoj, line 202: I think a word is missing here?

parejkoj (changes requested):
My main concern is about `do_normalized_calibration`, which I don't think is necessary if we're pretty sure that's the algorithm we want.

python/lsst/pipe/tasks/calibrateImage.py
erykoff: This is one of the things I wanted to ask you about.  If you think that it should not be optional, then I can clean this up.

parejkoj: Is there any reason you think it should be optional? If not, then I see no benefit to making it so.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -703,6 +749,11 @@ def _find_stars(self, exposure, background, id_generator):
 
         # Measure everything, and use those results to select only stars.
         self.star_measurement.run(sources, exposure)
+        if self.config.do_normalized_calibration:
+            self.star_normalized_calibration_flux.run(
```
parejkoj, line 753: Just checking on this: what's the point of running it again here, if we used it for the pseudo-ApCorr thing earlier?

erykoff: This is the code that is configured to only *apply* the normalization to the compensated tophats measured in `star_measurement`.  Will add a comment!

python/lsst/pipe/tasks/calibrateImage.py
erykoff: I'm willing to stand behind the new calibration fluxes, so I'll remove the optionality.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -710,6 +758,13 @@ def _find_stars(self, exposure, background, id_generator):
 
         # Measure everything, and use those results to select only stars.
         self.star_measurement.run(sources, exposure)
+        # Run the normalization calibration flux task to apply the
+        # normalization normalization correction to create normalized
```
taranu, line 762: normalization x2

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -352,6 +368,20 @@ def setDefaults(self):
         self.compute_summary_stats.starSelection = "calib_photometry_used"
         self.compute_summary_stats.starSelector.flags.good = ["calib_photometry_used"]
 
+    def validate(self):
+        super().validate()
+
+        # Ensure that the normalization calibration flux tasks
+        # are configured correctly.
+        if not self.psf_normalized_calibration_flux.do_measure_ap_corr:
+            raise RuntimeError(
```
taranu, line 377: Could/should be a `ValueError`?

tests/test_calibrateImage.py
```diff
@@ -168,7 +169,7 @@ def _check_run(self, calibrate, result):
 
         # Returned photoCalib should be the applied value, not the ==1 one on the exposure.
         self.assertFloatsAlmostEqual(result.applied_photo_calib.getCalibrationMean(),
-                                     self.photo_calib, rtol=2e-3)
+                                     self.photo_calib, rtol=1e-2)
```
taranu, line 172: Is the higher rtol worth commenting on? Is the actual value close to this or are you just adding some extra safety margin?

tests/test_isolatedStarAssociation.py
```diff
@@ -303,7 +303,8 @@ def test_make_all_star_sources(self):
                                                                              self.data_ref_dict)
 
         # Make sure we don't have any low s/n sources.
-        sn_min = np.min(source_cat['apFlux_12_0_instFlux']/source_cat['apFlux_12_0_instFluxErr'])
+        sn_min = np.min(source_cat['normCompTophatFlux_instFlux']
+                        / source_cat['normCompTophatFlux_instFluxErr'])
         self.assertGreater(sn_min, 10.0)
```
taranu, line 308: Is this for the same task that you changed to default to 8 for?

taranu (commented):
Minor comments/questions.

erykoff:
@parejkoj Are you happy with the changes on this PR?  I removed the optionality in `calibrateImage` and added appropriate comments.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -352,6 +368,20 @@ def setDefaults(self):
...
+
+        # Ensure that the normalization calibration flux tasks
+        # are configured correctly.
+        if not self.psf_normalized_calibration_flux.do_measure_ap_corr:
+            raise ValueError(
+                "psf_normalized_calibration_flux task must be configured with do_measure_ap_corr=True.",
+            )
+        if self.star_normalized_calibration_flux.do_measure_ap_corr:
+            raise ValueError(
+                "star_normalized_calibration_flux task must be configured with do_measure_ap_corr=False.",
```
parejkoj, line 382: I think these should be `ValidationError`, so that they give context?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -670,8 +711,15 @@ def _measure_aperture_correction(self, exposure, bright_sources):
             necessary for point source determination for the aperture correction
             calculation.
         """
-        result = self.measure_aperture_correction.run(exposure, bright_sources)
-        exposure.setApCorrMap(result.apCorrMap)
+        norm_ap_corr_map = exposure.info.getApCorrMap()
```
parejkoj, line 714: This is directly available on the exposure as a property:
```suggestion
        norm_ap_corr_map = exposure.apCorrMap
```

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -670,8 +711,15 @@ def _measure_aperture_correction(self, exposure, bright_sources):
...
+        ap_corr_map = self.measure_aperture_correction.run(exposure, bright_sources).apCorrMap
+
+        # Need to merge the aperture correction map from the normalization.
+        if norm_ap_corr_map:
+            for key in norm_ap_corr_map:
+                ap_corr_map[key] = norm_ap_corr_map[key]
+
+        exposure.setApCorrMap(ap_corr_map)
```
parejkoj, line 722: And you should be able to assign the property, too (I realize I didn't do that in the original, but I think I added it after I wrote this method):

```suggestion
        exposure.apCorrMap = ap_corr_map 
```

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -710,6 +758,13 @@ def _find_stars(self, exposure, background, id_generator):
 
         # Measure everything, and use those results to select only stars.
         self.star_measurement.run(sources, exposure)
+        # Run the normalization calibration flux task to apply the
+        # normalization correction to create normalized
+        # calibration fluxes.
+        self.star_normalized_calibration_flux.run(
+            exposure=exposure,
+            catalog=sources,
+        )
```
parejkoj, line 767: I'd just put this all on one line, but it's up to you; we've got 110 chars here.

parejkoj (commented):
Thanks for the added comments, that does help explain what's going on.

Please take a look at the docstring for `_measure_aperture_correction`: do you think it should be updated, or is that still correct?

Also, please see if it's worth adding a sentence to the task topic doc file. There's a numbered bullet about the aperture correction, and it's probably worth saying something there about this?

erykoff:
I've factored out the `CalibrateImageTask` changes to DM-44908 so as to not block the rest of this.

PR #946, opened 2024-06-12
DM-44802: Fix missing test coverage in CalibrateImage

parejkoj:
This fixes some real bugs that weren't being tested!

### DIFF
--- a/tests/test_calibrateImage.py
+++ b/tests/test_calibrateImage.py
@@ -216,6 +216,8 @@ class CalibrateImageTaskTests(lsst.utils.tests.TestCase):
             calibrate._handle_snaps([])
         with self.assertRaisesRegex(RuntimeError, "Can only process 1 or 2 snaps, not 3."):
             calibrate._handle_snaps(3*[self.exposure])
+        with self.assertRaisesRegex(RuntimeError, "must be either an afw Exposure"):
+            calibrate._handle_snaps("")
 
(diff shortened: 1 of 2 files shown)


tests/test_calibrateImage.py
```diff
@@ -216,6 +232,8 @@ def test_handle_snaps(self):
             calibrate._handle_snaps([])
         with self.assertRaisesRegex(RuntimeError, "Can only process 1 or 2 snaps, not 3."):
             calibrate._handle_snaps(3*[self.exposure])
+        with self.assertRaisesRegex(RuntimeError, "must be either an afw Exposure"):
```
erinleighh, line 235: I see an "either" without an "or" here. (Also, I know the guide doesn't say anything about error messages, but it'd be nice to have them as a complete sentence.)

erinleighh (approved):
Looks good! Only one tiny comment on one of the error messages. Thank you for adding more test coverage!

tests/test_calibrateImage.py
parejkoj: The error message itself (line 580 of the main code) is a complete sentence. This is testing that a regex (which in this case is just a partial text of the message) is included. There's not much point in fully writing out the message here, as checking the exact text of the message isn't as important as confirming that this particular RuntimeError was raised (as opposed to the other one, in the assert just above this).

PR #947, opened 2024-06-18
DM-44854: Add expTime and pixelScale to visitSummary and ccdVisitTable

laurenam:


jrmullaney (approved):

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -166,6 +166,7 @@ class ComputeExposureSummaryStatsTask(pipeBase.Task):
     - psfIxy
     - ra
     - dec
+    - pixelScale (arcsec/pixel)
```
jrmullaney, line 169: Should expTime also be included here? I know it's not really "calculated", but I feel that this list should say what's included in the summary, not just what's "calculated".

PR #948, opened 2024-06-20
DM-43831: Deprecate matched difference statistics

taranu:


python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -168,20 +172,25 @@ class DiffMatchedTractCatalogConfig(
     pipeBase.PipelineTaskConfig,
     pipelineConnections=DiffMatchedTractCatalogConnections,
 ):
+    _compute_stats: bool = False
+
```
leeskelvin, line 176: I read your comment on this on Jira, however, I think my preference would be to introduce an immediately deprecated boolean config field, which gets removed at the same time as the other deprecated assets. That's certainly the more standard approach, and I think the path of least potential confusion.

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -768,75 +814,74 @@ def run(
...
-            column_dist: (column_dummy, column_dist, column_dist_err, False),
-        }
-
-        # Cheat a little and do the first band last so that the color is
-        # based on the last band
+        data = None
         band_fluxes = [(band, config_flux) for (band, config_flux) in config.columns_flux.items()]
         n_bands = len(band_fluxes)
-        if n_bands > 0:
+        if self.config._compute_stats and (n_bands > 0):
```
leeskelvin, line 820: I imagine this whole `if` block needs an introductory `TODO` comment stating that it will be removed on DM-XXXXX as part of the deprecation process?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -333,23 +351,29 @@ def validate(self):
...
-    DIFF = MeasurementTypeInfo(
+    DIFF = SimpleNamespace(
         doc="difference (measured - reference)",
         name="diff",
     )
-    CHI = MeasurementTypeInfo(
+    CHI = SimpleNamespace(
         doc="scaled difference (measured - reference)/error",
         name="chi",
     )
```
leeskelvin, line 372: I don't object to the switch to `SimpleNamespace`, in fact, I rather prefer it. However, as this class is being deprecated and ultimately removed, I'm curious why the change?

leeskelvin (approved):

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: Leaving it as-is led to deprecation warnings on module import. I couldn't make a minimum reproducible example to debug the problem and so I just gave up and fixed it the easiest way I could.

PR #950, opened 2024-06-21
DM-44908: Add normalized calibration flux to CalibrateImageTask.

erykoff:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -320,11 +332,15 @@ def setDefaults(self):
...
                                          "base_ClassificationSizeExtendedness",
+                                         "base_CompensatedTophatFlux",
                                          ]
         self.star_measurement.slots.psfShape = "ext_shapeHSM_HsmPsfMoments"
         self.star_measurement.slots.shape = "ext_shapeHSM_HsmSourceMoments"
         # Only measure the apertures we need for star selection.
         self.star_measurement.plugins["base_CircularApertureFlux"].radii = [12.0]
+        self.star_measurement.plugins["base_CompensatedTophatFlux"].apertures = [12]
+
+        self.star_normalized_calibration_flux.do_measure_ap_corr = False
```
parejkoj, line 343: Please add a comment as to why you set this (I think I know, but it wouldn't be obvious to someone just looking at setDefaults).

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -352,6 +368,22 @@ def setDefaults(self):
         self.compute_summary_stats.starSelection = "calib_photometry_used"
         self.compute_summary_stats.starSelector.flags.good = ["calib_photometry_used"]
 
+    def validate(self):
+        super().validate()
+
+        # Ensure that the normalization calibration flux tasks
+        # are configured correctly.
+        if not self.psf_normalized_calibration_flux.do_measure_ap_corr:
+            msg = "psf_normalized_calibration_flux task must be configured with do_measure_ap_corr=True."
```
parejkoj, line 377: I think I want these two exception messages to have a `because ...` clause?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -521,9 +556,10 @@ def run(self, *, exposures, id_generator=None, result=None):
 
         result.psf_stars_footprints, result.background, candidates = self._compute_psf(result.exposure,
                                                                                        id_generator)
-        result.psf_stars = result.psf_stars_footprints.asAstropy()
+        self._normalize_calibration_flux(result.exposure, result.psf_stars_footprints)
+        self._measure_aperture_correction(result.exposure, result.psf_stars_footprints)
 
-        self._measure_aperture_correction(result.exposure, result.psf_stars)
+        result.psf_stars = result.psf_stars_footprints.asAstropy()
```
parejkoj, line 562: Huh... good catch! I am rather surprised that worked! I guess `self.measure_aperture_correction.run` only cares that the input duck types as a catalog in the most generic manner?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -654,13 +690,36 @@ def log_psf(msg):
...
         return detections.sources, background, psf_result.cellSet
 
+    def _normalize_calibration_flux(self, exposure, bright_sources):
+        """Normalize the calibration flux, using previously-measured
+        bright sources.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure to set the ApCorrMap on.
```
parejkoj, line 704: It is a bit confusing to have ApCorrMap mentioned here, even though that is what it is used for. Not sure what to say instead.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -670,8 +729,15 @@ def _measure_aperture_correction(self, exposure, bright_sources):
             necessary for point source determination for the aperture correction
             calculation.
         """
-        result = self.measure_aperture_correction.run(exposure, bright_sources)
-        exposure.setApCorrMap(result.apCorrMap)
+        norm_ap_corr_map = exposure.apCorrMap
+        ap_corr_map = self.measure_aperture_correction.run(exposure, bright_sources).apCorrMap
+
+        # Need to merge the aperture correction map from the normalization.
+        if norm_ap_corr_map:
```
parejkoj, line 736: I don't think this `if` is necessary, right? `exposure.apCorrMap` should always exist at the start of this method.

tests/test_calibrateImage.py
```diff
@@ -167,8 +168,10 @@ def _check_run(self, calibrate, result):
         self.assertTrue(np.isfinite(result.stars["coord_ra"]).all())
 
         # Returned photoCalib should be the applied value, not the ==1 one on the exposure.
+        # Note that this is very approximate because we are basing this comparison
+        # on just 2-3 stars.
```
parejkoj, line 172: Fair point! I didn't realize it was that unstable.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -654,13 +690,36 @@ def log_psf(msg):
         # Final measurement with the CRs removed.
         self.psf_source_measurement.run(detections.sources, exposure)
 
-        # PSF is set on exposure; only return candidates for optional saving.
+        # PSF is set on exposure; candidates are returned to use for
+        # calibration flux normalization and aperture corrections.
         return detections.sources, background, psf_result.cellSet
 
+    def _normalize_calibration_flux(self, exposure, bright_sources):
```
parejkoj, line 697: I'm not sure this needs to be a separate method from `_measure_aperture_correction`; I think I'd rename the latter to some combination and do both steps in there, since I don't see that we'd need either of them separately and they really do go together.

parejkoj (commented):
A few comments: thanks for catching my afw/pandas bug! I'm surprised that didn't cause trouble before.

python/lsst/pipe/tasks/calibrateImage.py
erykoff: You know, it probably doesn't.  It needed to be if `bright_sources` was an astropy table, but now that I have fixed that up this can be merged again.

python/lsst/pipe/tasks/calibrateImage.py
erykoff: I think so, but I'm also surprised that it worked and filled in the columns.  Certainly wouldn't be naively expected to work.

tests/test_calibrateImage.py
```diff
@@ -276,13 +278,18 @@ def test_measure_aperture_correction(self):
         """Test that _measure_aperture_correction() assigns an ApCorrMap to the
         exposure.
         """
+        # This test needs to be run without the normalized calibration
+        # which installs a smaller aperture correction map.
```
parejkoj, line 282: I don't think this comment is true any more?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -654,13 +695,18 @@ def log_psf(msg):
...
+        # calibration flux normalization and aperture corrections.
         return detections.sources, background, psf_result.cellSet
 
     def _measure_aperture_correction(self, exposure, bright_sources):
         """Measure and set the ApCorrMap on the Exposure, using
         previously-measured bright sources.
 
+        This task first normalizes the calibration flux
+        and then the full set of aperture corrections are measured
+        relative to this normalized calibration flux.
```
parejkoj, line 708: "This function" instead of "this task"?

Indentation is also odd.

parejkoj (approved):
One minor docstring comment, otherwise this looks good.

PR #952, opened 2024-06-28
DM-43593: calibrateImage improvements

parejkoj:


parejkoj:
Turns out I'd already added this exception functionality on DM-45295 (rebasing this was confusing until I realized that), but it had a bug in it, which I've fixed here and added a test to check it.

PR #956, opened 2024-07-09
DM-44889: Investigate feasibility of resurrecting matchBackgrounds.py background fitting algorithms

aemerywatkins:
The matchBackgrounds.py script was an initial attempt at matching multiple warped visit-level images to a given reference image. Taking the difference images between the reference and each subsequent image allows for each of these images to be modified to match the background of the reference image. These images may then be stacked (coadded), and a single background estimation made on the combined (and, importantly, deeper) coadd image.

The existing matchBackgroundsTask has not been utilized in some time. This ticket aims to explore the feasibility of resurrecting this code base, with a mind towards testing on HSC in the near-term.


python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,358 +19,473 @@
...
-                                       Add this to the science exposure to match the reference exposure.
-                - ``fitRMS``: The RMS of the fit. This is the sqrt(mean(residuals**2)).
-                - ``matchedMSE``: The MSE of the reference and matched images:
-                                  mean((refImage - matchedSciImage)**2);
-                    should be comparable to difference image's mean variance.
-                - ``diffImVar``: The mean variance of the difference image.
-                All fields except isReference will be None if isReference True or the fit failed.
+        result : `~lsst.afw.math.BackgroundList`
+            Differential background model
+            Add this to the science exposure to match the reference exposure.
```
TallJimbo: What "science exposure"?  This method takes `N` warps, so it seems like it needs to return `N-1` `BackgroundList` objects.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,358 +19,473 @@
...
-            refImageScaler = imageScalerList[refInd]
-
-        # refIndSet is the index of all exposures in expDataList that match the reference.
-        # It is used to avoid background-matching an exposure to itself. It is a list
-        # because it is possible (though unlikely) that expDataList will contain duplicates.
-        expKeyList = refExpDataRef.butlerSubset.butler.getKeys(expDatasetType)
-        refMatcher = DataRefMatcher(refExpDataRef.butlerSubset.butler, expDatasetType)
-        refIndSet = set(refMatcher.matchList(ref0=refExpDataRef, refList=expRefList))
+        if (numExp := len(warps)) < 1:
+            raise TaskError("No exposures to match")
```
TallJimbo: A production-ready task will need to work out some way to handle this case as well.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,358 +19,473 @@
...
+        # TODO: refactor below to construct blank bg model
+        im = refExposure.getMaskedImage()
+        blankIm = im.Factory(im, True)
+        blankIm.image.array *= 0
+
+        width = blankIm.getWidth()
+        height = blankIm.getHeight()
+        nx = width // self.config.binSize
+        if width % self.config.binSize != 0:
+            nx += 1
```
TallJimbo: This could be written as
```
nx = math.ceil(width / self.config.binSize)
```
(and same for `ny`, of course).

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,358 +19,473 @@
...
+        exposure.
 
-        Process creates a difference image of the reference exposure minus the science exposure, and then
-        generates an afw.math.Background object. It assumes (but does not require/check) that the mask plane
-        already has detections set. If detections have not been set/masked, sources will bias the
-        background estimation.
+        Process creates a difference image of the reference exposure minus the
+        science exposure, and then generates an afw.math.Background object. It
+        assumes (but does not require/check) that the mask plane already has
+        detections set. If detections have not been set/masked, sources will
```
TallJimbo: The whole point of background matching is that the astrophysical sources should subtract away, and I think that means we don't need to mask detections.  Even if the subtractions are't super clean, they should average out if we've done the right photometric scaling.

python/lsst/pipe/tasks/background.py
```diff
@@ -773,11 +825,310 @@ def getStatsImage(self):
...
+        skymap : `lsst.skymap.ringsSkyMap.RingsSkyMap`
+            Skymap object
+        tract : `int`
+            Placeholder Tract ID
+        transform : `lsst.afw.geom.TransformPoint2ToPoint2`
+            Transformation from tract coordinates to warp coordinates.
+        values : `lsst.afw.image.ImageF`
+            Measured background values.
+        numbers : `lsst.afw.image.ImageF`
+            Number of pixels in each background measurement.
```
TallJimbo: `__init__` parameter docs should go in the class docstring.  I don't think a docstring on an `__init__` actually goes anywhere useful.  Or at least it doesn't end up in all of the useful places.

python/lsst/pipe/tasks/background.py
```diff
@@ -773,11 +825,310 @@ def getStatsImage(self):
...
+
+        Parameters
+        ----------
+        warp : `lsst.afw.image.ExposureF`
+            Warped image corresponding to a single patch in a single visit
+        """
+        image = warp.getMaskedImage()
+        maskVal = image.getMask().getPlaneBitMask(self.config.mask)
+        # Photometric scaling necessary for contiguous background across tract
+        image.image.array *= warp.getPhotoCalib().instFluxToNanojansky(1)
```
TallJimbo: Prefer `PhotoCalib.calibrateImage(image)`.

That will also scale the variance plane, which this version neglects.

But actually all of our warps now have nJy units already, and I'm not positive that the `PhotoCalib` they have expects positions in tract coordinates, so it might be safest to just assume this has already been done.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
-        dtype=bool,
-        doc=("Use inverse-variance weighting when approximating background offset model? "
-             "This will fail when the background offset is constant "
-             "(this is usually only the case in testing with artificial images)."
-             "(usePolynomial=True)"),
+
+    approxWeighting = Field[bool](
+        doc="Use inverse-variance weighting when approximating the background offset model? This will fail "
+        "when the background offset is constant (usually only the case in testing with artificial images)."
+        "Only applied if ``usePolynomial=True``.",
```
TallJimbo: Another reason to just drop support for polynomials.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
-        min=0.
+        min=0.0,
     )
 
 
-class MatchBackgroundsTask(pipeBase.Task):
+class MatchBackgroundsTask(PipelineTask):
+    """Match the backgrounds of a list of warped exposures to a reference.
+
+    This task is a part of the background subtraction pipeline.
```
TallJimbo: Put this paragraph in a `Notes` section; the text right below the quick summary is supposed to just be a slightly longer summary (for reasons I don't understand, but that's how it is).

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
-                expDatasetType=expDatasetType,
-            )
-            refExpDataRef = expRefList[refInd]
-            refImageScaler = imageScalerList[refInd]
+        # TODO: matching currently done in fluence, not surface brightness,
+        # units.  Current solution for filling in non-overlapping regions
+        # between ref and other images is fitting a plane, but this fails as
+        # plane is typically buried beneath warping Jacobian pattern.  Unsure
+        # how to resolve this issue, but it's critical to resolve for BG-
+        # matching to be viable.
```
TallJimbo: I think the answer involves warping the inverse of the `backgroundToPhotometricRatio` image (see `CalibrateImageTask` connections), or when that is not available for an instrument, a detector-coordinate image of ones (which wouldn't be ones after it's warped).

But to use it you'd need to find a way to use a difference of two splines scaled by the two inverse ratios as your basis functions, and I don't see a way to do that without reimplementing the spline fitter.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
+        return Struct(backgroundInfoList=backgroundInfoList, matchedImageList=matchedImageList)
 
-        self.log.info("Matching %d Exposures", numExp)
+    @timeMethod
+    def _makeTractBackgrounds(self, warps, skyMap, refVisitId=None):
+        """If no reference visit ID is supplied, create full tract models of
+        the backgrounds of all visits.
+        If a reference visit ID is supplied, create full tract models of the
+        difference image backgrounds between all visits and the reference
+        visit.
```
TallJimbo: This returning a difference background in one case and a non-difference background in another case suggests that some refactoring is necessary.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
+                            raise ValueError("Image does not cover enough of ref image for order and binsize")
+                        elif self.config.undersampleStyle == "REDUCE_INTERP_ORDER":
+                            self.log.warning("Reducing order to %d", (minNumberGridPoints - 1))
+                            order = minNumberGridPoints - 1
+
+                    # TODO: we fit the full tract image, which has already been
+                    # binned, so we set binSize=1 when fitting.  But this
+                    # results in an all 0s variance image, meaning can't weight
+                    # by inverse variance when doing the fit.
+                    weightByInverseVariance = False
```
TallJimbo: Another reason to drop the polynomials.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
+        visits = []  # To ensure dictionary key order is correct
+        for vis in visitTractBackgrounds:
+            visits.append(vis)
+            # Fit a polynomial model to the full tract plane
+            tractBg = visitTractBackgrounds[vis].getStatsImage()
+            fitBg, _ = self._makeBackground(tractBg, binSize=1)
+            # TODO: as stated above, fitting a pre-binned image results in a
+            # null variance image.  But we want to add variance into the cost
+            # function.  How best to do that?  Below is a bad temporary
+            # solution, just assuming variance = mean
```
TallJimbo: I'd recommend computing an estimate of the variance earlier and attaching it to the tract backgrounds (i.e. as a scalar).  `variance = mean` is definitely not going to hold in nJy.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
-
-        Parameters
-        ----------
-        ref0 : `Unknown`
-            Data for ref 0.
-        ref1 : `Unknown`
-            Data for ref 1.
+        fluxZp = exposure.getPhotoCalib().instFluxToNanojansky(1)
+        exposure.image *= fluxZp
+        return fluxZp
```
TallJimbo: Since warps are now nJy, I don't think you need this method anymore, and deleting it is the easiest way to find the places that call it.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
+    def matchBackgrounds(self, warps, skyMap, refVisitId):
+        """Match science exposures' background level to that of reference
+        exposure.
+
+        Process creates binned images of the full focal plane (in tract
+        coordinates) for all visit IDs, subtracts each from a similarly
+        binned FFP reference image, then generates TractBackground
+        objects.  It assumes (but does not require/check) that the mask planes
+        already have detections set.  If detections have not been set/masked,
+        sources will bias the difference image background estimation.
```
TallJimbo: Shouldn't the sources subtract out, and hence not bias the difference image background estimation?

If it looks like they _are_ biasing the backgrounds, then that sort of throws the whole premise into doubt (unless it's something simple, like needing to use PSF-matched warps instead of direct warps).

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
+            maskIm = im.getMaskedImage()
+            # Matching must be done at common zeropoint
+            instFluxToNanojansky = self._fluxScale(im)
+            tractBg = diffTractBackgrounds[visId]
+            diffModel = tractBg.toWarpBackground(im)
+            bkgdIm = diffModel.getImage()
+            maskIm.image += bkgdIm
+            # Then convert everything back to counts
+            maskIm.image /= instFluxToNanojansky
+            bkgdIm /= instFluxToNanojansky
```
TallJimbo: More now-unnecessary pixel calibrations.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
-        }
+        doc="Type of statistic to estimate pixel value for the grid points.",
+        default="MEANCLIP",
+        allowed={"MEAN": "mean", "MEDIAN": "median", "MEANCLIP": "clipped mean"},
     )
-    undersampleStyle = pexConfig.ChoiceField(
-        doc="Behaviour if there are too few points in grid for requested interpolation style. "
-        "Note: INCREASE_NXNYSAMPLE only allowed for usePolynomial=True.",
+    # TODO: binning is done apart from fitting now, making INCREASE_NXNYSAMPLE
+    # option unusable here.  Unsure if this will cause problems.
```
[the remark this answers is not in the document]
aemerywatkins: Removing the polynomial fitting solves this, too, I think, since only polynomial fitting uses that option.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,683 +19,627 @@
...
+        binned FFP reference image, then generates TractBackground
+        objects.  It assumes (but does not require/check) that the mask planes
+        already have detections set.  If detections have not been set/masked,
+        sources will bias the difference image background estimation.
+
+        The TractBackground objects representing the difference image
+        backgrounds are then used to generate 'offset' images for each warp
+        comprising the full science exposure visit, which are then added to
+        each warp to match the background to that of the reference visit at the
+        warp's location within the tract.
```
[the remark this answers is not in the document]
TallJimbo: Ah, I hadn't realized that you were relying on some sort of extrapolation of the reference background.  I can see why that would suggest polynomials rather than splines, but I think even polynomials won't do that well enough for that to work in practice.  They might, however, be a decent way to blend multiple reference images together.

python/lsst/pipe/tasks/matchBackgrounds.py
```diff
@@ -19,358 +19,473 @@
...
     )
-    gridStatistic = pexConfig.ChoiceField(
+    bestRefWeightEdgeCoverage = RangeField(
+        dtype=float,
+        doc="Edge coverage weight (number of valid edge pixels) when calculating the best reference "
+        "exposure. Higher weights prefer exposures with high coverage. Ignored when a ref visit supplied.",
+        default=0.2,
+        min=0.0,
+        max=1.0,
+    )
```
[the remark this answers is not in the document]
aemerywatkins: Done, but should this live in a separate module or just at the top of this one?

python/lsst/pipe/tasks/matchBackgrounds.py
aemerywatkins: Could just log a warning and return a struct with two empty lists?

python/lsst/pipe/tasks/matchBackgrounds.py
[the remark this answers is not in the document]
TallJimbo: Either is fine.

PR #959, opened 2024-07-12
DM-45234: Some minor test cleanups

timj:
* Disable changes to global log levels (changing test order can break tests)
* Do not use getPackageDir when there is no need to.
* Minor edit to test_calibrate.py

### DIFF
--- a/tests/test_isPrimaryFlag.py
+++ b/tests/test_isPrimaryFlag.py
@@ -22,7 +22,6 @@
 import os
 import unittest
 import numpy as np
-import logging
 
 from lsst.geom import Point2I, Box2I, Extent2I
 from lsst.skymap import TractInfo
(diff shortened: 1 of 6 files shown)


tests/test_isPrimaryFlag.py
```diff
@@ -165,9 +165,6 @@ def setUp(self):
         charImTask = CharacterizeImageTask(config=charImConfig)
         self.charImResults = charImTask.run(self.exposure)
 
-        # set log level so that warnings do not display
-        logging.getLogger("lsst.calibrate").setLevel(logging.ERROR)
-
```
erykoff, line 170: Does this make some warnings appear?

erykoff (approved):

tests/test_isPrimaryFlag.py
timj: Depends what you mean. They don't appear by default anywhere. Pytest captures warnings and reports them if the tests failed.

PR #961, opened 2024-07-17
DM-45295: Raise an AlgorithmError when there are no psf_stars/stars cross matches

parejkoj:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -41,6 +43,23 @@
...
 
 
+class NoPsfStarsToStarsMatchError(pipeBase.AlgorithmError):
+    """Raised when there are no matches between the psf_stars and stars
+    catalogs.
+    """
+    def __init__(self, *, n_psf_stars, n_stars):
+        msg = (f"No psf stars out of {n_psf_stars} matched {n_stars} calib stars."
+               " Downstream processes probably won't have useful stars in this case."
+               " Is `star_source_selector` too strict or is this a bad image?")
```
sr525, line 53: Is there anyway to tell these two options apart with the information we have here?

parejkoj: Not obviously that I know of: it would require some analysis of the psf sources and psf shape. Probably worth doing, but that's a bigger task than this ticket.

PR #964, opened 2024-07-29
DM-45426: Implement changes to ensure parity between old and new warps

arunkannawadi:


doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.make_direct_warp.MakeDirectWarpTask.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Config Guidelines
+*****************
+
+The user must specify the size of the model PSF to
+which to match by setting `config.modelPsf.defaultFwhm` in units of pixels.
+The appropriate values depends on science case. In general, for a set of
+input images, this config should equal the FWHM of the visit with the worst
+seeing. The smallest it should be set to is the median FWHM. The defaults
+of the other config options offer a reasonable starting point.
```
TallJimbo, line 45: Please rewrap as one sentence per line, and do a sweep for other paragraphs in this file.

doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.make_direct_warp.MakeDirectWarpTask.rst
```diff
@@ -0,0 +1,140 @@
...
+
+The following list presents the most common problems that arise from a
+misconfigured `~ip.diffim.modelPsfMatch.ModelPsfMatchTask`
+and corresponding solutions. All assume the default Alard-Lupton kernel,
+with configs accessed via
+`config.warpAndPsfMatch.psfMatch.kernel['AL']`. Each item in the list
+is formatted as:
+Problem: Explanation. *Solution*
+
+Troublshooting PSF-Matching Configuration
```
TallJimbo, line 55:
```suggestion
Troubleshooting PSF-Matching Configuration
```

doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.make_direct_warp.MakeDirectWarpTask.rst
```diff
@@ -0,0 +1,140 @@
...
+
+In Depth
+========
+
+Config Guidelines
+*****************
+
+The user must specify the size of the model PSF to
+which to match by setting `config.modelPsf.defaultFwhm` in units of pixels.
+The appropriate values depends on science case. In general, for a set of
```
TallJimbo, line 42:
```suggestion
The appropriate values depend on science case. In general, for a set of
```

doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.make_direct_warp.MakeDirectWarpTask.rst
```diff
@@ -0,0 +1,140 @@
...
+input images, this config should equal the FWHM of the visit with the worst
+seeing. The smallest it should be set to is the median FWHM. The defaults
+of the other config options offer a reasonable starting point.
+
+The following list presents the most common problems that arise from a
+misconfigured `~ip.diffim.modelPsfMatch.ModelPsfMatchTask`
```
TallJimbo, line 48:
```suggestion
misconfigured `~lsst.ip.diffim.modelPsfMatch.ModelPsfMatchTask`
```

doc/lsst.pipe.tasks/tasks/lsst.pipe.tasks.make_direct_warp.MakeDirectWarpTask.rst
```diff
@@ -0,0 +1,140 @@
...
+with configs accessed via
+`config.warpAndPsfMatch.psfMatch.kernel['AL']`. Each item in the list
+is formatted as:
+Problem: Explanation. *Solution*
+
+Troublshooting PSF-Matching Configuration
+*****************************************
+
+Matched PSFs look boxy
+**********************
```
TallJimbo, line 59: Seems like this should be different level of header, to nest below the "Troubleshooting ..." heading.

I also suspect it would be better to nest the `Solution` headers under their corresponding problem headers, rather than making them siblings.  That's especially true if those headings ever appear in any kind of table of contents, where just "Solution" with no context doesn't make any sense.

python/lsst/pipe/tasks/make_direct_warp.py
```diff
@@ -316,6 +336,36 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+            bboxList, wcsList = [], []
+            for dataId in dataIdList:
+                row = visit_summary.find(dataId["detector"])
+                if row is None:
+                    raise RuntimeError(f"Unexpectedly incomplete visit_summary: {dataId=} is missing.")
+                bboxList.append(row.getBBox())
+                wcsList.append(row.getWcs())
+
+            # inputs["bboxList"] = bboxList
+            # inputs["wcsList"] = wcsList
```
TallJimbo, line 351: Should these be uncommented now?  Or deleted?

python/lsst/pipe/tasks/make_direct_warp.py
```diff
@@ -316,6 +336,36 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
 
         visit_summary = inputs["visit_summary"] if self.config.useVisitSummaryPsf else None
 
+        # Can this go in the run method?
```
TallJimbo, line 339: Should be doable, given that the `run` method takes `DeferredDatasetHandles`, and you can get data IDs from those.

python/lsst/pipe/tasks/make_direct_warp.py
```diff
@@ -409,30 +459,48 @@ def run(self, inputs, sky_info, visit_summary):
             )
             warpedExposure.setPsf(psfWarped)
 
+            if final_warp.photoCalib is not None and warpedExposure.photoCalib is not None:
+                ratio = (
+                    final_warp.photoCalib.getInstFluxAtZeroMagnitude()
+                    / warpedExposure.photoCalib.getInstFluxAtZeroMagnitude()
+                )
+            else:
+                ratio = 1
```
TallJimbo, line 468: Do we always expect the first logic branch here to fire in practice?  I'm wondering if we can somehow make it an error rather than a silent change in behavior if a `PhotoCalib` goes incorrectly missing upstream.

python/lsst/pipe/tasks/make_direct_warp.py
```diff
@@ -547,6 +620,28 @@ def process(
...
 
+    @staticmethod
+    def _filterInputs(indices, inputs):
+        """Filter task inputs by their indices.
+
+        Parameters
+        ----------
+        indices : `list` [`int`]
+        inputs : `dict` [`list`]
+            A dictionary of input connections to be passed to run.
```
TallJimbo, line 631: It's not at all clear what these collections actually contain.

python/lsst/pipe/tasks/make_psf_matched_warp.py
```diff
@@ -60,6 +62,12 @@ class MakePsfMatchedWarpConnections(
     },
 ):
     """Connections for MakePsfMatchedWarpTask"""
+    sky_map = Input(
```
TallJimbo, line 65: @erykoff, want to weigh in on what to call this connection?  I hate the proliferation of "sky" "map" variants, but I think you're less resigned to it than I am...

TallJimbo (approved):

python/lsst/pipe/tasks/make_direct_warp.py
arunkannawadi: I primarily intended this to run without issues if one were to give simulated images without a `photoCalib`. Would it be better to have a trivial `photoCalib` that is the same in all exposures instead?

arunkannawadi: On real data, for the very first `calexp`, the else branch would fire since `final_warp` has no `photoCalib` attached to it. The first `calexp`'s `photoCalib` gets attached to the `final_warp` and others are scaled accordingly. In the past, the ordering was not consistent resulting in apparent differences. Lauren had patched it up by ordering the refs by detector ID, but I take that butler gives the inputs in the same order now.

TallJimbo: Yes, I think expecting simulated data to have a trivial `PhotoCalib` is reasonable.

So I think we just need to raise if the `calexp` `PhotoCalib` is absent, while allowing the warp's to be absent (for the first iteration).

python/lsst/pipe/tasks/make_direct_warp.py
arunkannawadi: Agree. I lifted these from `makeWarp` instead of importing from there, given the anticipated deprecation. I've improved the wording a bit but not too worried given that it's a private method as opposed to this being a public method of `MakeWarpTask`.

python/lsst/pipe/tasks/make_direct_warp.py
arunkannawadi: Deleted them. I'm not a big fan of shoving more items in the `inputs` dictionary when we can pass them explicitly.

python/lsst/pipe/tasks/make_direct_warp.py
arunkannawadi: Alright, branching only based on the warp's `photoCalib`. I don't want to impose that this exception is made only for the first iteration. If there are no good pixels from the first visit, then the `photoCalib` comes from the second visit.

python/lsst/pipe/tasks/make_psf_matched_warp.py
arunkannawadi: Once @erykoff weighs in on `^[sS]ky[_]?[mM]ap$` this PR/ticket is good to go.

arunkannawadi: Thanks, Eli! This is also consistent with other usage of `sky_map` connection, mostly introduced by Eli elsewhere. For e.g., https://github.com/lsst/pipe_tasks/blob/e7c99ddb59f0c91ad7ed5a058e0712b776b49d51/python/lsst/pipe/tasks/healSparseMapping.py#L353

PR #967, opened 2024-08-14
DM-42176: Move PeekExposureTask from summit_utils

mfisherlevine:


arunkannawadi (approved):

tests/test_peekExposureTask.py
```diff
@@ -0,0 +1,104 @@
+#
+# LSST Data Management System
+# Copyright 2008-2017 AURA/LSST.
```
arunkannawadi, line 3: This is the old preamble.

python/lsst/pipe/tasks/peekExposure.py
```diff
@@ -0,0 +1,1276 @@
+# This file is part of summit_utils.
```
arunkannawadi, line 1: Rename to pipe_tasks

tests/test_peekExposureTask.py
mfisherlevine: Huh, it's probably like that in loads of these files then (I just copied and pasted from the quickFrameMeasurement tests as a starting point).

arunkannawadi: It is outdated there as well.

PR #969, opened 2024-08-16
DM-45773: Make initial isolated star source grouping work on sub-pixels.

erykoff:
The initial isolated star grouping is now broken into healpixels to save memory.  A larger nside_split will use less memory, but with more overhead/time.  This is why the default is relatively small.

Grouping is done on a healpixel plus a small boundary region that is 10x the match radius.  After grouping observations the list of associated objects is trimmed to the inner region of the healpixel.  In this way we avoid boundary effects.


jchiang87 (approved):
lgtm

PR #971, opened 2024-08-17
DM-45709: Explicitly evaluate pixelScale at the detector center for single frame processing

laurenam:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1254,10 +1254,9 @@ def run(self, visitSummaryRefs):
                 for detector_id in summaryTable["id"]
             ]
             ccdEntry["detector"] = summaryTable["id"]
-            pixToArcseconds = np.array([vR.getWcs().getPixelScale().asArcseconds() if vR.getWcs()
-                                        else np.nan for vR in visitSummary])
-            ccdEntry["seeing"] = visitSummary["psfSigma"] * np.sqrt(8 * np.log(2)) * pixToArcseconds
-
+            ccdEntry["seeing"] = (
+                visitSummary["psfSigma"] * visitSummary["pixelScale"] * np.sqrt(8 * np.log(2))
+            )
```
leeskelvin, line 1259: The removal of the `else` here I guess leans into the `visitSummary` table *always* being the source of truth, right? I.e., it will have a value assigned for `pixelScale` for each detector, or `NaN`, and we shouldn't need to add extra logic here to catch that?

leeskelvin (approved):

python/lsst/pipe/tasks/postprocess.py
laurenam: That is indeed my thinking.

BrunoSanchez (approved):

PR #972, opened 2024-08-22
DM-45577: Add option to drop PSF fit table connection

taranu:
If using shapelet PSF parameters, these are in the deepCoadd_meas catalog and so there's no need to try to load a multiprofit PSF fit table that might not even exist.


leeskelvin (approved):
Nothing contentious here; approved, and good to merge pending review of the other PRs on this ticket.

PR #973, opened 2024-08-23
DM-45664:  Fix incorrect definition for mosaics

czwa:


erykoff (approved):
This indeed looks like the only such mistake in this file.

PR #977, opened 2024-09-10
DM-45042: ComputeExposureSummaryStats needs to be able to turn off "update" code

laurenam:


python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -418,14 +468,18 @@ def update_psf_stats(
...
-            maxDistToNearestPsf = maximum_nearest_psf_distance(
-                image_mask,
-                psf_cat,
-                sampling=self.config.psfSampling,
-                bad_mask_bits=self.config.psfBadMaskPlanes
-            )
-            summary.maxDistToNearestPsf = float(maxDistToNearestPsf)
+        if self.config.psfSampling is None:
+            self.log.info("Note: not computing grid-based maxDistToNearestPsf model fidelity. Setting "
+                          "it to NaN.")
```
parejkoj, line 473: No need to say "setting it to NaN", that's the default value of all non-computed things.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -331,33 +376,38 @@ def update_psf_stats(
...
-                        summary.psfSigma,
-                        sampling=self.config.psfGridSampling,
-                        bad_mask_bits=self.config.psfBadMaskPlanes,
-                    )
-                summary.psfApCorrSigmaScaledDelta = float(psfApCorrSigmaScaledDelta)
+        if self.config.psfGridSampling is None:
+            self.log.info("Note: not computing grid-based PSF & ApCorr model fidelity metrics. Setting "
+                          "psfTraceRadiusDelta, psfApFluxDelta, & psfApCorrSigmaScaledDelta to NaN.")
+        else:
+            if image_mask is not None:
```
parejkoj, line 383: Why is image_mask optional? I can't think of any case where we'd not have access to the mask, and certainly the three places that we call this method, we pass a mask.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -418,14 +468,18 @@ def update_psf_stats(
...
-        if image_mask is not None:
-            maxDistToNearestPsf = maximum_nearest_psf_distance(
-                image_mask,
-                psf_cat,
-                sampling=self.config.psfSampling,
-                bad_mask_bits=self.config.psfBadMaskPlanes
-            )
-            summary.maxDistToNearestPsf = float(maxDistToNearestPsf)
+        if self.config.psfSampling is None:
+            self.log.info("Note: not computing grid-based maxDistToNearestPsf model fidelity. Setting "
```
parejkoj, line 472: I'm not sure I understand what "model fidelity" means here in context?

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -331,33 +376,38 @@ def update_psf_stats(
...
-                        image_mask,
-                        image_ap_corr_field,
-                        summary.psfSigma,
-                        sampling=self.config.psfGridSampling,
-                        bad_mask_bits=self.config.psfBadMaskPlanes,
-                    )
-                summary.psfApCorrSigmaScaledDelta = float(psfApCorrSigmaScaledDelta)
+        if self.config.psfGridSampling is None:
+            self.log.info("Note: not computing grid-based PSF & ApCorr model fidelity metrics. Setting "
+                          "psfTraceRadiusDelta, psfApFluxDelta, & psfApCorrSigmaScaledDelta to NaN.")
```
parejkoj, line 381: No need to say you're setting them to NaN, that's the default for uncomputed values.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -331,33 +376,38 @@ def update_psf_stats(
...
-                    psfApCorrSigmaScaledDelta = compute_ap_corr_sigma_scaled_delta(
-                        image_mask,
-                        image_ap_corr_field,
-                        summary.psfSigma,
-                        sampling=self.config.psfGridSampling,
-                        bad_mask_bits=self.config.psfBadMaskPlanes,
-                    )
-                summary.psfApCorrSigmaScaledDelta = float(psfApCorrSigmaScaledDelta)
+        if self.config.psfGridSampling is None:
+            self.log.info("Note: not computing grid-based PSF & ApCorr model fidelity metrics. Setting "
```
parejkoj, line 380: Not sure what "model fidelity" means here?

parejkoj (changes requested):
I think you need to strip this down to only have "doX" configs for the slowest components. The rest of the basic values we should always try to calculate. There's no point in adding "doX" settings for things that we'll always want to compute.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -40,6 +40,41 @@
...
 class ComputeExposureSummaryStatsConfig(pexConfig.Config):
     """Config for ComputeExposureSummaryTask"""
+    doUpdatePsfStats = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Do update the psf statistics. Note that the PSF and apCorr model fidelity "
+        "delta metrics are tied up in the same function as the more basic PSF metrics (e.g. "
+        "psfSigma and size/shape residuals). If the basic ones are still desired, but the "
+        "computation-intesive delta metrics are not, leave this as true and set "
+        "config.psfGridSampling to None. Set to False if speed is of the essence.",
```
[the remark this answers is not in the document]
laurenam: Agreed.  I removed that as an option and added a note in the comments which parameters are always (i.e. non-optionally) run when this task is called.  I also changed the grid-based ones to `doUpdateX`-style config options (which I do think is more obvious than setting the sampling sizes to `None`.  The `ap_pipe` commit was updated to reflect this.)

python/lsst/pipe/tasks/computeExposureSummaryStats.py
parejkoj: Please ask about that on slack and file a ticket if there's no known reason: as I said, all three places in the stack that currently call this pass `mask`, and it doesn't make sense for it to be optional to me.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -40,6 +40,50 @@
...
+    doUpdateApCorrModelStats = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Update the grid-based apCorr model fidelity statistic (psfApCorrSigmaScaledDelta)? "
+        "Set to False if speed is of the essence.",
+    )
+    doUpdateMaxDistToNearestPsfStats = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Update the grid-based maximun distance to the nearest PSF star fidelity statistic "
```
parejkoj, line 58:
```suggestion
        doc="Update the grid-based maximum distance to the nearest PSF star fidelity statistic "
```

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -40,6 +40,50 @@
...
+        dtype=bool,
+        default=True,
+        doc="Update the masked image (i.e. skyNoise & meanVar) statistics? Set to False "
+        "if speed is of the essence.",
+    )
+    doUpdateEffectiveTimeStats = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Update the effective time statistics? Set to False if speed is of the essence.",
+    )
```
parejkoj, line 86: I think you should remove these: the statistics computed on a grid should not be a performance hit, so there's no need to disable them.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
parejkoj: Why not be explicit about it? "PSF & ApCorr model minimum/maximum range metrics"? From the docstrings on those functions, that seems like what they are, no?

tests/test_computeExposureSummaryStats.py
```diff
@@ -101,12 +101,51 @@ def testComputeExposureSummary(self):
...
+        expSummaryTaskNoUpdates.config.doUpdatePhotoCalibStats = False
+        expSummaryTaskNoUpdates.config.doUpdateBackgroundStats = False
+        expSummaryTaskNoUpdates.config.doUpdateMaskedImageStats = False
+        expSummaryTaskNoUpdates.config.doUpdateEffectiveTimeStats = False
+
+        summary = expSummaryTaskNoUpdates.run(exposure, None, background)
+        # Test the outputs
+        nan = float("nan")
+        self.assertTrue(summary.ra, nan)
+        self.assertTrue(summary.dec, nan)
```
parejkoj, line 125: `assertTrue`? I don't think this is doing what you think it is:

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue

Did you mean `self.assertTrue(np.isnan(summary.ra))`, and similarly for all the tests below? Note that you cannot test equality for NaN, because it's not equal to itself.

parejkoj (commented):
I definitely

python/lsst/pipe/tasks/computeExposureSummaryStats.py
laurenam: Huh, just want to double check that you really don't want control over any but the grid-based ones.  When we discussed this (way back when!), I was left with the strong feeling that you wanted control over almost all of them for AP (e.g. in the ticket description you mention `update_wcs_stats` and `update_masked_image_deltas` taking >1s per image as being a problem).  Does even if AP decides to keep many of them at present, does it hurt to have the option already available in case opinions on that change (for AP or other caller of this task)?

If you do want these removed, just for the record, can you specify on the ticket that you no longer consider any but the grid-based computations an issue?

python/lsst/pipe/tasks/computeExposureSummaryStats.py
laurenam: I'm gonna push back on this again.  It's pre-existing and unrelated to this ticket.  I am not bothered by it, so would not feel inclined to weigh in on, let alone initiate, a discussion about it.  If you feel strongly about it, proceed as you will!

tests/test_computeExposureSummaryStats.py
laurenam: Thanks for catching that!

python/lsst/pipe/tasks/computeExposureSummaryStats.py
parejkoj: I've waffled back and forth on this a bit. The reason to not have `doX` configs for things that there would be no real reason to turn off is that it makes the code and config listings more complicated for no real benefit. I'll go add a comment on the ticket about this.

I've been doing some profiling but wanted to run on some OR4 data (which is more realistic than DC2) to get a better sense of the slow points.

PR #979, opened 2024-09-17
DM-43077: Convert all tasks to use CalibrateImageTask outputs

parejkoj:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -188,30 +188,33 @@ def run(self, catalogs, tract, patch):
...
 
     catalog = connectionTypes.Input(
         doc="Input full-depth catalog of sources produced by CalibrateTask",
-        name="{catalogType}src",
+        name="{catalogType}initial_stars_footprints_detector",
         storageClass="SourceCatalog",
         dimensions=("instrument", "visit", "detector")
     )
     outputCatalog = connectionTypes.Output(
         doc="Catalog of sources, `src` in DataFrame/Parquet format. The 'id' column is "
```
leeskelvin, line 203: Legacy usage of `src` in this doc string.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -188,30 +188,33 @@ def run(self, catalogs, tract, patch):
         return catalog
 
 
+# TODO: should deprecate this?
```
leeskelvin, line 191: Are you planning on performing these deprecations on this ticket?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -281,6 +284,7 @@ class WriteRecalibratedSourceTableConfig(WriteSourceTableConfig,
     )
 
 
+# TODO: deprecate, since reprocessVisitImage does this more thoroughly?
```
leeskelvin, line 287: If you're asking for an opinion, I can see the merit of adding deprecation flags here. If `reprocessVisitImage` can do a better job, I don't see the need to keep this around. With that said, is this beyond the scope of this ticket? 

If deprecations are set up, please also set up the future removal ticket, blocked by the appropriate version.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1122,6 +1126,7 @@ def _combineExposureMetadata(self, visit, dataRefs):
 class ConsolidateSourceTableConnections(pipeBase.PipelineTaskConnections,
                                         defaultTemplates={"catalogType": ""},
                                         dimensions=("instrument", "visit")):
+    # TODO: Deprecate the dataframe connection?
```
leeskelvin, line 1129: And replace it with another storage class? Or just a deprecation and removal?

schemas/ForcedSource.yaml
```diff
@@ -95,8 +95,9 @@ calexpFlags:
     - base_PixelFlags_flag_saturatedCenter
     - base_PixelFlags_flag_crCenter
     - base_PixelFlags_flag_suspectCenter
-    - base_PixelFlags_flag_streak
-    - base_PixelFlags_flag_streakCenter
+    # Streak flags not yet propagated from difference imaging.
```
leeskelvin, line 98: If these are being left in but commented, please add a `TODO` and a DM ticket number for their eventual reinstatement.

schemas/Object.yaml
```diff
@@ -769,8 +769,9 @@ flags:
     - base_PixelFlags_flag_sensor_edgeCenter
     - base_PixelFlags_flag_suspect
     - base_PixelFlags_flag_suspectCenter
-    - base_PixelFlags_flag_streak
-    - base_PixelFlags_flag_streakCenter
+    # Streak flags not yet propagated from difference imaging.
```
leeskelvin, line 772: `TODO` and ticket link here too please.

schemas/Source.yaml
```diff
@@ -380,8 +381,9 @@ flags:
    - base_PixelFlags_flag_saturatedCenter
    - base_PixelFlags_flag_suspect
    - base_PixelFlags_flag_suspectCenter
-   - base_PixelFlags_flag_streak
-   - base_PixelFlags_flag_streakCenter
+   # Streak flags not yet propagated from difference imaging.
```
leeskelvin, line 384: `TODO` and ticket link here please.

schemas/Source.yaml
```diff
@@ -4,7 +4,8 @@
 # See the DPDD for more information about the output: https://lse-163.lsst.io
 funcs:
     sourceId:
-        functor: Index
+        functor: Column
+        args: id
```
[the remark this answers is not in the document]
parejkoj: See the comment on this commit: https://github.com/lsst/pipe_tasks/pull/979/commits/1df49118bb78f876ffcfa569a13a9ee7f211b5b5

The `Index` line didn't work with ArrowAstropy when I tried it, and this treats it just like a normal column.

I'm surprised this file works with `calibrateImage` outputs, since that task is not configured to produce an output catalog with many of the fields that this `Source.yaml` expects. That's why there's a new dedicated `.yaml` file for that task's outputs on another commit.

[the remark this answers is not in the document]
TallJimbo: > The `Index` line didn't work with ArrowAstropy when I tried it, and this treats it just like a normal column.

Do you remember what didn't work?  I also expected it not to work, but unless I didn't test what I thought I did, restoring the original didn't cause a failure in `transform[Pre]SourceTable` or `consolidate[Pre]SourceTable` for me.  I'm running a ci_hsc in Jenkins now with this one bit reverted (on DM-46991) so that'll hopefully check what I tried to do locally.

> I'm surprised this file works with calibrateImage outputs, since that task is not configured to produce an output catalog with many of the fields that this Source.yaml expects. 

In my incremental approach I've reconfigured the task to produce all of those fields in DRP (except `calib_detected`).  We've got analysis tools that need some of them, but the pipeline steps we have now are really not amenable to adding the multiple invocations we need of those tools (for initial stars, recalibrated stars, and final sources), so I want to fix that on DM-47320 before trying to review which analysis tools we run when, and I want to do that before trimming down the `calibrateImage` output fields back to [close to] the defaults.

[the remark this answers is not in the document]
parejkoj: > Do you remember what didn't work?

No, I don't remember any of the details at this point. I'm pretty sure one of those listed tasks just failed when I tried to use it with ArrowAstropy inputs. It was probably ci_imsim or ci_hsc where the failure occurred.

>  I've reconfigured the task to produce all of those fields in DRP

Oof, that's what I was afraid of. Good luck. I hope we really can get it back to the defaults, and/or any non-default values are ones that we can look over with AP, too.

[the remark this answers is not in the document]
TallJimbo: Ok, I figured it out - without this change, `transformSourceTable` does misbehave, but only by replacing the real source ID column with `range(0, N)`, and it turns out nothing actually raises an exception (even downstream _anywhere_ in the pipeline). A lot of tasks just produce no outputs, because `finalizeCharacterization` doesn't produce any PSFs.

TallJimbo:
Closing this PR, since all work has moved to other tickets.

PR #981, opened 2024-09-25
DM-44612: Deprecate old warping tasks

arunkannawadi:


python/lsst/pipe/tasks/warpAndPsfMatch.py
```diff
@@ -43,6 +46,8 @@ class WarpAndPsfMatchConfig(pexConfig.Config):
     )
 
 
+@deprecated(reason="The WarpAndPsfMatchTask is no longer in use. Will be removed after v28.",
+            version="v28.0", category=FutureWarning)
```
cmsaunders, line 50: Maybe just add here the name of the task that is replacing this.

cmsaunders (approved):

python/lsst/pipe/tasks/warpAndPsfMatch.py
arunkannawadi: This has no replacement (un)fortunately. I could point to `MakeDirectWarpTask` and `MakePsfMatchedWarpTask` instead.

PR #982, opened 2024-09-27
DM-40855: Add deprecation warning to fakes related tasks

BrunoSanchez:


python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -142,6 +149,11 @@ def _add_fake_sources(exposure, objects, calibFluxRadius=12.0, logger=None):
             exposure[subBox].mask.array |= bitmask
 
 
+@deprecated(
```
isullivan, line 152: Probably better to not add a warning to this function, since it is called once for every fake source in the catalog. Plus, it was a private function that is only used here.

python/lsst/pipe/tasks/matchFakes.py
```diff
@@ -37,7 +37,14 @@
 
 from lsst.pipe.tasks.insertFakes import InsertFakesConfig
 
+from deprecated.sphinx import deprecated
 
+
+@deprecated(
```
isullivan, line 43: No need to add a deprecation warning to the Connections class

python/lsst/pipe/tasks/matchFakes.py
```diff
@@ -330,6 +342,11 @@ def _getVectors(self, ras, decs):
         return vectors
 
 
+@deprecated(
```
isullivan, line 345: No need to add a deprecation warning to the Connections class

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -41,7 +41,14 @@
 from lsst.skymap import BaseSkyMap
 from lsst.pipe.tasks.calibrate import CalibrateTask
 
+from deprecated.sphinx import deprecated
 
+
+@deprecated(
```
isullivan, line 47: No need to add a deprecation warning to the Connections class

python/lsst/pipe/tasks/processCcdWithFakes.py
```diff
@@ -620,6 +637,11 @@ def copyCalibrationFields(self, calibCat, sourceCat, fieldsToCopy):
         return newCat
 
 
+@deprecated(
```
isullivan, line 640: No need to add a deprecation warning to the Connections class

python/lsst/pipe/tasks/insertFakes.py
```diff
@@ -40,7 +40,14 @@
 from lsst.pex.exceptions import LogicError, InvalidParameterError
 from lsst.geom import SpherePoint, radians, Box2D, Point2D
 
+from deprecated.sphinx import deprecated
 
+
+@deprecated(
```
isullivan, line 46: No need to deprecate a private function.

isullivan (approved):
Looks good. I don't think the private functions or the Connections classes need deprecation warnings.

PR #983, opened 2024-09-27
DM-42979: Use calibFlux for source selector

abudlong:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -371,6 +371,8 @@ def setDefaults(self):
         self.star_selector["science"].doUnresolved = True
         self.star_selector["science"].doSignalToNoise = True
         self.star_selector["science"].signalToNoise.minimum = 10.0
+        self.star_selector["science"].signalToNoise.fluxField = "slot_CalibFlux_instFlux"
+        self.star_selector["science"].signalToNoise.errField = "slot_CalibFlux_instFluxErr"
```
parejkoj, line 375: What was the default for these? I would have thought that `CalibFlux` was the default, but ScienceSourceSelector defaults to `PsfFlux`. That surprises me, and I wonder if we have other places that have been inadvertently using the wrong flux?

Although, you shouldn't have a negative flux if we're restricting to S/N>10, since N is always positive, no?

isullivan: These defaulted to PsfFlux, and none of the later source selectors apply a N/N cut. They all assume that has been taken care of by this first selector, so if a source with valid PsfFlux but negative CalibFlux makes it through this selection it gets used for photometry.

isullivan (approved):
We verified these changes fix the warning and underlying problem, and do not change the photometric calibration for LSSTComCam. More details on Jira.

erykoff (approved):

PR #987, opened 2024-10-07
DM-46699: Dummy output of GetRegionTimeFromVisitTask confuses provenance tools

kfindeisen:
This PR fixes a bug where `GetRegionTimeFromVisitTask` would cause analysis of quantum graphs to crash.


eigerx (approved):
lgtm!

PR #989, opened 2024-10-09
DM-46693: Add ability to re-apply bgModel1 in SkyCorrectionTask

aemerywatkins:
The current `SkyCorrectionTask` has three principal steps:

1. `bgModel1` - a large-scale sky estimation and subtraction

2.  `sky` - a subtraction of the sky frame

3.  `bgModel2` - a small-scale sky estimation and subtraction

If one has `doSky = True`, then `bgModel1` is implied. However, during testing of [DM-44889](https://rubinobs.atlassian.net/browse/DM-44889), it became apparent that the need for a `doSky = True` -only dataset output was required. This necessitates that `bgModel1` is generated, used, and then undone.

This ticket adds that functionality into the current `SkyCorrectionTask`. 

Note the prior related removal ticket of `doBgModel1`, [DM-37429](https://rubinobs.atlassian.net/browse/DM-37429), which may be relevant/of use here.

[DM-44889]: https://rubinobs.atlassian.net/browse/DM-44889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DM-37429]: https://rubinobs.atlassian.net/browse/DM-37429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/lsst/pipe/tasks/skyCorrection.py
+++ b/python/lsst/pipe/tasks/skyCorrection.py
@@ -300,6 +300,8 @@ class SkyCorrectionTask(PipelineTask):
         Returns
         -------
         results : `Struct` containing:
+            skyFrameScale : `float`
+                Scale factor applied to the sky frame.
             skyCorr : `list` [`lsst.afw.math.BackgroundList`]
                 Detector-level sky correction background lists.
             calExpMosaic : `lsst.afw.image.ExposureF`
@@ -318,8 +320,9 @@ class SkyCorrectionTask(PipelineTask):
         initialBackgroundIndex = len(calBkgs[0]._backgrounds) - 1
 
         # Subtract a scaled sky frame from all input exposures
+        skyFrameScale = None
         if self.config.doSky:
-            self._subtractSkyFrame(calExps, skyFrames, calBkgs)
+            skyFrameScale = self._subtractSkyFrame(calExps, skyFrames, calBkgs)
 
         # Adds full-fp bg back onto exposures, removes it from list
         if self.config.undoBgModel1:
@@ -341,7 +344,9 @@ class SkyCorrectionTask(PipelineTask):
             skyCorrExtras, camera, self.config.binning, ids=calExpIds, refExps=calExps
         )
 
-        return Struct(skyCorr=calBkgs, calExpMosaic=calExpMosaic, calBkgMosaic=calBkgMosaic)
+        return Struct(
+            skyFrameScale=skyFrameScale, skyCorr=calBkgs, calExpMosaic=calExpMosaic, calBkgMosaic=calBkgMosaic
+        )
 
     def _restoreOriginalBackgroundRefineMask(self, calExps, calBkgs):
         """Restore original background to each calexp and invert the related
@@ -562,6 +567,11 @@ class SkyCorrectionTask(PipelineTask):
             Sky frame calibration data for the input detectors.
         calBkgs : `list` [`lsst.afw.math.BackgroundList`]
             Background lists associated with the input calibrated exposures.
+
+        Returns
+        -------
+        scale : `float`
+            Scale factor applied to the sky frame.
         """
         skyFrameBgModels = []
         scales = []
@@ -577,6 +587,7 @@ class SkyCorrectionTask(PipelineTask):
             # also updating the calBkg list in-place
             self.sky.subtractSkyFrame(calExp.getMaskedImage(), skyFrameBgModel, scale, calBkg)
         self.log.info("Sky frame subtracted with a scale factor of %.5f", scale)
+        return scale
 
     def _binAndMosaic(self, exposures, camera, binning, ids=None, refExps=None):
         """Bin input exposures and mosaic across the entire focal plane.
--- a/python/lsst/pipe/tasks/skyCorrection.py
+++ b/python/lsst/pipe/tasks/skyCorrection.py
@@ -178,6 +178,11 @@ class SkyCorrectionConfig(PipelineTaskConfig, pipelineConnections=SkyCorrectionC
         dtype=FocalPlaneBackgroundConfig,
         doc="Initial background model, prior to sky frame subtraction",
     )
+    undoBgModel1 = Field(
+        dtype=bool,
+        default=False,
+        doc="If True, adds back initial background model after sky and removes bgModel1 from the list",
+    )
     sky = ConfigurableField(
         target=SkyMeasurementTask,
         doc="Sky measurement",
@@ -209,6 +214,11 @@ class SkyCorrectionConfig(PipelineTaskConfig, pipelineConnections=SkyCorrectionC
         self.bgModel2.xSize = 256
         self.bgModel2.ySize = 256
         self.bgModel2.smoothScale = 1.0
+
+    def validate(self):
+        super().validate()
+        if self.undoBgModel1 and not self.doSky and not self.doBgModel2:
+            raise ValueError("If undoBgModel1 is True, task requires at least one of doSky or doBgModel2.")
 
 
 class SkyCorrectionTask(PipelineTask):
@@ -278,11 +288,11 @@ class SkyCorrectionTask(PipelineTask):
 
         Parameters
         ----------
-        calExps : `list` [`lsst.afw.image.exposure.ExposureF`]
+        calExps : `list` [`lsst.afw.image.ExposureF`]
             Detector calibrated exposure images for the visit.
         calBkgs : `list` [`lsst.afw.math.BackgroundList`]
             Detector background lists matching the calibrated exposures.
-        skyFrames : `list` [`lsst.afw.image.exposure.ExposureF`]
+        skyFrames : `list` [`lsst.afw.image.ExposureF`]
             Sky frame calibration data for the input detectors.
         camera : `lsst.afw.cameraGeom.Camera`
             Camera matching the input data to process.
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -178,6 +177,11 @@ class SkyCorrectionConfig(PipelineTaskConfig, pipelineConnections=SkyCorrectionC
         dtype=FocalPlaneBackgroundConfig,
         doc="Initial background model, prior to sky frame subtraction",
     )
+    doBgModel1 = Field(
+        dtype=bool,
+        default=True,
+        doc="If False, adds back initial background model after sky",
+    )
```
fred3m, line 184: I would change the name of this config parameter. I see the symmetry in making it similar to doBgModel2 but in that case the config option chooses whether or not to do the final cleanup and background subtraction. Here, it looks like the initial background subtraction is done in order to do the sky subtraction, then it is added back in if `doBgModel1 == False`. So I would recommend calling this `restoreOriginalBackground` or something similar that doesn't make it sound as if the model 1 background subtraction is skipped completely.

fred3m (commented):
Just a few minor comments, otherwise LGTM.

python/lsst/pipe/tasks/skyCorrection.py
aemerywatkins: How about `restoreBgModel1`?  To keep the symmetry but clarify the usage.  And then the default value will swap to `False`.

fred3m: Maybe I misunderstand the code, but it isn't restoring bgModel1, right? I thought that it's restoring the background before bgModel1 was made and subtracted. Maybe `undoBgModel1` or `removeBgModel1` or `revertBgModel1`?

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -387,6 +400,37 @@ def _restoreBackgroundRefineMask(self, calExps, calBkgs):
             skyCorrBases.append(skyCorrBase)
         return calExps, skyCorrBases
 
+    def _restoreVisitBackground(self, calExps, calBkgs):
+        """Restores the full focal-plane background to a visit.
+        Runs after _subtractSkyFrame() if doBgModel1=False.
+
+        Parameters
+        ----------
+        calExps : `list` [`lsst.afw.image.exposure.ExposureF`]
```
leeskelvin, line 409: This can just be:
```
calExps : `list` [`lsst.afw.image.ExposureF`]
```

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -387,6 +400,37 @@ def _restoreBackgroundRefineMask(self, calExps, calBkgs):
...
+            Calibrated exposures to be background subtracted.
+        calBkgs : `list` [`lsst.afw.math._backgroundList.BackgroundList`]
```
leeskelvin, line 411: This can also just be:
```
calBkgs : `list` [`lsst.afw.math.BackgroundList`]
```

python/lsst/pipe/tasks/skyCorrection.py
leeskelvin: (and in the returns below)

python/lsst/pipe/tasks/skyCorrection.py
leeskelvin: (and in the returns below)

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -387,6 +400,37 @@ def _restoreBackgroundRefineMask(self, calExps, calBkgs):
...
+            Background lists associated with the input calibrated exposures.
+
+        Returns
+        -------
+        calExps : `list` [`lsst.afw.image.maskedImage.MaskedImageF`]
+            Background subtracted exposures for creating a focal plane image.
+        calBkgs : `list` [`lsst.afw.math._backgroundList.BackgroundList`]
+            Updated background lists with a visit-level model removed.
```
leeskelvin, line 419: There's a return section here, but this function doesn't seem to return anything? Instead, it seems to edit in-place?

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -387,6 +400,37 @@ def _restoreBackgroundRefineMask(self, calExps, calBkgs):
...
+        for calExp, calBkg in zip(calExps, calBkgs):
+            image = calExp.getMaskedImage()
+
+            # Restore full focal-plane background in calexp; remove from BGList
+            skyCorrBase = calBkg[-2][0].getImageF()
+            image += skyCorrBase
+            calBkg._backgrounds.pop(-2)
+
+            self.log.info(
+                "Detector %d: FFP background restored",
```
leeskelvin, line 430: I think this log message could do with being more verbose, e.g.:
```
"Detector %d: Initial background model prior to sky frame subtraction (bgModel1) has been restored",
```

This does however now clash with usage of "initial background" on lines 377 and 396, which refer to the "original background" at the calexp level. I would favor renaming these two instances to "original background", and preserve initial background as a reference to `bgModel1`.

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -28,14 +28,13 @@
...
-    SkyMeasurementTask,
-)
-from lsst.pipe.tasks.visualizeVisit import VisualizeMosaicExpConfig, VisualizeMosaicExpTask
+from lsst.pipe.base import (PipelineTask, PipelineTaskConfig,
+                            PipelineTaskConnections, Struct)
+from lsst.pipe.tasks.background import (FocalPlaneBackground,
+                                        FocalPlaneBackgroundConfig,
+                                        MaskObjectsTask, SkyMeasurementTask)
+from lsst.pipe.tasks.visualizeVisit import (VisualizeMosaicExpConfig,
+                                            VisualizeMosaicExpTask)
```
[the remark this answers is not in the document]
leeskelvin: I agree, this formatting change should probably be reverted. When I run this file through my `black`/`isort` formatters, it also reverts it to its original format as well.

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -387,6 +400,37 @@ def _restoreBackgroundRefineMask(self, calExps, calBkgs):
             skyCorrBases.append(skyCorrBase)
         return calExps, skyCorrBases
 
+    def _restoreVisitBackground(self, calExps, calBkgs):
```
leeskelvin, line 403: I would suggest renaming this to `_restoreInitialBackground`.

In tandem with this, the current method `_restoreBackgroundRefineMask` should probably also be renamed to `_restoreOriginalBackgroundRefineMask`, to avoid any potential confusion.

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -387,6 +400,37 @@ def _restoreBackgroundRefineMask(self, calExps, calBkgs):
...
+        Parameters
+        ----------
+        calExps : `list` [`lsst.afw.image.exposure.ExposureF`]
+            Calibrated exposures to be background subtracted.
+        calBkgs : `list` [`lsst.afw.math._backgroundList.BackgroundList`]
+            Background lists associated with the input calibrated exposures.
+
+        Returns
+        -------
+        calExps : `list` [`lsst.afw.image.maskedImage.MaskedImageF`]
```
leeskelvin, line 416: This worried me originally, as the input `calExps` are `ExposureF`, but the output is a `MaskedImageF`? However, I think this returns section is going away in any case (see other related comment). You can probably replace this section with a `Notes` stating that the inputs are modified in-place.

python/lsst/pipe/tasks/skyCorrection.py
leeskelvin: The order of operations facilitated here is:

1. subtract bgModel1
2. subtract sky
3. remove bgModel1 from the background list
4. subtract bgModel2

At the end of this, the `skyCorr` output background list should consist of:

1. the inverted original detector-level background elements (4 of them)
2. no bgModel1 (this has been used to fit the sky frame, and then removed from the bg list)
3. sky frame
4. bgModel2

With that in mind, I do agree that naming this config `doBgModel1` makes it seem like a similar operation to `doBgModel2`, but it does act differently. I wasn't sure which alternative to endorse; I realize that in describing the order of operations and dataset outputs above I twice referred to this as a removal. However, out of context, "remove BG Model 1" sounds like you're subtracting the model, which is the opposite of what we're doing here.

So with that said, I think `undoBgModel1` is a nice compromise. I think we would need to make the usage of this config clear in the doc string though.

python/lsst/pipe/tasks/skyCorrection.py
leeskelvin: PS - if we're changing the associated config to `undoBgModel1`, it probably also makes sense to rename this method `_undoInitialBackground` as well.

fred3m (approved):
Looks like a couple of linting changes required to pass the tests. Other than that this LGTM.

PR #993, opened 2024-10-14
DM-46627: Move bands_fit from MPF config to CoaddMultibandFitSubConfig

taranu:
This really should have been in the base class all along instead of being specific to MultiProFit - multiband fitters are going to have to specify which bands they fit one way or another and picking this option seems fine.


fred3m (approved):
LGTM

PR #994, opened 2024-10-15
DM-46684: Add Metrics to calibrateImage task metadata

jrmullaney:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -912,3 +935,21 @@ def _summarize(self, exposure, stars, background):
         # applied calibration). This needs to be checked.
         summary = self.compute_summary_stats.run(exposure, stars, background)
         exposure.info.setSummaryStats(summary)
+
+    def _recordFlaggedPixelFractions(self, exposure):
```
ctslater, line 939: For consistency I'd always call pixels Masked rather than Flagged. (though admittedly "mask" is a bad holdover from the era of only having a single "mask" plane, and that keeps confusing new users)

python/lsst/pipe/tasks/repair.py
```diff
@@ -227,3 +227,4 @@ def cosmicRay(self, exposure, keepCRs=None):
 
         text = "kept" if keepCRs else "interpolated over"
         self.log.info("Identified and %s %s cosmic rays.", text, num)
+        self.metadata['cosmic_ray_count'] = num
```
isullivan, line 230: Super minor point, but you used double quotes everywhere else.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -636,16 +641,20 @@ def _compute_psf(self, exposure, id_generator):
         cell_set : `lsst.afw.math.SpatialCellSet`
             PSF candidates returned by the psf determiner.
         """
-        def log_psf(msg):
+        def log_psf(msg, metadataPrefix=None):
             """Log the parameters of the psf and background, with a prepended
-            message.
+            message. If metadataPrefix is not None, then the psf sigma and
+            media background is written to the task metadata, prefixed by the
+            the passed string.
```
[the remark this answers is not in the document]
isullivan: Coming back to this, I believe you only use this function to add to the metadata in one place. So, perhaps you could use a flag such as `addMetadata` that defaults to False, and if it is set write to a predefined metadata entry.

isullivan (approved):
Looks good, some suggestions for you to consider.

PR #995, opened 2024-10-17
DM-46933: remove references to STREAK mask and prep tasks for CalibrateImage

TallJimbo:


python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -613,6 +688,17 @@ def compute_psf_and_ap_corr_map(self, visit, detector, exposure, src, isolated_s
         measured_src : `lsst.afw.table.SourceCatalog`
             Updated source catalog with measurements, flags and aperture corrections.
         """
+        if self.config.remove_initial_photo_calib:
+            if initial_photo_calib is None:
+                self.log.warning("No initial photo calib found for visit %d, detector %d", visit, detector)
+                return None, None, None
+            if not initial_photo_calib._isConstant:
+                # TODO DM-46720: remove this limitation and usage of private (why?!) property.
```
leeskelvin, line 696: Do we want to leave `why?!` on `main`? That seems like a discussion to be had outside of the scope of this ticket.

python/lsst/pipe/tasks/makeWarp.py
```diff
@@ -107,6 +116,33 @@ def __init__(self, *, config=None):
...
+                "is not an error), but it may mean that 'config.removeInitialPhotoCalib' should be False.",
+                label,
+                data_id,
+            )
+            raise pipeBase.NoWorkFound("No initial photo calibs.")
+        elif not self.config.removeInitialPhotoCalib and inputs["initialPhotoCalibList"]:
+            log.warning(
+                "For %s quantum %s, input collections have initial photo calib datasets but "
+                "'config.removeInitialPhotoCalib=False'.  This is either a very unusual collection "
+                "search path or (more likely) a bad configuration.  Not that this config option should "
```
leeskelvin, line 139: Possible "Not" typo again.

python/lsst/pipe/tasks/makeWarp.py
```diff
@@ -526,6 +578,26 @@ def _prepareCalibratedExposures(self, *, visitSummary, calExpList=[], wcsList=No
...
+            # Remove initial photometric calibration, if we expect one.
+            if self.config.removeInitialPhotoCalib:
+                if initialPhotoCalib is None:
+                    self.log.warning(
+                        "Detector id %d for visit %d has no initialPhotoCalib and will "
+                        "not be used in the warp", detectorId, row["visit"],
+                    )
+                    continue
+                if not initialPhotoCalib._isConstant:
+                    # TODO DM-46720: remove this limitation and usage of private (why?!) property.
```
leeskelvin, line 595: "why?!" again

schemas/initial_stars_detector_standardized.yaml
```diff
@@ -0,0 +1,196 @@
...
+        functor: Column
+        args: slot_Centroid_yErr
+    # x_y_Cov: not available
+    ra:
+        functor: RAColumn
+    dec:
+        functor: DecColumn
+
+   # RFC-924: Temporarily keep a duplicate "decl" entry for backwards
+   # compatibility. To be removed after September 2023.
```
leeskelvin, line 42: Should the post-September 2023 removal be a part of this ticket?

leeskelvin (approved):

schemas/initial_stars_detector_standardized.yaml
TallJimbo: It definitely shouldn't be part of this ticket; I think it might be disruptive (hence the RFC).  And while September 2023 has long passed, I think that was a "no earlier than" date, not a deadline.

TallJimbo:
I've moved the commits here that received most of the comments over to DM-46993 (no, that's not a typo of this ticket), where I expect to ultimately drop them rather than merge them.  I'll address the comments here there if the plan changes.

PR #996, opened 2024-10-19
DM-46966: Make calibrating pixel values optional in CalibrateImageTask

TallJimbo:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -292,6 +297,16 @@ class CalibrateImageConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Cali
         doc="Task to to compute summary statistics on the calibrated exposure."
     )
 
+    do_calibrate_pixels = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc=(
+            "If True, apply the photometric calibration to the image pixels "
+            "and background model, and attach an identify PhotoCalib to "
```
erykoff, line 305: Identity not identify

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -880,45 +899,62 @@ def _fit_astrometry(self, exposure, stars):
...
+
+    def _apply_photometry(self, exposure, background):
+        """Apply the photometric model attached to the exposure to the
+        exposure's pixels and an associated background model.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure with the target `lsst.afw.image.PhotoCalib` attached.
+            On return, pixel values will be calibrated and an identify
```
erykoff, line 940: Identity

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -880,45 +899,62 @@ def _fit_astrometry(self, exposure, stars):
...
         exposure : `lsst.afw.image.Exposure`
             Exposure that is being fit, to get PSF and other metadata from.
-            Modified to be in nanojanksy units, with an assigned photoCalib
-            identically 1.
+            Has the fit `lsst.afw.image.PhotoCalib` attached, with pixel values
+            unchanged.
         stars : `lsst.afw.table.SourceCatalog`
             Good stars selected for use in calibration.
+        background : `lsst.afw.math.BackgroundList`
+            Background model to convert to nanojansky units in place.
```
parejkoj, line 915: You got rid of this parameter, no?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -880,45 +899,62 @@ def _fit_astrometry(self, exposure, stars):
...
 
         Returns
         -------
         calibrated_stars : `lsst.afw.table.SourceCatalog`
             Star catalog with flux/magnitude columns computed from the fitted
-            photoCalib.
+            photoCalib (instFlux columns are retained as well).
         matches : `list` [`lsst.afw.table.ReferenceMatch`]
             Reference/stars matches used in the fit.
```
parejkoj, line 923: It looks like we need to also add `matchMeta` to this output list, too. Sorry I'd neglected that.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -880,45 +899,62 @@ def _fit_astrometry(self, exposure, stars):
...
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure with the target `lsst.afw.image.PhotoCalib` attached.
+            On return, pixel values will be calibrated and an identity
+            photometric transform will be attached.
+        background : `lsst.afw.math.BackgroundList`
+            Background model to convert to nanojansky units in place.
+        """
+        photo_calib = exposure.getPhotoCalib()
```
parejkoj, line 945: Why not use `exposure.photoCalib` below, instead of assigning it? Doesn't really matter, I just don't like proliferating the `getX` when we don't have to.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -928,17 +964,15 @@ def _summarize(self, exposure, stars, background):
...
             Good stars selected used in calibration.
         background : `lsst.afw.math.BackgroundList`
             Background that was fit to the exposure during detection of the
-            above stars.
+            above stars.  Should be in instrumental units.
         """
-        # TODO investigation: because this takes the photoCalib from the
-        # exposure, photometric summary values may be "incorrect" (i.e. they
-        # will reflect the ==1 nJy calibration on the exposure, not the
-        # applied calibration). This needs to be checked.
```
parejkoj, line 941: Well, at least I'd left a TODO to reflect this that can be retired!

tests/test_calibrateImage.py
```diff
@@ -169,11 +171,17 @@ def _check_run(self, calibrate, result):
...
-        self.assertFloatsAlmostEqual(result.applied_photo_calib.getCalibrationMean(),
-                                     self.photo_calib, rtol=1e-2)
+        if expect_calibrated_pixels:
+            # Fit photoCalib should be the applied value if we calibrated
+            # pixels, not the ==1 one on the exposure.
+            photo_calib = result.applied_photo_calib
+            self.assertEqual(result.exposure.getPhotoCalib().getCalibrationMean(), 1.0)
+        else:
+            self.assertIsNone(result.applied_photo_calib)
+            photo_calib = result.exposure.getPhotoCalib()
```
parejkoj, line 181:
```suggestion
            photo_calib = result.exposure.photoCalib
```

tests/test_calibrateImage.py
```diff
@@ -169,11 +171,17 @@ def _check_run(self, calibrate, result):
...
-        # Returned photoCalib should be the applied value, not the ==1 one on the exposure.
-        # Note that this is very approximate because we are basing this comparison
-        # on just 2-3 stars.
-        self.assertFloatsAlmostEqual(result.applied_photo_calib.getCalibrationMean(),
-                                     self.photo_calib, rtol=1e-2)
+        if expect_calibrated_pixels:
+            # Fit photoCalib should be the applied value if we calibrated
+            # pixels, not the ==1 one on the exposure.
+            photo_calib = result.applied_photo_calib
+            self.assertEqual(result.exposure.getPhotoCalib().getCalibrationMean(), 1.0)
```
parejkoj, line 178:
```suggestion
            self.assertEqual(result.exposure.photoCalib.getCalibrationMean(), 1.0)
```

tests/test_calibrateImage.py
```diff
@@ -228,6 +236,18 @@ def test_run_no_optionals(self):
         self.assertNotIn("astrometry_matches", result.getDict())
         self.assertNotIn("photometry_matches", result.getDict())
 
+    def test_run_no_calibrate_pixels(self):
+        """Test that run() returns reasonable values to be butler put when
+        do_calibrate_pixels=False.
```
parejkoj, line 241: I guess this came from my original wording, but this is much clearer.

```suggestion
        """Test that run() returns reasonable values when
        ``do_calibrate_pixels=False``.
```

tests/test_calibrateImage.py
```diff
@@ -595,6 +616,36 @@ def test_runQuantum_no_optional_outputs(self):
             # Restore the one we removed for the next test.
             connections[optional] = temp
 
+    def test_runQuantum_no_calibrate_pixels(self):
```
parejkoj, line 619: Should note in the test docstring that this is testing that the `applied_photo_calib` output has been removed.

parejkoj (commented):
Should setting this new config to False also require changing the exposure output name to something other than `initial_pvi`, so that we can keep track of which ones were calibrated and which not? That's my main worry here, since right now the only way to know is to check whether the attached PhotoCalib is `==1`.

TallJimbo:
> Should setting this new config to `False` also require changing the exposure output name to something other than `initial_pvi`, so that we can keep track of which ones were calibrated and which not? That's my main worry here, since right now the only way to know is to check whether the attached `PhotoCalib` is `==1`.

I'm a little worried about this, but I think putting constraints on the dataset type name might be the kind of fix that's worse than the problem, especially with the names still somewhat in flux.

To get at the root issue, what do you think about setting `BUNIT` in the FITS header?  That's not as nice as some afw object that maps to and from the FITS header, but it's a very established external standard, and arguably all we really need for external users.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -949,6 +949,7 @@ def _apply_photometry(self, exposure, background):
                                        photo_calib.getCalibrationErr(),
                                        bbox=exposure.getBBox())
         exposure.setPhotoCalib(identity)
+        exposure.metadata["BUNIT"] = "nJy"
```
parejkoj, line 952: Although I'd much rather have this done within PhotoCalib itself, I guess it's not the worst thing to do it here, even better if we also do it in `reprocessVisitImage`, and the diffim and coadd code.

python/lsst/pipe/tasks/calibrateImage.py
parejkoj: Oh, right, good point.

parejkoj:
You don't have to do it on this ticket, but please file a ticket about setting BUNIT everywhere we're writing an nJy image, and/or getting PhotoCalib to do it automatically if possible.

parejkoj (approved):

TallJimbo:
As you might have seen, I've already added `BUNIT` to `CalibrateImageTask` here, and I've created [DM-47408](https://rubinobs.atlassian.net/browse/DM-47408) for the rest.

PR #1000, opened 2024-11-05
DM-46852: Return None for the PSF if the computed shape is NAN

isullivan:


parejkoj (commented):
I think this is the most important fix on this ticket, and possibly the only one that really matters.

Although, what happens in diffim detection for the None PSF case here? Thinking about it, since this is DRP-specific, the `pvi` won't be written, and so diffim won't be run, so I guess it's fine?

isullivan:
This is used in diffim is `subtractImages._applyExternalCalibrations`, where it looks for an updated PSF from the `finalVisitSummary`, and applies it if it is not `None`. So if it is `None`, then it just keeps the original PSF attached to the exposure.

parejkoj (approved):

PR #1001, opened 2024-11-06
DM-47318 - Change getRegionTimeFromVisit input to calibrateImage output

Gerenjie:
Change getRegionTimeFromVisit input dummy_visit to a calibrateImage output so that we can run it before mpSkyEphemerisQueryTask


python/lsst/pipe/tasks/getRegionTimeFromVisit.py
```diff
@@ -33,13 +33,10 @@ class GetRegionTimeFromVisitConnections(pipeBase.PipelineTaskConnections,
...
 
     dummy_visit = pipeBase.connectionTypes.Input(
-        doc="Placeholder connection to provide visit-detector records and "
-            "constrain data IDs to images we're processing. Use of a catalog "
-            "dataset lets us also control how late in the pipeline the task "
-            "is run.",
-        name="{fakesType}{coaddName}Diff_diaSrcTable",
-        storageClass="DataFrame",
-        dimensions={"instrument", "visit", "detector"},
+        doc="Exposure with region and time, I hope!",
```
kfindeisen, line 36: I don't think this doc is an improvement over the original, not least because it's _not_ an exposure, it's a catalog.

python/lsst/pipe/tasks/getRegionTimeFromVisit.py
```diff
@@ -33,13 +33,10 @@ class GetRegionTimeFromVisitConnections(pipeBase.PipelineTaskConnections,
                                         defaultTemplates={"coaddName": "goodSeeing", "fakesType": ""}):
```
kfindeisen, line 33: Possibly add a comment that these templates are kept in case `dummy_visit` is a dataset type that needs them?

tests/test_getRegionTimeFromVisit.py
```diff
@@ -23,15 +23,14 @@
...
-import lsst.utils.tests
-from lsst.sphgeom import ConvexPolygon, UnitVector3d
+from lsst.afw.table import SourceCatalog
 import lsst.daf.butler
 import lsst.daf.butler.tests as butlerTests
 import lsst.pipe.base.testUtils as pipeTests
-
 from lsst.pipe.tasks.getRegionTimeFromVisit import GetRegionTimeFromVisitTask
+from lsst.sphgeom import ConvexPolygon, UnitVector3d
+import lsst.utils.tests
```
kfindeisen, line 33: In the future, please [put this kind of reorganization on its own commit](https://developer.lsst.io/work/flow.html#git-commit-organization-best-practices).

kfindeisen (approved):
Generally looks good, but please revert the change to `dummy_visit.doc` -- the original text still applies to the new catalog and provides more of an explanation to any developer who might happen across this code.

PR #1004, opened 2024-11-08
DM-47307: Clarify nVisitsMin docstring in good seeing selector

mrawls:


python/lsst/pipe/tasks/selectImages.py
```diff
@@ -647,8 +647,13 @@ class BestSeeingQuantileSelectVisitsConfig(pipeBase.PipelineTaskConfig,
         max=1,
     )
     nVisitsMin = pexConfig.Field(
-        doc="At least this number of visits selected and supercedes quantile. For example, if 10 visits "
-            "cover this patch, qMin=0.33, and nVisitsMin=5, the best 5 visits will be selected.",
+        doc="The ideal minimum number of visits to select. In regimes with many visits, "
```
parejkoj, line 650: I think I'd use another wording than "ideal" here. I'm having trouble coming up with something better that I like, though. Maybe something like this?

```suggestion
        doc="The minimum number of visits to select, if qMin and qMax were to select fewer. In regimes with many visits, "
```

parejkoj (approved):
That's very thorough, and definitely an improvement. See if you like some variation on my modification, but thanks for clearing this up.

PR #1010, opened 2024-11-21
DM-34875: switch most DataFrame connections to ArrowAstropy

TallJimbo:


python/lsst/pipe/tasks/isolatedStarAssociation.py
```diff
@@ -319,14 +321,13 @@ def _make_all_star_sources(self, tract_info, source_table_ref_dict):
...
-            df.reset_index(inplace=True)
+        for visit in source_table_handle_dict:
+            source_table_ref = source_table_handle_dict[visit]
+            tbl = source_table_ref.get(parameters={'columns': all_columns})
 
-            goodSrc = self.source_selector.selectSources(df)
+            goodSrc = self.source_selector.selectSources(tbl)
 
-            table = df[persist_columns][goodSrc.selected].to_records()
+            table = tbl[persist_columns][goodSrc.selected].as_array().view(np.recarray)
```
erykoff, line 330: I'm not sure why we need the `.view(np.recarray)` here?  Or why (equivalently) `np.asarray(tbl[persist_columns][goodSrc.selected])` doesn't work.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -35,17 +35,21 @@
            "TransformForcedSourceTableConfig", "TransformForcedSourceTableTask",
            "ConsolidateTractConfig", "ConsolidateTractTask"]
 
+import dataclasses
 import functools
-import pandas as pd
```
erykoff, line 39: 😄

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -35,17 +35,21 @@
...
 import logging
-import numpy as np
 import numbers
 import os
 
+import numpy as np
+import pandas as pd
```
erykoff, line 45: Oh... sigh.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -244,15 +247,16 @@ def run(self, catalog, visit, detector, **kwargs):
...
         result : `~lsst.pipe.base.Struct`
             ``table``
-                `DataFrame` version of the input catalog
+                `astropy.table.Table` version of the input catalog
         """
         self.log.info("Generating DataFrame from src catalog visit,detector=%i,%i", visit, detector)
-        df = catalog.asAstropy().to_pandas().set_index("id", drop=True)
-        df["visit"] = visit
+        tbl = catalog.asAstropy()
+        tbl.add_index("id")
```
erykoff, line 254: This makes me nervous.  Why do we need the index here?  We aren't persisting it... Also astropy indexes are wonky and I don't want to rely on them since the support seems nonexistent.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -767,10 +773,15 @@ class TransformObjectCatalogConnections(pipeBase.PipelineTaskConnections,
...
         dimensions=("tract", "patch", "skymap"),
-        storageClass="DataFrame",
+        storageClass="ArrowAstropy",
         name="objectTable"
     )
 
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        if config.multilevelOutput:
+            self.outputCatalog = dataclasses.replace(self.outputCatalog, storageClass="DataFrame")
```
erykoff, line 783: This is some next-level python stuff here.  So you can just do this to change one element of a dataclass?  And `self.outputCatalog.storageClass = "DataFrame"` doesn't work for some reason?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -797,7 +808,9 @@ class TransformObjectCatalogConfig(TransformCatalogBaseConfig,
         dtype=bool,
         default=False,
         doc=("Whether results dataframe should have a multilevel column index (True) or be flat "
-             "and name-munged (False).")
+             "and name-munged (False).  If True, the output storage class will be "
+             "set to DataFrame, since astropy tables do not support multi-level indexing."),
+        deprecated="Support for multi-level outputs is deprecated and will be removed after v29.",
```
erykoff, line 813: I thought stuff downstream expected this multilevel format?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -950,8 +966,8 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
         self.log.info("Concatenating %s per-patch Object Tables",
                       len(inputs["inputCatalogs"]))
-        df = pd.concat(inputs["inputCatalogs"])
-        butlerQC.put(pipeBase.Struct(outputCatalog=df), outputRefs)
+        table = astropy.table.vstack(inputs["inputCatalogs"], join_type="exact")
```
erykoff, line 969: This was a major memory hog on real RC2 data before.  I assume this is also a major memory hog; I want to confirm that it isn't *more* of a memory hog now?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1231,16 +1247,16 @@ def run(self, visitSummaryRefs):
...
             # 'visitId' is the key to the visitId table. They are the same.
             # Technically you should join to get the visit from the visit
             # table.
-            ccdEntry = ccdEntry.rename(columns={"visit": "visitId"})
+            ccdEntry.rename_column("visit", "visitId")
 
             # RFC-924: Temporarily keep a duplicate "decl" entry for backwards
             # compatibility. To be removed after September 2023.
-            ccdEntry["decl"] = ccdEntry.loc[:, "dec"]
+            ccdEntry["decl"] = ccdEntry["dec"]
```
erykoff, line 1259: Since it's now November 2024 can we just remove this?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1370,8 +1387,8 @@ def run(self, visitSummaries):
             # mirror1Temp, mirror2Temp, mirror3Temp, domeTemp, externalTemp,
             # dimmSeeing, pwvGPS, pwvMW, flags, nExposures.
 
-        outputCatalog = pd.DataFrame(data=visitEntries)
-        outputCatalog.set_index("visitId", inplace=True, verify_integrity=True)
+        outputCatalog = astropy.table.Table(rows=visitEntries)
+        outputCatalog.add_index("visitId")
```
erykoff, line 1391: Again, the astropy table index thing.

tests/test_isolatedStarAssociation.py
```diff
@@ -22,8 +22,9 @@
 """Test IsolatedStarAssociationTask.
 """
 import unittest
+
+import astropy.table
 import numpy as np
-import pandas as pd
```
erykoff, line 26: 🎉

tests/test_isolatedStarAssociation.py
```diff
@@ -186,9 +186,8 @@ def _make_simdata(self,
                     # Make one star have low s/n
                     table['normCompTophatFlux_instFlux'][0] = 1.0
 
-                df = pd.DataFrame(table)
-                df.set_index('sourceId', inplace=True)
-                data_refs.append(lsst.pipe.base.InMemoryDatasetHandle(df, storageClass="DataFrame"))
+                tbl = astropy.table.Table(table)
+                handles.append(lsst.pipe.base.InMemoryDatasetHandle(tbl, storageClass="DataFrame"))
```
erykoff, line 190: Should this be `DataFrame` or `ArrowAstropy`?

tests/test_isolatedStarAssociation.py
```diff
@@ -234,9 +233,9 @@ def test_match_primary_stars(self):
         """Test matching primary stars."""
         # Stack all the sources; we do not want any cutting here.
         tables = []
-        for data_ref in self.data_refs:
-            df = data_ref.get()
-            tables.append(df.to_records())
+        for handle in self.handles:
+            tbl = handle.get()
+            tables.append(tbl.as_array().view(np.recarray))
```
erykoff, line 238: Again `np.asarray(tbl)`?

tests/test_isolatedStarAssociation.py
```diff
@@ -271,9 +270,9 @@ def test_match_sources(self):
         """Test _match_sources source to primary matching."""
         # Stack all the sources; we do not want any cutting here.
         tables = []
-        for data_ref in self.data_refs:
-            df = data_ref.get()
-            tables.append(df.to_records())
+        for handle in self.handles:
+            tbl = handle.get()
+            tables.append(tbl.as_array().view(np.recarray))
```
erykoff, line 275: Also here.

erykoff (approved):
Overall looks good; a bunch of small comments.

tests/test_isolatedStarAssociation.py
TallJimbo: Fixed.  But I guess nothing actually cares, given what the test actually does with the handle.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -78,6 +83,96 @@ def flattenFilters(df, noDupCols=["coord_ra", "coord_dec"], camelCase=False, inp
     return newDf
 
 
+class TableVStack:
```
erykoff, line 86: This is great, and I want to make sure it's used as widely as possible.  Is this the best place for this code?

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -78,6 +83,96 @@ def flattenFilters(df, noDupCols=["coord_ra", "coord_dec"], camelCase=False, inp
...
+                column = table[name]
+                column_cls = type(column)
+                self.result[name] = column_cls.info.new_like([column], self.capacity, name=name)
+                self.result[name][:len(table)] = column
+            self.index = len(table)
+            self.result.meta = table.meta.copy()
+        else:
+            next_index = self.index + len(table)
+            for name in table.colnames:
+                self.result[name][self.index:next_index] = table[name]
```
erykoff, line 148: You promise "strict" in the docstring, but this doesn't actually check that the column names/types are the same (or extra columns).  Is this because it would be too inefficient and we just yolo?

erykoff (approved):
This all looks great!  Some small comments.

python/lsst/pipe/tasks/postprocess.py
TallJimbo: I think it's fine - all current usage is in this file, and I think other usage is more likely to be in `pipe_tasks` than upstream of it.

python/lsst/pipe/tasks/postprocess.py
TallJimbo: Yeah, I meant "strict" as in, "it's on you, the caller, to guarantee this, because I'm gonna assume it's the case."  But I don't think the checks would be that expensive, as long as it's just a numpy dtype comparison for each column block.  I'll add it.

PR #1011, opened 2024-11-22
DM-46991: Add PreSource.yaml and update Source.yaml to work with calibrateImage.

TallJimbo:
These need to differ only because they're being produced by different tasks with different ways of handling IDs.

I'm making minor adjustments instead of switching to the existing configuration for initial_stars_detector because I don't want to reconfigure downstream analysis tasks (which require many measurement columns that calibrateImage doesn't run by default) on this ticket.  I expect the final configuration to be somewhere in between.

### DIFF
--- a/schemas/PreSource.yaml
+++ b/schemas/PreSource.yaml
@@ -4,7 +4,8 @@
 # See the DPDD for more information about the output: https://lse-163.lsst.io
 funcs:
     sourceId:
-        functor: Index
+        functor: Column
+        args: id
     coord_ra:
         # reference position required by db. Not in DPDD
         functor: CoordColumn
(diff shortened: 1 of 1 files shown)


schemas/PreSource.yaml
```diff
@@ -0,0 +1,456 @@
+# This file defines the mapping between the columns in a single visit+detector
+# source table and their respective DPDD-style column names, as used by
+# `lsst.pipe.tasks.postprocess.TransformSourceTableTask`.
+# See the DPDD for more information about the output: https://lse-163.lsst.io
+funcs:
+    sourceId:
+        functor: Column
+        args: id
```
TallJimbo, line 8: The two lines above are the only difference between this file and `Source.yaml`.  I should be able to unify them again on [DM-34875](https://rubinobs.atlassian.net/browse/DM-34875), which makes `writeRecalibratedSourceTable` proudce an `ArrowAstropy` output just like `calibrateImage` does.

natelust (approved):

PR #1012, opened 2024-11-27
DM-47796: Deprecate some coadd-processing connections and configs.

TallJimbo:


python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
```diff
@@ -213,12 +229,17 @@ def run(self, coadd, mergedDetections, idFactory):
         return Struct(measureCatalog=sources)
 
 
-class DeblendCoaddSourcesMultiTask(DeblendCoaddSourcesBaseTask):
+class DeblendCoaddSourcesMultiTask(PipelineTask):
```
fred3m, line 232: Do we want to rename this to `DeblendCoaddSourcesTask` now that there is no `DeblendCoaddSouresSingleTask`?

fred3m (approved):
It looks like you got everything here. I left one suggestion to also deprecate the multi-part of the task name, since it only runs multiband. Other than that this LGTM.

python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
TallJimbo: I don't think it's worth the additional disruption right now (though if we did rename it, I'd want to get `Sources` out of the names of all of the coadd-processing tasks, too).

PR #1014, opened 2024-11-27
DM-47787: Add check on psf size to MeasurePsfTask

parejkoj:
This will catch bad psfs returned from both psfex and piff.

Unfortunately, there are no tests for MeasurePsf, so there's not a good place to put a test of this new exception (e.g. with a mocked-bad psf). I checked it with one of the calibrateImage tests, removing the `not` and seeing that the appropriate exception was raised.

### DIFF
--- a/python/lsst/pipe/tasks/measurePsf.py
+++ b/python/lsst/pipe/tasks/measurePsf.py
@@ -21,6 +21,8 @@
 
 __all__ = ["MeasurePsfConfig", "MeasurePsfTask"]
 
+import numpy as np
+
 import lsst.afw.display as afwDisplay
 import lsst.afw.math as afwMath
 import lsst.meas.algorithms as measAlg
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/measurePsf.py
```diff
@@ -242,6 +265,10 @@ def run(self, exposure, sources, expId=0, matches=None):
                                                        flagKey=self.usedKey)
         self.log.info("PSF determination using %d/%d stars.",
                       self.metadata.getScalar("numGoodStars"), self.metadata.getScalar("numAvailStars"))
+        if not np.isfinite((psfSize := psf.computeShape(psf.getAveragePosition()).getDeterminantRadius())):
+            raise NonfinitePsfShapeError(psf_size=psfSize)
```
isullivan, line 269: We should document that this method can now raise a `NonfinitePsfShapeError`

isullivan (approved):
I verified that this raises as expected on the dataId that triggered the original ticket.

PR #1019, opened 2025-01-07
DM-48145: Improve calibrateImage partial-outputs failure handling (including downstream)

TallJimbo:


python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -695,8 +695,8 @@ def compute_psf_and_ap_corr_map(self, visit, detector, exposure, src, isolated_s
         try:
             psf_selection_result = self.make_psf_candidates.run(selected_src, exposure=exposure)
         except Exception as e:
-            self.log.warning('Failed to make psf candidates for visit %d, detector %d: %s',
-                             visit, detector, e)
+            self.log.exception('Failed to make psf candidates for visit %d, detector %d: %s',
```
leeskelvin, line 705: While we're here, can we change `psf` to `PSF`? And on L716 as well.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -631,34 +640,71 @@ def run(self, *, exposures, id_generator=None, result=None):
...
+        result.background = None
+        summary_stat_catalog = None
+        # Some exposure components are set to initial placeholder objects
+        # while we try to bootstrap them.  If we fail before we fit for them,
+        # we want to reset those components to None so the placeholders don't
+        # masquerade as the real thing.
+        have_fit_psf = False
+        have_fit_astrometry = False
+        have_fit_photometry = False
+        try:
```
leeskelvin, line 654: I note that you also asked @parejkoj about this block. I also can't think of a better alternative. My go-to would normally be to break this down into smaller sub-blocks, one for the PSF, one for photometry, one for astrometry, but that would involve a fair amount of logic when calling to ensure the prior step succeeded. With that in mind, I think I can go along with this large try/except block here too.

leeskelvin (approved):
Thanks Jim, this looks good to me. The changes make sense, and the clean-ups are welcome.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -631,34 +640,71 @@ def run(self, *, exposures, id_generator=None, result=None):
...
+                result.photometry_matches = lsst.meas.astrom.denormalizeMatches(photometry_matches,
+                                                                                photometry_meta)
+        except pipeBase.AlgorithmError:
+            if not have_fit_psf:
+                result.exposure.setPsf(None)
+            if not have_fit_astrometry:
+                result.exposure.setWcs(None)
+            if not have_fit_photometry:
+                result.exposure.setPhotoCalib(None)
+            # Summary stat calculations can handle missing components gracefully,
```
parejkoj, line 697: This and the other long comments are useful for understanding the logic, but I'm definitely worried that this commit requires a couple of paragraphs of comments to explain why it's implemented this way.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -631,34 +640,71 @@ def run(self, *, exposures, id_generator=None, result=None):
 
         result.exposure = self.snap_combine.run(exposures).exposure
         self._recordMaskedPixelFractions(result.exposure)
+        self.log.info("Initial PhotoCalib: %s", result.exposure.getPhotoCalib())
+
+        result.background = None
+        summary_stat_catalog = None
+        # Some exposure components are set to initial placeholder objects
```
parejkoj, line 647: This is making me want the "this component is valid" property to be something that rides along with the persisted data. I don't mean making a component `None`, I mean the component itself having a `isValid` of some kind. That's a much bigger task than this ticket though (and probably something to delay until shoefits). I'd considered something like your approach here when implementing the original AnnotatedError system, but it wasn't necessary at the time, and I didn't like how it complicated the code; I still don't, but given how many downstream tasks were implemented to work with whatever is passed to them, I think it may be the least worst current option.

The initial values aren't really "placeholders", though, they're the observatory's best guess at that component: PSF and photometry should be None (since we don't get information about those from the telescope), but astrometry is probably reasonable. This goes back to the old debate about whether we should clear header values that we then do something with (e.g. ones that go into VisitInfo). I think marking things bad, i.e. `None`, when we couldn't fit them is fine, but it does remove information about what the system thought it was doing. 

We should probably do something similar in `ReprocessVisitImage`, to make sure components that didn't get a visit-level fit are cleared, no? I'm not sure how all that flows through, though, given the number of Tasks involved.

TallJimbo: Yeah, some sort of validity flag would definitely be better, but we just can't afford to put effort into it right now.

My intent for `ReprocessVisitImage` was that it would actually give up and not write anything if any of its inputs were not available (and that includes giving up if some preliminary input was available but the final one was not).  At that stage we've past the point of being able to recover from problems and I right now think it's more important to ensure the output data products are consistent across all data IDs vs. present for as many data IDs as possible.  I should go make sure that's actually the case, though...

TallJimbo: Confirmed,  `ReprocessVisitImage` gives up if any component is missing: https://github.com/lsst/drp_tasks/blob/6b83ee06942be41d1c30d07f791d03b89f7ac071/python/lsst/drp/tasks/reprocess_visit_image.py#L350-L366

parejkoj: Ok, I think that's the correct approach for that task.

PR #1020, opened 2025-01-07
DM-7847: Add mid-level drivers for measurement algorithms

enourbakhsh:


python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,379 @@
...
+            assert self.deblend.target == measDeblender.SourceDeblendTask
+        elif self.deblender is not None:
+            raise ValueError(f"Invalid deblender value: {self.deblender}")
+
+    def setDefaults(self):
+        super().setDefaults()
+        if self.deblender == "scarlet":
+            self.deblend.retarget(scarlet.ScarletDeblendTask)
+        elif self.deblender == "meas_deblender":
+            self.deblend.retarget(measDeblender.SourceDeblendTask)
```
enourbakhsh, line 93: Would it be better to rename this method to `setDeblender()` and just call it in the `setDefaults` method? This way my`__setattr__` method becomes:
```python
def __setattr__(self, key, value):
    super().__setattr__(key, value)
    if key == "deblender":
        self.setDeblender()
```
It just makes more sense to me this way.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,379 @@
...
+        image,
+        bands=None,
+        band=None,
+        mask=None,
+        variance=None,
+        psf=None,
+        wcs=None,
+        photo_calib=None,
+        id_generator=None,
+    ):
```
fred3m, line 173: I would recommend using type hints in your function signature, then you can remove the types from the docstring of Parameters. This prevents bitrotting and will also help people using an IDE with a mypy-like linter.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,379 @@
...
+            provided as an `Exposure`, `MultibandExposure`, or a list of
+            `Exposure` objects.
+        wcs: `~lsst.afw.image.Wcs`, optional
+            The World Coordinate System (WCS) model for the input image. Will
+            be ignored if ``image`` is provided as an `Exposure`,
+            `MultibandExposure`, or a list of `Exposure` objects.
+        photo_calib : `~lsst.afw.image.PhotoCalib`, optional
+            Photometric calibration model for the input image. Will be ignored
+            if ``image`` is provided as an `Exposure`, `MultibandExposure`, or
+            a list of `Exposure` objects.
```
fred3m, line 213: I think that this API is confusing. Instead of having three optional sets of parameters in `run`, one for `Exposure`s, and one for `Image`s (which requires a lot of additional parameters that are ignored if the `image` is an `Exposure`), and one for `MultibandExposure`s, I recommend making run take an `Exposure` and create a new method, `runFromImage` that takes the parameters required if an image is passed and creates an exposure from those images to call the normal `run`. I think that this will make the API easier for users and simplify the codebase. See my other comments about single-band vs multiband

fred3m (changes requested):
After looking at the code I actually think that it would be better to separate this into multiple tasks, a `SingleBandMeasurementDriverTask` and a `MultiBandMeasurementDriverTask` that both inherit from a `MeasurementDriverBaseTask`. I would also break up `run` into multiple subtasks that are implemented in the MeasurementDriverBaseTask (ie. `_detection` and `measurement`) that are the same for both the single-band and multi-band versions.

This would allow you to give more flexibility to the multi-band version, which could optionally use https://github.com/lsst/drp_tasks/blob/main/python/lsst/drp/tasks/assemble_chi2_coadd.py to build a chi^2 coadd and use that for detection (as opposed to having to choose a reference band). So I would make `band` a config option and if `band` is `None` (the default), it would detect on a chi^2 coadd and perform measurement in all bands, producing a measurement catalog in each band. Otherwise if a user wants measurements in multiple bands then they would have to run the driver multiple times, doing detection and deblending again even though the results will be the same.

python/lsst/pipe/tasks/measurementDriver.py
enourbakhsh: Done! I also did it in other places.

python/lsst/pipe/tasks/measurementDriver.py
enourbakhsh: I like all these suggestions. I implemented them in the new version.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,379 @@
...
+    detection = pexConfig.ConfigurableField(
+        target=measAlgorithms.SourceDetectionTask,
+        doc="Task to detect sources to return in the output catalog.",
+    )
+
+    deblender = pexConfig.ChoiceField[str](
+        doc="The deblender to use.",
+        default="meas_deblender",
+        allowed={"meas_deblender": "Deblend using meas_deblender", "scarlet": "Deblend using scarlet"},
+    )
```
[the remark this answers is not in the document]
enourbakhsh: The problem is that I couldn’t directly set the deblender in `deblend` because it’s defined as a `ConfigurableField`, which can only be retargeted. This was done to enable subconfigs like `config.deblend.tinyFootprintSize = 3`. I left it as is, but if I missed something and you know a workaround, let me know — I’ll give it a try.

python/lsst/pipe/tasks/measurementDriver.py
enourbakhsh: See my comment about why I used both `deblend` and `deblender` in the config.

enourbakhsh (commented):
Thanks @fred3m for your thorough review.

> After looking at the code I actually think that it would be better to separate this into multiple tasks, a `SingleBandMeasurementDriverTask` and a `MultiBandMeasurementDriverTask` that both inherit from a `MeasurementDriverBaseTask`. I would also break up `run` into multiple subtasks that are implemented in the `MeasurementDriverBaseTask` (ie. `_detection` and `measurement`) that are the same for both the single-band and multi-band versions.

Done! This was definitely an improvement.

> This would allow you to give more flexibility to the multi-band version, which could optionally use https://github.com/lsst/drp_tasks/blob/main/python/lsst/drp/tasks/assemble_chi2_coadd.py to build a chi^2 coadd and use that for detection (as opposed to having to choose a reference band). So I would make `band` a config option and if `band` is `None` (the default), it would detect on a chi^2 coadd and perform measurement in all bands, producing a measurement catalog in each band. Otherwise if a user wants measurements in multiple bands then they would have to run the driver multiple times, doing detection and deblending again even though the results will be the same.

Good idea. I'll include it in my improvement ticket if you're okay with it.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,519 @@
...
+
+
+class MeasurementDriverBaseTask(pipeBase.Task):
+    """Base class for the mid-level driver running detection, deblending
+    (optional), and measurement algorithms in one go.
+
+    This driver simplifies the process of applying a small set of measurement
+    algorithms to images by abstracting away schema and table boilerplate. It
+    is particularly suited for simple use cases, such as processing images
+    without neighbor-noise-replacement or extensive configuration.
```
enourbakhsh, line 123: This *neighbor-noise-replacement* part was added based on the ticket description, but I’m not sure why it’s such a big deal for the driver task. It can be turned on or off easily with `config.measurement.doReplaceWithNoise`.

python/lsst/pipe/tasks/measurementDriver.py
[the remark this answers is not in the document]
fred3m: Sorry, I was thinking that you remove the field from the config entirely. What I was thinking is that you could use
```
if isinstance(self.config.deblender, ScarletDeblendTask):
    # use meas_extensions_scarlet deblending
else:
    # use meas_deblender
```
and add a `doDeblend` field like other tasks use. While it's still the same number of parameters, it eliminates all of the extra methods in the config and I think that it's a cleaner API, as you could default to multi-band using `ScarletDeblendTask` in `MultiBandMeasurementDriverConfig.deblender` and single-band using `SourceDeblendTask` in `SingleBandMeasurementDriverConfig.deblender`.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,519 @@
...
+        self._measureSources()
+
+        return self.catalog
+
+    def _detectSources(self):
+        """Run the detection subtask to identify sources in the image."""
+        self.log.info(f"Running detection on a {self.exposure.width}x{self.exposure.height} pixel exposure")
+        self.catalog = self.detection.run(self.catalog, self.exposure).sources
+
+    def _deblendSources(self):
```
fred3m, line 226: I would make _deblendSources an abstract method and override it in the single-band and multi-band versions.

fred3m (changes requested):
Thanks for making the changes @enourbakhsh, I do like this version much better.

Looking back at the ticket I think that there's still some desired functionality that's missing. It might help to model it after MeasureMergedCoaddSourcesTask in https://github.com/lsst/pipe_tasks/blob/main/python/lsst/pipe/tasks/multiBand.py with a few convenience additions:
- add a `doDetect`, `doDeblend`, and `doMeasure` config options so that a user can run any subset of the tasks.
- Add an optional `catalog` parameter to the `run` method so that if `doDetect` is off, deblending and measurement can still be done by extending the input catalog
- To make the above work you'll need to move the schema code into an `_initializeSchema` method that can be called from `run` if `doDetect` is `False` and a catalog is passed to the `run` method. To me this is part of the desired functionality to reduce the schema boilerplate, as I (and I think others) often forget house to create a new source catalog with the appropriate schema and have all of the columns copied over correctly.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,519 @@
...
+            self.schema = afwTable.SourceTable.makeMinimalSchema()
+        else:
+            self.schema = schema
+
+        # Add coordinate error fields to avoid missing field issues in the
+        # schema.
+        afwTable.CoordKey.addErrorFields(self.schema)
+
+        # Standard subtasks to run in sequence.
+        self.subtaskNames = ["detection", "deblend", "measurement"]
```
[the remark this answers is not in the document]
enourbakhsh: Done!

python/lsst/pipe/tasks/measurementDriver.py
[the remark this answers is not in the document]
enourbakhsh: Followed your suggestion and made `_deblendSources` abstract, with each subclass having its own `deblend` configurable field and overriding the method in the single-band and multi-band versions.

enourbakhsh (commented):
> Looking back at the ticket I think that there's still some desired functionality that's missing. It might help to model it after `MeasureMergedCoaddSourcesTask` in https://github.com/lsst/pipe_tasks/blob/main/python/lsst/pipe/tasks/multiBand.py with a few convenience additions:

Added some functionalities from `MeasureMergedCoaddSourcesTask`. Let me know if it is not enough.

> - add a doDetect, doDeblend, and doMeasure config options so that a user can run any subset of the tasks.

Done!

> - Add an optional catalog parameter to the run method so that if `doDetect` is off, deblending and measurement can still be done by extending the input catalog

Done! I'm also curious about handling cases where we input a deblended catalog and only need to run `measurement`. What's the safest way to distinguish between a post-detection and post-deblending catalog so we can use that in input validation?

> To make the above work you'll need to move the schema code into an `_initializeSchema` method that can be called from `run` if `doDetect` is `False` and a catalog is passed to the run method. To me this is part of the desired functionality to reduce the schema boilerplate, as I (and I think others) often forget house to create a new source catalog with the appropriate schema and have all of the columns copied over correctly.

Done! The schema part was a bit of a headache, but it should work now.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,690 @@
...
+        self.exposure = self.mExposure[self.band]
+        self.log.info(f"Using '{self.band}' band as the reference band for band-specific tasks")
+
+        self.catalog = catalog
+        self.idGenerator = idGenerator
+
+        return super().run()
+
+    def _deblendSources(self):
+        self.catalog, modelData = self.deblend.run(mExposure=self.mExposure, mergedSources=self.catalog)
```
fred3m, line 623: modelData should be added to the struct that is returned, as it contains important information about blends that is often useful for debugging.

fred3m (changes requested):
Thanks for integrating all of the changes. I think that there is still a little more needed to clean up the task. Thanks for your patience, but it's really close now.

enourbakhsh (commented):
Thanks for the review! I applied most suggestions as given but handled a few differently.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,1052 @@
...
+            # the same point coming from different measurement plugins.
+            if self.config.psfCache > 0:
+                # Set a hard limit on the number of PSFs to cache.
+                exp.psf.setCacheCapacity(self.config.psfCache)
+            else:
+                # Auto-size the cache based on the number of measurement
+                # plugins. We assume each algorithm tries to evaluate the PSF
+                # twice, which is more than enough since many don't evaluate it
+                # at all, and there's no *good* reason for any algorithm to
+                # evaluate it more than once.
```
enourbakhsh, line 320: Borrowed this from [here](https://github.com/lsst/drp_tasks/blob/15d4fdd68ac8aaa73b705fe3093f7fe28bdc657e/python/lsst/drp/tasks/forcedPhotCoadd.py#L324).

fred3m (changes requested):
I see what you're trying to do but I still think that this is unnecessarily complicated and often not what the user intends. Multi-band processing is fundamentally different in a lot of ways from single-band processing and I imagine that more updates to these drivers in the future will make them diverge even more.

For example, in multi-band it isn't likely that the user will want to perform detection in a single reference band. While I don't want it to hold up this ticket, and myself or someone else can implement it in the future, we'll either want to build a chi^2 coadd that combines all of the bands and do detection on that or do detection on all of the bands and merge those detections in the same way that the current DRP pipeline detects on coadds.

Also, even multi-band deblending exports a single catalog, which is what the user likely wants if they are just running processing through deblending, so they should only get one catalog as a result. And in some cases maybe the user _does_ only want to run measurement in a single band.

So I think that it's good to include things like  `detectSources`, `_measureSources`, `_toContiguous`, etc. in the main driver task like you're currently doing, but they should be called from different `run` methods.

enourbakhsh (commented):
> I see what you're trying to do but I still think that this is unnecessarily complicated and often not what the user intends. Multi-band processing is fundamentally different in a lot of ways from single-band processing and I imagine that more updates to these drivers in the future will make them diverge even more.
> 

Good point, I’ve made the `run` method in the base class purely abstract and added separate `run` methods for the driver tasks, as you suggested.

> For example, in multi-band it isn't likely that the user will want to perform detection in a single reference band. While I don't want it to hold up this ticket, and myself or someone else can implement it in the future, we'll either want to build a chi^2 coadd that combines all of the bands and do detection on that or do detection on all of the bands and merge those detections in the same way that the current DRP pipeline detects on coadds.
> 

These are all great ideas - I’ve filed [DM-48927](https://rubinobs.atlassian.net/browse/DM-48927) for them.

> Also, even multi-band deblending exports a single catalog, which is what the user likely wants if they are just running processing through deblending, so they should only get one catalog as a result. And in some cases maybe the user _does_ only want to run measurement in a single band.
> 

Oh, I made the multi-band driver output multiple catalogs instead of just one, because I was slightly confused by your previous comment:

>> Maybe call this `refBand`, since it's the reference band used for detection. You could potentially add a config option `useReferenceForMeasurement` (or something like that) to the `MultiBandMeasurementDriverConfig` class that tells it to run measurement on only the reference band, otherwise it will default to generating a catalog in each band.

That said, [DM-48927](https://rubinobs.atlassian.net/browse/DM-48927) will further explore these improvements.

> So I think that it's good to include things like `detectSources`, `_measureSources`, `_toContiguous`, etc. in the main driver task like you're currently doing, but they should be called from different `run` methods.

They’re now being called from separate `run` methods.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,690 @@
...
+        Might seem a bit clunky, but this workaround is necessary to ensure
+        that the Schema is consistent across all subtasks.
+        """
+        if self.catalog is None:
+            if self.schema is None:
+                # Create a minimal Schema that will be extended by tasks.
+                self.schema = afwTable.SourceTable.makeMinimalSchema()
+
+                # Add coordinate error fields to avoid missing field issues.
+                afwTable.CoordKey.addErrorFields(self.schema)
```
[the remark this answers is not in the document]
fred3m: Yes, I agree, sorry if I wasn't clear. My point was that there may be some times where a user passes a catalog that does not have those columns and it will cause measurement to fail. I think that you might need to check the schema of the catalog to see if it already has `coord_raError` and _also_ call `afwTable.CoordKey.addErrorFields(self.schema)` in that case too.

python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -0,0 +1,1171 @@
...
+            `measureOnlyInRefBand` is enabled, this will only contain the
+            reference-band catalog; otherwise, it will contain a catalog for
+            each band.
+        modelData :
+            Multiband scarlet models produced during deblending.
+        """
+        self.log.info(f"Deblending using '{self._Deblender}' on {len(catalog)} detection footprints")
+
+        # Run the deblender on the multi-band exposure.
+        catalog, modelData = self.deblend.run(mExposure, catalog)
```
fred3m, line 1104: Please cache modelData (in other words `self.modelData = modelData`. This will be useful for debugging.

fred3m (approved):
Looks great! Thanks for your patience, I just left a few small comments.

python/lsst/pipe/tasks/measurementDriver.py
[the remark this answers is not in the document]
enourbakhsh: Done!

python/lsst/pipe/tasks/measurementDriver.py
enourbakhsh: Done here:
```python
if self.config.doDeblend:
    catalogs, self.modelData = self._deblendSources(mExposure, catalog, refBand=refBand)
```

PR #1023, opened 2025-01-14
DM-47526: Add allow_missing_bands for multiband object fitting

taranu:


python/lsst/pipe/tasks/fit_coadd_multiband.py
```diff
@@ -323,14 +359,33 @@ def build_catexps(self, butlerQC, inputRefs, inputs) -> list[CatalogExposureInpu
...
+                # This should be handled by any subtasks
+                dataId_band = dataId.to_simple(minimal=True)
+                dataId_band.dataId["band"] = band
+                catexp = CatalogExposureInputs(
+                    catalog=afwTable.SourceCatalog(),
+                    exposure=None,
+                    table_psf_fits=astropy.table.Table(),
+                    dataId=dataId.from_simple(dataId_band, universe=dataId.universe),
+                    id_tract_patch=id_tp,
+                )
```
fred3m, line 387: I wonder if it would be useful to have an `EmptyExposureInputs` class that inherits from `CatalogExposureInputs` but only takes a dataId and id_tract_patch. This would make it clear here that this is just a placeholder.

fred3m (approved):
LGTM.

python/lsst/pipe/tasks/fit_coadd_multiband.py
taranu: Hmm, it's possible. I was also consider just implementing an `is_dummy`/`is_placeholder` method that just returns `len(self.catalog) == 0`. I decided to defer committing to a particular interface until I have a clearer understanding of the benefits/drawbacks of dummy data.

PR #1024, opened 2025-01-15
DM-47730: Use new refcat loader method to improve refcat flux field checks.

TallJimbo:
The new implementation no longer assumes there is a refcat shard overlapping the center of a tract, which was important for ci_hsc.

### DIFF
--- a/tests/test_loadReferenceCatalog.py
+++ b/tests/test_loadReferenceCatalog.py
@@ -63,6 +63,14 @@ class TrivialLoader(ReferenceObjectLoader):
     def loadPixelBox(self, bbox, wcs, referenceFilter, **kwargs):
         return self.loadSkyCircle(None, None, referenceFilter)
 
+    def loadSchema(self, filterName):
+        refCat = self.make_synthetic_refcat(_synthCenter, _synthFlux)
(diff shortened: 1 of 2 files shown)


tests/test_loadReferenceCatalog.py
```diff
@@ -63,6 +63,14 @@ def loadSkyCircle(self, ctrCoord, radius, filterName, **kwargs):
     def loadPixelBox(self, bbox, wcs, referenceFilter, **kwargs):
         return self.loadSkyCircle(None, None, referenceFilter)
 
+    def loadSchema(self, filterName):
```
jrmullaney, line 66: When does this get used within this test suite?
I guess what I'm asking is - why is it needed (not a criticism, just a query)?

jrmullaney (approved):
Looks good. I'm just curious aboout something in the test suite.

tests/test_loadReferenceCatalog.py
TallJimbo: The base class implementation of this method that I just added in `meas_algorithms` uses `self.refCats`, which is set to an empty list by this testing subclass.  So I needed to override it here to do something else.

It's used whenever `LoadReferenceCatalogTask._determineFluxFields` is called.

PR #1025, opened 2025-01-29
DM-47632: Make summary stat calculation more robust on partial-output error cases.

TallJimbo:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -658,7 +653,6 @@ def run(self, *, exposures, id_generator=None, result=None):
             result.psf_stars_footprints, result.background, _ = self._compute_psf(result.exposure,
                                                                                   id_generator)
             have_fit_psf = True
-            summary_stat_catalog = result.psf_stars_footprints
```
erykoff, line 661: From looking at the current version of the code, my understanding was that this was set here in case the next task failed and it bumped out to the except block.  Is this no longer needed?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -410,11 +410,6 @@ def setDefaults(self):
         self.photometry.match.sourceSelection.doRequirePrimary = False
         self.photometry.match.sourceSelection.doUnresolved = False
 
-        # All sources should be good for PSF summary statistics.
-        # TODO: These should both be changed to calib_psf_used with DM-41640.
-        self.compute_summary_stats.starSelection = "calib_photometry_used"
-        self.compute_summary_stats.starSelector.flags.good = ["calib_photometry_used"]
```
erykoff, line 416: Why is this removed and not changed to (e.g.) `calib_psf_used`?

TallJimbo: `calib_psf_used` is already the default.

python/lsst/pipe/tasks/calibrateImage.py
TallJimbo: I was originally trying to have the summary statistics calculations (specifically the PSF ones, which are the only ones that use a catalog at all) use the initial set of detections and measurements when the second round wasn't available due to a failure.  But it turns out it doesn't have the right columns for that to work, and I didn't think it was worth it to add more columns to the initial round or add more logic to the summary stats to make it work.  There just aren't enough failures in the particular window where this fallback is relevant (it's basically just aperture correction measurement failures) to be worth the special-casing to get a little bit more summary stat info out on detectors we can't use downstream anyway.

erykoff (approved):

PR #1026, opened 2025-02-02
DM-48316: Add model PSF ellipticity thresholds in single frame processing

laurenam:


jrmullaney (approved):
LGTM.
Should the new metrics be included in the test_computeExposureSummaryStats (although I note that none of the other star-derived metrics are in the test suite)?

PR #1027, opened 2025-02-03
DM-35396: Strip provenance headers before merging tables

timj:
Without this the astropy merge code complains that duplicate headers exist and picks the first value. Butler will update the provenance itself.

### DIFF
--- a/python/lsst/pipe/tasks/postprocess.py
+++ b/python/lsst/pipe/tasks/postprocess.py
@@ -1345,6 +1345,9 @@ class MakeCcdVisitTableTask(pipeBase.PipelineTask):
                 continue
             visitInfo = visitSummary[0].getVisitInfo()
 
+            # Strip provenance to prevent merge confusion.
+            strip_provenance_from_fits_header(visitSummary.metadata)
+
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -156,7 +157,11 @@ def extend(self, table):
                     raise TypeError(f"Type mismatch on column {name!r}: {out_col.dtype} != {in_col.dtype}.")
                 self.result[name][self.index:next_index] = table[name]
             self.index = next_index
-            self.result.meta = astropy.utils.metadata.merge(self.result.meta, table.meta)
+            # Butler provenance should be stripped on merge. It will be
+            # added by butler on write. No attempt is made here to combine
+            # provenance from multiple input tables.
+            self.result.meta = merge_headers([self.result.meta, table.meta], mode="drop")
```
timj, line 163: Previously it was merging by taking the first value and warning. This change drops duplicate headers. `first` can be used instead.

TallJimbo (approved):

PR #1028, opened 2025-02-03
DM-48730: Update healsparse bad pixel mask to be faster and use much less memory.

erykoff:
The previous version was using a ton of memory trying to fill in NO_DATA regions unnecessarily.  The new version will only fill in bad pixels that have expected coverage, and does it in a much more memory efficient way using healsparse API.

### DIFF
--- a/python/lsst/pipe/tasks/healSparseMapping.py
+++ b/python/lsst/pipe/tasks/healSparseMapping.py
@@ -27,6 +27,7 @@ __all__ = ["HealSparseInputMapTask", "HealSparseInputMapConfig",
            "ConsolidateHealSparsePropertyMapTask"]
 
 from collections import defaultdict
+import esutil
 import warnings
 import numbers
 import numpy as np
@@ -207,8 +208,6 @@ class HealSparseInputMapTask(pipeBase.Task):
         self.ccd_input_map.metadata = metadata
 
         # Create a temporary map to hold the count of bad pixels in each healpix pixel
-        self._ccd_input_pixels = self.ccd_input_map.valid_pixels
-
         dtype = [(f"v{visit}", np.int64) for visit in self._bits_per_visit.keys()]
 
         with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/healSparseMapping.py
```diff
@@ -250,6 +251,10 @@ def mask_warp_bbox(self, bbox, visit, mask, bit_mask_value):
         if self.ccd_input_map is None:
             raise RuntimeError("Must run build_ccd_input_map before mask_warp_bbox")
 
+        if len(self._ccd_input_pixels) == 0:
```
yalsayyad, line 254: This is perfectly readable, and you don't need to change this one, but in future dev guide prefers using the fact that empty sequences are false: https://developer.lsst.io/python/style.html#comparisons like 
```if not self._ccd_input_pixels:```

File: python/lsst/pipe/tasks/healSparseMapping.py [hunk 2]
```diff
@@ -262,23 +267,20 @@ def mask_warp_bbox(self, bbox, visit, mask, bit_mask_value):
...
-
-        # Add these to the accumulator map.
-        # We need to make sure that the "primary" array has valid values for
-        # this pixel to be registered in the accumulator map.
-        pix_to_add, = np.where(bad_hpix_count > 0)
-        count_map_arr = self._ccd_input_bad_count_map[min_bad_hpix + pix_to_add]
-        primary = self._ccd_input_bad_count_map.primary
-        count_map_arr[primary] = np.clip(count_map_arr[primary], 0, None)
+        # Check if any of these "bad" pixels are in the valid footprint.
+        match_input, match_bad = esutil.numpy_util.match(self._ccd_input_pixels, bad_hpix, presorted=True)
```
yalsayyad, line 271: There's no possible way  `self._ccd_input_pixels` wouldn't be sorted right?

yalsayyad (approved):

python/lsst/pipe/tasks/healSparseMapping.py
erykoff: It should be impossible with the way we are making the healsparse map, but I have added an explicit sorting step when they are made just in case.

python/lsst/pipe/tasks/healSparseMapping.py
erykoff: You know, I never liked that idiom but decided to try it and it was a 100% disastrous failure.  That idiom *does not work with numpy arrays*.

PR #1035, opened 2025-02-09
Remove the too stringent test from test_coaddBase.py

arunkannawadi:


Gerenjie (approved):
Looks good to me pending the Jenkins pass!

PR #1038, opened 2025-02-12
DM-30993: Implement reference catalog culling for astrometry in SFM

laurenam:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -402,7 +402,11 @@ def setDefaults(self):
 
         # Only reject sky sources; we already selected good stars.
         self.astrometry.sourceSelector["science"].doFlags = True
-        self.astrometry.sourceSelector["science"].flags.bad = ["sky_source"]
+        self.astrometry.sourceSelector["science"].flags.good = ["calib_psf_candidate"]
+        self.astrometry.sourceSelector["science"].flags.bad = []
```
isullivan, line 406: Why is `"sky_source"`  removed from the bad flag list? I imagine if someone overrides the good flag list, they would want sky sources removed no matter what.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -482,23 +486,25 @@ def __init__(self, initial_stars_schema=None, **kwargs):
...
             initial_stars_schema = afwTable.SourceTable.makeMinimalSchema()
 
-        # These fields let us track which sources were used for psf and
+        # These fields let us track which sources were used for psf modeling
         # aperture correction calculations.
         self.psf_fields = ("calib_psf_candidate", "calib_psf_used", "calib_psf_reserved",
                            # TODO DM-39203: these can be removed once apcorr is gone.
                            "apcorr_slot_CalibFlux_used", "apcorr_base_GaussianFlux_used",
-                           "apcorr_base_PsfFlux_used")
+                           "apcorr_base_PsfFlux_used", "calib_astrometry_used")
```
isullivan, line 507: I suggest adding `"calib_astrometry_used"` before the TODO comment, so it is clear what can be removed on that ticket.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -654,22 +657,32 @@ def run(self, *, exposures, id_generator=None, result=None):
...
+            if "astrometry_matches" in self.config.optional_outputs:
+                result.astrometry_matches = lsst.meas.astrom.denormalizeMatches(astrometry_matches,
+                                                                                astrometry_meta)
 
+            # Run the stars_detection subtask for the photometric calibration.
             result.stars_footprints = self._find_stars(result.exposure, result.background, id_generator)
             self._match_psf_stars(result.psf_stars_footprints, result.stars_footprints)
+
+            # Update the source cooordinates with the current wcs.
+            afwTable.updateSourceCoords(result.exposure.wcs, sourceList=result.stars_footprints)
```
isullivan, line 675: Consider whether you should also set `include_covariance=True` to update the coordinate error fields.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -654,22 +657,32 @@ def run(self, *, exposures, id_generator=None, result=None):
...
+
             summary_stat_catalog = result.stars_footprints
             result.stars = result.stars_footprints.asAstropy()
             self.metadata["star_count"] = np.sum(~result.stars["sky_source"])
 
-            astrometry_matches, astrometry_meta = self._fit_astrometry(
-                result.exposure, result.stars_footprints
-            )
+            # Validate the astrometric fit. Send in the stars_footprints
+            # catalog so tha its coords get set to NaN if te fit is deemed
```
isullivan, line 682:
```suggestion
            # catalog so that its coords get set to NaN if the fit is deemed
```

isullivan (approved):
A few minor suggestions and questions

python/lsst/pipe/tasks/calibrateImage.py
laurenam: Because the sky sources aren't run at this stage.  I removed it because it caused an error due to this column not existing.

python/lsst/pipe/tasks/calibrateImage.py
laurenam: I think that's the default: https://github.com/lsst/afw/blob/main/include/lsst/afw/table/wcsUtils.h#L64

parejkoj:
Please update the `CalibrateImageTask.rst` task topic file to describe that the psf stars are now used for astrometry, and when that happens in the data flow.

PR #1039, opened 2025-02-13
DM-48649: Expose ImageScaler's scale

arunkannawadi:


python/lsst/pipe/tasks/scaleZeroPoint.py
```diff
@@ -43,6 +46,11 @@ class ImageScaler:
     def __init__(self, scale=1.0):
         self._scale = scale
 
+    @property
+    def scale(self) -> float:
+        """Scale that it applies to a specified image."""
+        return self._scale
+
```
arunkannawadi, line 53: I'm not sure if there's any utility in keeping an instance's `scale` unmodified once created.

TallJimbo: When in doubt, less mutability tends to be better.

TallJimbo (approved):

arunkannawadi:
The `scale` property won't work on `SpatialImageScaler` unless we redefine it appropriately. This needs a bit more work.

arunkannawadi:
Closing this PR since there's no longer a need to access `scale`, thanks to @fred3m 's heroic effort in making warps have pixels calibrated in `nJy`.

arunkannawadi:
Reopening this PR since the `scale` property is needed until we support making cell-coadds from the old warps. Introducing it as a deprecated property, and have a somewhat less-motivated one for the `SpatialImageScaler`.

TallJimbo (approved):

PR #1040, opened 2025-02-14
DM-48826 : add SSO association to drpAssociationPipe

Gerenjie:


python/lsst/pipe/tasks/associationUtils.py
```diff
@@ -193,3 +193,71 @@ def query_disc(nside, ra, dec, max_rad, min_rad=0):
         pixels = pixels[match]
 
     return pixels
+
+
+def objID_to_ssObjectID(objID, flags=0):
```
TallJimbo, line 198: Our naming conventions would write this as
```
obj_id_to_ss_object_id
```
or
```
objIdToSsObjectId
```
(same for other symbols in this file)

python/lsst/pipe/tasks/drpAssociationPipe.py
```diff
@@ -52,20 +55,66 @@ class DrpAssociationPipeConnections(pipeBase.PipelineTaskConnections,
...
+        multiple=True
+    )
+    wcsRefs = pipeBase.connectionTypes.Input(
+        doc="Reference to wcs of each exposure",
+        name="calexp.wcs",
+        storageClass="Wcs",
+        dimensions=("instrument", "visit", "detector"),
+        deferLoad=True,
+        multiple=True
+    )
```
TallJimbo, line 98: I would prefer for these (`visitInfo`, `bbox`, `wcs`)  to all be replaced by a single `finalVisitSummary` input, as I think that'll be a lot less I/O overall and pull in the best WCS instead of one we would not normally use at this stage (and in the future, might not want to keep on disk at this point in the processing).

Another possibility would be to use `pvi.<component>` instead of `calexp.<component>`, as `pvi` will be available and will have the final WCS, but I think using `finalVisitSummary` is still slightly better.

python/lsst/pipe/tasks/drpAssociationPipe.py
```diff
@@ -74,6 +123,13 @@ class DrpAssociationPipeConnections(pipeBase.PipelineTaskConnections,
         dimensions=("tract", "patch"),
     )
 
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+
+        if not config.doSolarSystemAssociation:
+            self.inputs.remove("solarSystemObjectTables")
+            self.outputs.remove("associatedSsSources")
```
TallJimbo, line 131: Should `unassociatedSsObjects` be removed here, too?

I'd also recommend slightly different syntax for this, just because it's a bit easier to read (what you have is not incorrect):
```suggestion
            del self.associatedSsSources
```

python/lsst/pipe/tasks/drpAssociationPipe.py
```diff
@@ -52,20 +55,66 @@ class DrpAssociationPipeConnections(pipeBase.PipelineTaskConnections,
         deferLoad=True,
         multiple=True
     )
+    ssObjectTableRefs = pipeBase.connectionTypes.Input(
+        doc="Reference to catalogs of SolarSolarSystem objects expected to be "
+            "observable in each (visit, detector).",
+        name="preloaded_DRP_SsObjects",
+        storageClass="ArrowAstropy",
+        dimensions=("instrument", "visit", "detector"),
+        minimum=0,
```
TallJimbo, line 64: I like this solution, if it means what I think it means; if you run this task, with `doSolarSystemAssociation=True`, but in a repo with no ephemerides for some or all of the visits, it just skips doing the solar system associations, and does regular DIAObject association only?  Does it warn?

TallJimbo (approved):

python/lsst/pipe/tasks/drpAssociationPipe.py
Gerenjie: Had I used your syntax, I would have noticed that I renamed `solarSystemObjectTables` into `ssObjectTableRefs`. Changed!

python/lsst/pipe/tasks/drpAssociationPipe.py
Gerenjie: That's better. I hadn't realized finalVisitSummary contained all those.

ebellm (commented):
A couple of small things.  I don't see any tests for `SolarSystemAssociationTask`--it would be valuable to have some, although I think it would be okay with me to implement them on another ticket.

PR #1041, opened 2025-02-18
DM-48783: Add illumination correction support to CalibrateImageTask and SkyCorrectionTask

erykoff:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -89,6 +89,21 @@ class CalibrateImageConnections(pipeBase.PipelineTaskConnections,
...
+        dimensions=["instrument", "detector", "physical_filter"],
+        isCalibration=True,
+    )
+    illumination_correction = connectionTypes.PrerequisiteInput(
+        name="illuminationCorrection",
+        doc="Illumination correction frame.",
+        storageClass="Exposure",
+        dimensions=["instrument", "detector", "physical_filter"],
+        isCalibration=True,
+    )
```
TallJimbo, line 105: So, someday when we have a `reference_flux_flat` that is is more sophisticated, is the idea that we'd have a `reference_flux_flat` connection _instead_ of this one, or that this task would keep the `illumination_correction` connection but also get some other input connection(s) (for at least dome flat, I think, and maybe some wavelength-dependent stuff, too)?

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -1027,6 +1201,7 @@ def _apply_photometry(self, exposure, background):
 
         assert photo_calib._isConstant, \
             "Background calibration assumes a constant PhotoCalib; PhotoCalTask should always return that."
+        # FIXME: what to do about background_to_photometric_ratio here?
```
TallJimbo, line 1204: I think "nothing", since we don't scale the persisted background model by the ratio when we do not apply the photometry to the image, either.  And that establishes the convention that if you want to add the background back in, you have to use the context manager.

The opposite convention - where the background subtraction task fits the background and then divides it by the `backgroundToPhotometricRatio` in order to let downstream code add it back in directly to a photometric-flattened image - might be less error-prone downstream, but I don't think the `BackgroundList` class is up to that, so let's be pragmatic and choose the convention we can actually implement right now.

python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -344,6 +384,14 @@ def run(self, calExps, calBkgs, skyFrames, camera):
             skyCorrExtras, camera, self.config.binning, ids=calExpIds, refExps=calExps
         )
 
+        # Question: should this uncorrection be done before the binning and display?
```
TallJimbo, line 387: I'd guess we probably want the display images to be flattened for background-subtraction.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -578,7 +653,15 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
-    def run(self, *, exposures, id_generator=None, result=None):
+    def run(
+        self,
+        *,
+        exposures,
+        id_generator=None,
+        result=None,
+        background_flat=None,
+        illumination_correction=None,
+    ):
```
TallJimbo, line 664: Just hanging this thread at a semi-arbitrary location: can we have this do a check of the input image metadata, to look for something ISR could put there, to confirm that it has been dome-flattened only?

TallJimbo (approved):

python/lsst/pipe/tasks/calibrateImage.py
erykoff: The answer to this partly depends on where we put the SED-matching of an SED-matched flat.  I don't have an answer to that right now.
But I do believe the input(s) will have to change somewhat, maybe just in an additive fashion.  The output -- the applied `background_to_photometric_ratio` -- I *hope* can stay the same, however.

python/lsst/pipe/tasks/calibrateImage.py
erykoff: I've added appropriate metadata for the isr tasks (which really should have been there all along), and check it here.  At the moment I'm raising a `RuntimeError` but I am not sure if that's appropriate.  (But it might be because it's a crashworthy programming mistake).

TallJimbo: I think `RuntimeError` is okay.  I think it's not an `AlgorithmError`.  It might be a `lsst.pipe.base.FatalQuantumError`, which means "stop everything, the configuration and/or inputs are bad!".

TallJimbo (approved):

PR #1043, opened 2025-02-22
DM-49097 : Create empty DataFrame rather than convert from Table

Gerenjie:
drpAssociationPipe would fail when no diaSources were found due to an astropy-to-Pandas conversion on an empty Table with a datetime64 column. The solution checks to see if the Table is empty before converting, in which case it replaces the Table with a new empty DataFrame with matching columns.

### DIFF
--- a/python/lsst/pipe/tasks/drpAssociationPipe.py
+++ b/python/lsst/pipe/tasks/drpAssociationPipe.py
@@ -275,7 +275,10 @@ class DrpAssociationPipeTask(pipeBase.PipelineTask):
                 unassociatedSsObjects = unassociatedSsObjects[ssObjInTractPatch]
                 nSsSrc = ssInTractPatch.sum()
                 nSsObj = ssObjInTractPatch.sum()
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/drpAssociationPipe.py
```diff
@@ -275,7 +275,10 @@ def visitDetectorPair(dataRef):
                 unassociatedSsObjects = unassociatedSsObjects[ssObjInTractPatch]
                 nSsSrc = ssInTractPatch.sum()
                 nSsObj = ssObjInTractPatch.sum()
-                diaCat = ssoAssocResult.unAssocDiaSources.to_pandas()
+                if len(ssoAssocResult.unAssocDiaSources) > 0:
```
yalsayyad, line 278: This is fine to leave and perfectly readable, but in the future, we prefer to use the fact that empty sequences are `False`: https://developer.lsst.io/python/style.html#is-and-is-not-should-be-used-when-comparing-to-none

yalsayyad (approved):

PR #1044, opened 2025-02-24
DM-47320: Add config option to more fully control property map names.

TallJimbo:


python/lsst/pipe/tasks/healSparseMapping.py
```diff
@@ -856,6 +882,25 @@ def __init__(self, *, config=None):
                 self.inputs.remove(f"{name}_map_sum")
                 self.outputs.remove(f"{name}_consolidated_map_sum")
 
+        if config.connections.prefix:
+            # If the 'prefix' connection template is set, replace
+            # '{coaddName}Coadd_' with that; this is a fully-backwards
+            # compatible way of overriding more than just the
+            # currently-configurable part of the prefix.
+            for connection_name in self.inputs | self.outputs:
+                if connection_name == "sky_map":
```
leeskelvin, line 891: Is it ok hard-coding this here? The alternative would be to configure it I imagine, but that may be unnecessary.

TallJimbo: Yeah, I think this is fine; it's hard-coding the name of a connection, not the name of the dataset type it's configured to, so it's no different from saying `self.sky_map`.

PR #1056, opened 2025-03-12
DM-49274: Stop sky objects appearing in NO_DATA regions

leeskelvin:


arunkannawadi (approved):
It'd be great to split commit acee1a36f833db907008e0fc449ee331102ea76b into two - one rearranging the names and one adding the nodata flag, but I am not insisting on it.

schemas/ForcedSource.yaml
```diff
@@ -85,15 +85,16 @@ calexpFlags:
...
+    - base_PixelFlags_flag_cr
+    - base_PixelFlags_flag_crCenter
     - base_PixelFlags_flag_edge
     - base_PixelFlags_flag_interpolated
-    - base_PixelFlags_flag_saturated
-    - base_PixelFlags_flag_cr
-    - base_PixelFlags_flag_bad
-    - base_PixelFlags_flag_suspect
     - base_PixelFlags_flag_interpolatedCenter
+    - base_PixelFlags_flag_nodata
```
mrawls, line 94: Please also add `base_PixelFlags_flag_nodataCenter` and then @jrmullaney won't have to

mrawls: I suggest doing this in all the schemas where other flags have "center" variants

jrmullaney: I'm not sure adding it in this block will work. Aren't these pulling flags from the calexp, rather than the diff? My understanding is that the nodata flag that I want in the forcedSourceTable comes from the diff catalog, not the calexp. See [here](https://github.com/lsst/pipe_tasks/blob/0a090414267e120f1476ae44f50a65f7bf309798/schemas/ForcedSource.yaml#L76).
Having said that, if @leeskelvin wants to add that in a block above (as in that link-to PR), then I could delete my PR and have it all done in one go.

mrawls: That is probably true, but I still think we should add it here! I forgot this schema was for the merged forcedSource catalog, and not for the one upstream a step or two before the merge happens, which doesn't care what flavor of exposure it is run on.

leeskelvin: I don't mind adding it here, and the 4 other schema files at the same time. Troubled by this though:
> I'm not sure adding it in this block will work.
If I add *nodataCenter into all 5 schemas, will something break?

leeskelvin:
It wouldn't be an @arunkannawadi review without at least _one_ request to split a commit into multiple commits :wink:. Commit split! Just waiting on a reply from @jrmullaney / @mrawls Re: `*nodataCenter` flag additions into the schemas.

schemas/ForcedSource.yaml
mrawls: That feels like a "be sure to run ci_*" flavored question 😅

When @jrmullaney said he wasn't sure it would work, I think he meant it might not let him do the downstream selection/filtering on merged forced source tables to the extent we want, not that it would break anything.

leeskelvin: I've added `base_PixelFlags_flag_nodataCenter` to the five schema YAMLs in this repo. Currently running Jenkins with:
```
lsst_distrib lsst_ci lsst_middleware lsst_sitcom ci_cpp ci_hsc ci_imsim ci_middleware
```

leeskelvin: After discussion with @jrmullaney, I've removed this column. The column needed is different than a straight-forward propagation, and requires more thought and testing. @jrmullaney will follow this up on [DM-49554](https://rubinobs.atlassian.net/browse/DM-49554).

PR #1058, opened 2025-03-17
DM-49423: LoadReferenceCatalogue to return np.nan as mag for non-detections, as opposed to 99

jrmullaney:


python/lsst/pipe/tasks/loadReferenceCatalog.py
```diff
@@ -279,7 +279,7 @@ def _formatCatalog(self, refCat, filterList):
                 # nan_to_num replaces nans with zeros, and this ensures
                 # that we select magnitudes that both filter out nans and are
                 # not very large (corresponding to very small fluxes), as "99"
```
erykoff, line 281: This comment above needs to be fixed (no longer 99).

erykoff (approved):
Looks good with a required comment change and of course jenkins + ci_hsc/ci_imsim

python/lsst/pipe/tasks/loadReferenceCatalog.py
jrmullaney: I did think about that, but the code below is still replacing the nans with 99s, but only to make the "good" mask. I could just get rid of the reference to 99, so as not to confuse things.

PR #1059, opened 2025-03-18
Change 'filters' to 'bands' for multiband images and exposures

fred3m:


python/lsst/pipe/tasks/measurementDriver.py
```diff
@@ -1092,15 +1092,15 @@ def _ensureValidInputs(
                 else:
                     measInfo = (
                         "while subtasks downstream of deblending will be run in each of "
-                        f"the {mExposure.filters} bands"
+                        f"the {mExposure.bands} bands"
```
arunkannawadi, line 1095: Not for this ticket, but noting that there's some unnecessary string formatting going on ahead of logging it.

arunkannawadi (approved):

PR #1071, opened 2025-03-21
DM-49074: Cast table columns to single precision in SDM-standardization functors

ctslater:


TallJimbo (approved):
I'm sad that the pragmatic way to do this is to explicitly cast a bunch of `_func` results (i.e. rather than declare a column type somewhere), but I also understand that that _is_ the pragmatic way to do this.

PR #1074, opened 2025-03-25
DM-49673: Add per-band spatialOrder config option in piff and add a minNPsfStar criterion for detector rejection

laurenam:


python/lsst/pipe/tasks/selectImages.py
```diff
@@ -283,6 +283,31 @@ class PsfWcsSelectImagesConfig(pipeBase.PipelineTaskConfig,
...
         optional=True,
     )
+    minNPsfStarPerBand = pexConfig.DictField(
+        keytype=str,
+        itemtype=float,
+        default={
+            "u": 10,
+            "g": 22,
+            "r": 22,
+            "i": 22,
```
PFLeget, line 293: Why 10/22? Is it to allow a threshold with the outlier rejection and based on what you observed with the weird PSF fit?

PFLeget (approved):

python/lsst/pipe/tasks/selectImages.py
laurenam: Yup...left some buffer for outlier rejection.  This _should_ be a no-op when DM-49688 is merged.

PR #1075, opened 2025-03-25
DM-49005 Changes and updates to the LSST RGB codebase

natelust:


fred3m (approved):
LGTM, contingent on our earlier discussion about backporting either the new API or a warning when importing from this package that the API is in development.

natelust:
I talked with Jim, and he said he was supportive of at minimum putting in the warning, or change the module path to have a _ or both as a back port, IF there is no appetite to back port all the changes. I will figure that out starting tomorrow after this is getting into main.

PR #1084, opened 2025-03-27
DM-49727: Add ComputeObjectEpochsTask

taranu:
This is a minimally altered version of ObjectEpochTableTask from analysis_tools (which I intend to deprecate shortly) but using `deepCoadd_obj` as an input instead of `objectTable`, so that it can more easily be merged into `objectTable`.

### DIFF
--- a/python/lsst/pipe/tasks/functors.py
+++ b/python/lsst/pipe/tasks/functors.py
@@ -755,8 +755,11 @@ class MultibandColumn(Column):
     def band_to_check(self):
         return self._band_to_check
 
+
+class MultibandSinglePrecisionFloatColumn(MultibandColumn):
+    """A float32 MultibandColumn"""
     def _func(self, df):
-        return df[self.col].astype(np.float32)
+        return super()._func(df).astype(np.float32)
 
 
 class SinglePrecisionFloatColumn(Column):
--- a/python/lsst/pipe/tasks/postprocess.py
+++ b/python/lsst/pipe/tasks/postprocess.py
@@ -1122,7 +1122,18 @@ class TransformObjectCatalogTask(TransformCatalogBaseTask):
             handle_multiband = handles_multi[dataset]
             df_dataset = handle_multiband.get()
             if isinstance(df_dataset, astropy.table.Table):
-                df_dataset = df_dataset.to_pandas().set_index(name_index, drop=False)
+                # Allow astropy table inputs to already have the output index
+                if name_index not in df_dataset.colnames:
+                    if self.config.primaryKey in df_dataset.colnames:
+                        name_index_ap = self.config.primaryKey
+                    else:
+                        raise RuntimeError(
+                            f"Neither of {name_index=} nor {self.config.primaryKey=} appear in"
+                            f" {df_dataset.colnames=} for {dataset=}"
+                        )
+                else:
+                    name_index_ap = name_index
+                df_dataset = df_dataset.to_pandas().set_index(name_index_ap, drop=False)
             elif isinstance(df_dataset, afwTable.SourceCatalog):
                 df_dataset = df_dataset.asAstropy().to_pandas().set_index(name_index, drop=False)
             # TODO: should funcDict have noDup funcs removed?
--- a/python/lsst/pipe/tasks/postprocess.py
+++ b/python/lsst/pipe/tasks/postprocess.py
@@ -912,6 +912,13 @@ class TransformObjectCatalogConnections(pipeBase.PipelineTaskConnections,
         name="{coaddName}Coadd_Sersic_multiprofit",
         deferLoad=True,
     )
+    inputCatalogEpoch = connectionTypes.Input(
+        doc="Catalog of mean epochs for each object per band.",
+        dimensions=("tract", "patch", "skymap"),
+        storageClass="ArrowAstropy",
+        name="object_epoch",
+        deferLoad=True,
+    )
     outputCatalog = connectionTypes.Output(
         doc="Per-Patch Object Table of columns transformed from the deepCoadd_obj table per the standard "
             "data model.",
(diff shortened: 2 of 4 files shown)


python/lsst/pipe/tasks/compute_object_epochs.py
```diff
@@ -0,0 +1,133 @@
...
+        multiple=True,
+    )
+
+
+class ComputeObjectEpochsConfig(
+    pipeBase.PipelineTaskConfig,
+    pipelineConnections=ComputeObjectEpochsConnections,
+):
+    bands = pexConfig.ListField(
+        doc=("Bands in objectCat to be combined with `objectCat_selectors` to build objectCat column names."),
```
TallJimbo, line 65: I don't see `objectCat_selectors` anywhere in this file.

python/lsst/pipe/tasks/compute_object_epochs.py
```diff
@@ -0,0 +1,133 @@
...
+
+from astropy.table import Table
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+import numpy as np
+
+
+class ComputeObjectEpochsConnections(
+    pipeBase.PipelineTaskConnections,
+    dimensions=("tract", "skymap"),
```
TallJimbo, line 30: I wonder if this should be `patch`, instead, given that it takes one patch-level input catalog and produced one patch-level output catalog, and the tract-level `epochMap` can't be that expensive to load once per patch.

python/lsst/pipe/tasks/compute_object_epochs.py
```diff
@@ -0,0 +1,133 @@
...
+        ----------
+        cat : `astropy.table.Table`
+            Catalog containing object positions.
+        epochMapDict: `dict` [`DeferredDatasetHandle`]
+            Dictionary of handles for healsparse maps containing the mean epoch
+            for positions in the reference catalog.
+
+        Returns
+        -------
+        epochDf = `astropy.table.Table`
```
TallJimbo, line 94: This may have originally been my fault, but `Df` is a strange suffix for something that is now an `astropy.table.Table`.

python/lsst/pipe/tasks/compute_object_epochs.py
```diff
@@ -0,0 +1,133 @@
...
+    def getEpochs(self, cat, epochMapDict):
+        """Get mean epoch of the visits corresponding to object position.
+
+        Parameters
+        ----------
+        cat : `astropy.table.Table`
+            Catalog containing object positions.
+        epochMapDict: `dict` [`DeferredDatasetHandle`]
+            Dictionary of handles for healsparse maps containing the mean epoch
+            for positions in the reference catalog.
```
TallJimbo, line 90: Should state that the keys are `str` band names, including in the docstring type.

python/lsst/pipe/tasks/compute_object_epochs.py
```diff
@@ -0,0 +1,133 @@
...
+
+        Returns
+        -------
+        epochDf = `astropy.table.Table`
+            Catalog with mean epoch of visits at each object position.
+        """
+        allEpochs = {}
+        for band in self.config.bands:
```
TallJimbo, line 98: Given how it's used, I think it's worth considering whether we should drop the `bands` config option.  Unlike other tasks that only take `band`-free inputs like the standardized object table, this task does take a property map with `band` in its data IDs, so there's not really any danger of that set of bands exploding to all of the bands in the data repository.  And it's always nice to have one fewer `bands` config to have to configure when setting up a new variant of a pipeline.

TallJimbo (approved):

python/lsst/pipe/tasks/compute_object_epochs.py
taranu: I'm not actually sure what the connections-level dimensions do so I defer to your expertise on that.

TallJimbo: It sets the dimensions of a single quantum, so the level of parallelization the task runs at.  I do think `patch` makes more sense, but that will require changes to other parts of the task; the input and output catalogs would be `multiple=False` and hence they wouldn't appear to `runQuantum` as lists anymore.

taranu: Oh, I see. I can't see any other reason it would've been written this way other than to avoid loading the epoch map for every patch. It seems almost certain to be I/O bound, though, so is there some other benefit to having deferred loading of a known set of patch tables vs running them all independently?

TallJimbo: Running them separately makes error- and missing-data handling a lot easier, since you don't have to worry about making sure one bad patch doesn't bring down the others.  It's also just a lot more familiar; usually when we have a task that does a gather-scatter like this, it's because there algorithmically needs to be a gather-scatter, and here that's just not true.

taranu: Alright. Is there any point in keeping either the `deferLoad` or `deferGraphConstraint` options, then?

python/lsst/pipe/tasks/compute_object_epochs.py
taranu: Hmm. Perhaps Clare intended to use analysis_tools selectors and put it in the package for that reason, but that doesn't seem to have actually happened.

python/lsst/pipe/tasks/compute_object_epochs.py
TallJimbo: I think you still want `deferLoad=True` to make it easier to load just a few columns.  I don't think you need `deferGraphConstraint`, but I'm not actually sure why that was there before, either.

TallJimbo (approved):

PR #1085, opened 2025-03-28
DM-49737 : Select associated diaSources based on diaSourceId

Gerenjie:


python/lsst/pipe/tasks/drpAssociationPipe.py
```diff
@@ -270,7 +270,9 @@ def visitDetectorPair(dataRef):
                                                    innerTractSkyRegion,
                                                    skyInfo.wcs)
                 associatedSsSources = associatedSsSources[ssInTractPatch]
-                associatedSsDiaSources = associatedSsDiaSources[ssInTractPatch]
+                assocDiaSrcIds = set(associatedSsSources['diaSourceId'])
+                diaSrcMask = [ID in assocDiaSrcIds for ID in associatedSsDiaSources['diaSourceId']]
```
enourbakhsh, line 274: Nitpick: `ID` feels like a constant, and `id` is a built-in function. I'd go with something like `src_id` or `diaId` etc.

python/lsst/pipe/tasks/drpAssociationPipe.py
```diff
@@ -270,7 +270,9 @@ def visitDetectorPair(dataRef):
...
+                associatedSsDiaSources = associatedSsDiaSources[np.array(diaSrcMask)]
```
enourbakhsh, line 275: Could this ever return empty? If so, is that okay, or is it handled somewhere else?

enourbakhsh (approved):
Looks good! I just have a couple of suggestions/questions.

PR #1086, opened 2025-03-28
DM-49576: Use nsecs directly rather than going via python datetime

timj:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1524,7 +1524,7 @@ def run(self, visitSummaryRefs):
                 visitSummary["psfSigma"] * visitSummary["pixelScale"] * np.sqrt(8 * np.log(2))
             )
             ccdEntry["skyRotation"] = visitInfo.getBoresightRotAngle().asDegrees()
-            ccdEntry["expMidpt"] = np.datetime64(visitInfo.getDate().toPython(), "ns")
+            ccdEntry["expMidpt"] = np.datetime64(visitInfo.date.nsecs(scale=dafBase.DateTime.UTC), "ns")
```
timj, line 1527: @ktlim I assume we are allowed to assume that the dafBase.DateTime epoch (which is defined in the C++) is the same as np.datetime64 epoch (defined in the numpy docstring) without having to explicitly ask for a DateTime with nsecs==0 and calculate the offset from the numpy reference epoch.

timj:
```python
>>> np.datetime64(0, "ns")
np.datetime64('1970-01-01T00:00:00.000000000')
>>> DateTime(nsecs=0).toString(DateTime.UTC)
'1969-12-31T23:59:51.999918240Z'
```
and of course they are technically different...

timj: On the other hand the code before was likely broken:
```
>>> x = DateTime(nsecs=0, scale=DateTime.UTC)
>>> x
DateTime("1970-01-01T00:00:08.000082000", TAI)
>>> import numpy as np
>>> np.datetime64(x.toPython(), "ns")
np.datetime64('1970-01-01T00:00:08.000082000')
>>> np.datetime64(x.toPython(timescale=DateTime.UTC), "ns")
np.datetime64('1970-01-01T00:00:00.000000000')
>>> np.datetime64(x.nsecs(scale=DateTime.UTC), "ns")
np.datetime64('1970-01-01T00:00:00.000000000')
```
and would have needed the `timescale` parameter.

python/lsst/pipe/tasks/postprocess.py
```diff
@@ -1524,7 +1524,7 @@ def run(self, visitSummaryRefs):
                 visitSummary["psfSigma"] * visitSummary["pixelScale"] * np.sqrt(8 * np.log(2))
             )
             ccdEntry["skyRotation"] = visitInfo.getBoresightRotAngle().asDegrees()
-            ccdEntry["expMidpt"] = np.datetime64(visitInfo.getDate().toPython(), "ns")
+            ccdEntry["expMidpt"] = np.datetime64(visitInfo.date.nsecs(scale=dafBase.DateTime.TAI), "ns")
```
ktlim, line 1527: So we have Python `datetime`, Numpy `datetime64`, Astropy `Time`, and our own `lsst.daf.base.DateTime`?  Ugh.

ktlim (approved):

PR #1089, opened 2025-04-01
DM-49787: write standardized ForcedSource tables as ArrowAstropy, and don't write empty ones

TallJimbo:


erykoff (approved):
I wish we could replace that `outputCatalog = pd.concat(dfs)` because it's an insane memory hog, but that would require refactoring all the transformation functors I fear.

TallJimbo:
I don't think we'd need to refactor the functors (we already transform before we concatenate, after all), but I was only willing to do on this ticket what I was willing to backport to v29, and doing the conversion any earlier seemed too risky for that.

PR #1090, opened 2025-04-01
DM-47711 Add lsstRGB image format to hips code generation (and make it somewhat parallel)

natelust:
This change set 
 * adds lsstRGB image format to hips image generation as an option
 * adds the ability to output in webp image format
 * adds the option to run hips generation in a somewhat parallel way.

### DIFF
--- a/python/lsst/pipe/tasks/hips.py
+++ b/python/lsst/pipe/tasks/hips.py
@@ -1647,14 +1647,17 @@ class GenerateColorHipsConfig(GenerateHipsConfig, pipelineConnections=GenerateCo
     )
 
     def setDefaults(self):
+        super().setDefaults()
         self.rgbGenerator: PrettyPictureConfig
-        self.rgbGenerator.imageRemappingConfig.absMax = 1700
+        self.rgbGenerator.imageRemappingConfig.absMax = 300
         # self.rgbGenerator.imageRemappingConfig.quant = 1
         self.rgbGenerator.luminanceConfig.Q = 0.7
         self.rgbGenerator.doPSFDeconcovlve = False
         self.rgbGenerator.exposureBrackets = [0.9, 1, 3]
+        self.rgbGenerator.exposureBrackets = None
         self.rgbGenerator.luminanceConfig.stretch = 220
-        return super().setDefaults()
+        self.rgbGenerator.localContrastConfig.doLocalContrast = False
+        return
 
 
 class GenerateColorHipsTask(GenerateHipsTask):
--- a/python/lsst/pipe/tasks/hips.py
+++ b/python/lsst/pipe/tasks/hips.py
@@ -1003,7 +1003,20 @@ class GenerateHipsTask(pipeBase.PipelineTask):
                 Q=self.config.png_gray_asinh_softening,
             )
         else:
-            colorstr = "".join(bands)
+            match self.config.rgbStyle:
+                case "lupton":
+                    png_color_mapping = AsinhMapping(
+                        self.config.png_color_asinh_minimum,
+                        self.config.png_color_asinh_stretch,
+                        Q=self.config.png_color_asinh_softening,
+                    )
+
+                    bcb = self.config.blue_channel_band
+                    gcb = self.config.green_channel_band
+                    rcb = self.config.red_channel_band
+                    colorstr = f"{bcb}{gcb}{rcb}"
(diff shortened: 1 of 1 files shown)


python/lsst/pipe/tasks/hips.py
```diff
@@ -948,16 +950,7 @@ def run(self, bands, max_order, hips_exposure_handle_dict, do_color=False):
...
-                self.config.png_color_asinh_minimum,
-                self.config.png_color_asinh_stretch,
-                Q=self.config.png_color_asinh_softening,
-            )
-
-            bcb = self.config.blue_channel_band
-            gcb = self.config.green_channel_band
-            rcb = self.config.red_channel_band
-            colorstr = f"{bcb}{gcb}{rcb}"
+            colorstr = "".join(bands)
```
TallJimbo: It is possible for bands to be more than one color (e.g. the narrow-band filters for HSC).  Would this cause a problem in those cases?

python/lsst/pipe/tasks/hips.py
```diff
@@ -1619,33 +1592,70 @@ class GenerateColorHipsConfig(GenerateHipsConfig, pipelineConnections=GenerateCo
         doc="Band to use for blue channel of color pngs.",
         dtype=str,
         default="g",
+        deprecated=(
+            "This field is deprecated, set options in the rgbGenerator sub-config"
+            "This will be removed in version 29"
+        ),
```
TallJimbo: I think these need to say "removed after version 29", right?  Since you're planning to backport the deprecation but not the removal to v29?

Also, if these deprecated fields are now ignored, please include that in the deprecation messages.

python/lsst/pipe/tasks/hips.py
```diff
@@ -860,6 +871,46 @@ class GenerateHipsConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Genera
...
         default=8.0,
     )
+    parallel_highest_order = pexConfig.Field[bool](
+        doc=(
+            "If this is set to True, each of the highest order hips pixels will"
+            " be put into their own quanta, and can be run in parallel. The "
+            "trade off is the highest order must be run alone by setting "
+            "min_order to be the the same as the healpix dimension. This will "
+            "skip writing all sky info, which must be done in another"
+            " invocation of this task."
```
TallJimbo, line 881: Please make this consistent about putting line-break spaces at either the end or the beginning.

python/lsst/pipe/tasks/hips.py
```diff
@@ -860,6 +871,46 @@ class GenerateHipsConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Genera
...
+        ),
+        default=False,
+    )
+    skip_highest_image = pexConfig.Field[bool](
+        doc=(
+            "This option should be used if in another task instance"
+            "parallel_highest_order was set to True. This option will skip"
+            "making and writing png for the highest order."
```
TallJimbo, line 889: This multi-line string doesn't have spaces at the line breaks.

python/lsst/pipe/tasks/hips.py
```diff
@@ -936,16 +1003,20 @@ def run(self, bands, max_order, hips_exposure_handle_dict, do_color=False):
...
+                    )
 
-            bcb = self.config.blue_channel_band
-            gcb = self.config.green_channel_band
-            rcb = self.config.red_channel_band
-            colorstr = f"{bcb}{gcb}{rcb}"
+                    bcb = self.config.blue_channel_band
+                    gcb = self.config.green_channel_band
+                    rcb = self.config.red_channel_band
+                    colorstr = f"{bcb}{gcb}{rcb}"
```
TallJimbo, line 1017: I'm guessing this is a lot of work, but it would be really nice if the `lupton` case could just be a different `rgbGenerator` implementation, with that `ConfigurableField` maybe replaced with a `RegistryField`.

Is the goal to deprecate the `lupton` version shortly?  If not, do you need to un-deprecate any of its config parameters?

TallJimbo (approved):

python/lsst/pipe/tasks/hips.py
natelust: Yeah, the goal is to keep Lipton only in the short term as we transition to the new framework, so it is preserved mostly as is.

PR #1110, opened 2025-04-21
DM-48367: Add DiffMatchedTractCatalog prefix_best_coord option

taranu:
This makes it possible to add a "best available" RA/dec column when combining reference catalogs by matching them before matching with the target catalog.


psferguson (approved):
This looks good to me

PR #1111, opened 2025-04-21
DM-45569: Add color from fgcm for fitting a PSF color term in Piff.

PFLeget:


python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -434,6 +465,18 @@ def _make_selection_schema_mapper(self, input_schema):
 
         selection_schema.setAliasMap(input_schema.getAliasMap())
 
+        selection_schema.addField(
+            'psfColorValue',
+            type=np.float32,
```
arunkannawadi, line 470: Would `np.float16` be sufficient here and above?

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -622,6 +689,12 @@ def compute_psf_and_ap_corr_map(self, visit, detector, exposure, src, isolated_s
         # We need to copy over the calib_psf flags because they were not in the mapper
         measured_src['calib_psf_candidate'] = selected_src['calib_psf_candidate']
         measured_src['calib_psf_reserved'] = selected_src['calib_psf_reserved']
+        if hasattr(exposure.filter, "bandLabel"):
```
arunkannawadi, line 692:
```suggestion
        if exposure.filter.hasBandLabel():
```

PR #1112, opened 2025-04-22
DM-50431: Pass Exposure metadata from multiBand to propagateSourceFlags

arunkannawadi:
I have confirmed that this is a no-op for current coadd+measurements by running on one arbitrary patch.

### DIFF
--- a/python/lsst/pipe/tasks/multiBand.py
+++ b/python/lsst/pipe/tasks/multiBand.py
@@ -178,9 +178,13 @@ class DetectCoaddSourcesTask(PipelineTask):
     def runQuantum(self, butlerQC, inputRefs, outputRefs):
         inputs = butlerQC.get(inputRefs)
         idGenerator = self.config.idGenerator.apply(butlerQC.quantum.dataId)
-        inputs["idFactory"] = idGenerator.make_table_id_factory()
-        inputs["expId"] = idGenerator.catalog_id
-        outputs = self.run(**inputs)
+        exposure = inputs.pop("exposure")
+        assert not inputs, "runQuantum got more inputs than expected."
+        outputs = self.run(
+            exposure=exposure,
+            idFactory=idGenerator.make_table_id_factory(),
+            expId=idGenerator.catalog_id,
+        )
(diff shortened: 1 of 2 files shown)


python/lsst/pipe/tasks/multiBand.py
```diff
@@ -294,6 +300,14 @@ class MeasureMergedCoaddSourcesConnections(
...
         deferLoad=True,
     )
+    finalVisitSummaryHandles = cT.Input(
+        doc="Final visit summary table",
+        name="finalVisitSummary",
+        storageClass="ExposureCatalog",
+        dimensions=("instrument", "visit"),
+        multiple=True,
+        deferLoad=True,
+    )
```
TallJimbo, line 310: This is now needed because we're not putting the per-{visit, detector} WCSs into the cell-based coadds?

I think it'd be a lot cleaner for `PropagateSourceFlags` to use the existing {ra, dec} columns and avoid needing a {visit, detector} WCS at all, but that might look more like a complete rewrite of the class and I don't want that to hold up cell-based coadds integration.  It might actually be best to move propagating the source flags out of this task and into one of the postprocessing tasks, especially since I think we want to stand up more general object-source matching.  I don't want any of that musing to hold up this work, but I think I should go make some tickets, and I figure it might be useful for you to get a sense for where things might be headed in your current work.

TallJimbo (approved):

python/lsst/pipe/tasks/multiBand.py
arunkannawadi: > This is now needed because we're not putting the per-{visit, detector} WCSs into the cell-based coadds?

Yes. And I think that was always the plan (see [here](https://github.com/lsst/cell_coadds/blob/7e2847429e6b122cd7f2202d1519ecc57b28bca7/python/lsst/cell_coadds/_stitched_coadd.py#L176))

> but that might look more like a complete rewrite of the class

It might be less than a complete rewrite, but I agree about that not holding up this. I'm happy to create those tickets, but if you do, please add me as a Watcher.

PR #1114, opened 2025-04-23
DM-45990: Create initial_pvi background metric

enourbakhsh:


mrawls (approved):
Please rebase into a single commit. I suggest you add a line to the `update_background_stats` docstring that makes it clear this method adds two entries to `summary`: `skyBg` and `skyBgNormRange`, and (bonus level) a one-sentence description of what each one is. Otherwise it feels like stealth undocumented things that would warrant a "Returns" docstring if it weren't just updating `summary` in place.

tests/test_computeExposureSummaryStats.py
```diff
@@ -106,12 +106,20 @@ def testComputeExposureSummary(self):
...
+        approxStyle = afwMath.ApproximateControl.UNKNOWN
+        approxOrderX = 0
+        approxOrderY = 0
+        approxWeighting = False
         backobj = afwMath.makeBackground(exposure.getMaskedImage().getImage(), bctrl)
         background = afwMath.BackgroundList()
-        background.append(backobj)
+        background.append(
+            (backobj, interpStyle, undersampleStyle, approxStyle, approxOrderX, approxOrderY, approxWeighting)
+        )
```
enourbakhsh, line 122: Needed to update this to support the `background.getImage()` call in the task.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -22,21 +22,22 @@
...
+import lsst.pipe.base as pipeBase
+import numpy as np
+from astropy.coordinates import AltAz, EarthLocation, SkyCoord
+from astropy.time import Time
+from lsst.daf.base import DateTime
 from lsst.meas.algorithms import ScienceSourceSelectorTask
 from lsst.utils.timer import timeMethod
-import lsst.ip.isr as ipIsr
+from scipy.stats import median_abs_deviation as sigmaMad
+from skimage.util import view_as_windows
```
enourbakhsh, line 40: Simply organized the imports here.

PR #1119, opened 2025-04-30
DM-50589: Add error for bad centroiding during PSF fit

parejkoj:
The test was carefully constructed to successfully measure a (garbage) PSF but then have centroiding fail. This can happen under real conditions, but it's quite sensitive to the nature of the PSF badness and how that interacts with CR repair.

### DIFF
--- a/tests/test_calibrateImage.py
+++ b/tests/test_calibrateImage.py
@@ -42,7 +42,8 @@ import lsst.meas.extensions.psfex
 import lsst.meas.base
 import lsst.meas.base.tests
 import lsst.pipe.base.testUtils
-from lsst.pipe.tasks.calibrateImage import CalibrateImageTask, NoPsfStarsToStarsMatchError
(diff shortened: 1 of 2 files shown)


tests/test_calibrateImage.py
```diff
@@ -306,6 +307,41 @@ def test_compute_psf(self):
         # display = lsst.afw.display.getDisplay()
         # display.mtv(self.exposure)
 
+    def test_compute_psf_bad_centroids(self):
+        """Test that we raise an appropriate error if all measured centroids
+        are bad in compute_psf.
+        The root cause of this is likely an interaction between the PSF fit
+        and cosmic ray repair. an ugly true PSF can result in the centers of
```
erykoff, line 314: an -> An

erykoff (approved):
Minor typo in a docstring; otherwise looks good!

PR #1121, opened 2025-05-02
DM-49008: Add options to use cell-based coadds

arunkannawadi:


python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
```diff
@@ -161,9 +175,20 @@ def __init__(self, *, config=None):
         del self.fluxCatalogs
         del self.templateCatalogs
 
+        if config:
+            if config.useCellCoadds:
+                del self.coadds
+            else:
+                del self.coadds_cell
+                del self.backgrounds
```
TallJimbo, line 183: Given that you also need to dispatch on which of these connections to use in `runQuantum`, it may be better to use a single connection and let the config option change the storage class, via something like:
```
self.coadds = dataclasses.replace(self.coadds, storageClass="MutipleCellCoadd")
```
That does mean that you'd have to set the connection name in addition to setting `useCellCoadds=True` to when configuring a pipeline to use cell-based coadds, but I think it'd simplify `runQuantum` implementations and reduce the number of connections we'll have to deprecate later.

python/lsst/pipe/tasks/fit_coadd_multiband.py
```diff
@@ -198,9 +212,19 @@ def adjustQuantum(self, inputs, outputs, label, data_id):
         return adjusted_inputs, {}
 
     def __init__(self, *, config=None):
+        super().__init__(config=config)
+        if config is None:
+            return
```
TallJimbo, line 217: FWIW I've gotten the habit just doing
```
assert isinstance(config, DerivedConfigClass)`
```
here; I think the `=None` in the base class is just unnecessary, but I haven't ever gotten around to fixing it.

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -604,11 +642,23 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+        if self.config.useCellCoadds:
+            multiple_cell_coadd = inputs.pop("exposure_cells")
+            stitched_coadd = multiple_cell_coadd.stitch()
+            exposure = stitched_coadd.asExposure()
+            background = inputs.pop("background")
+            exposure.image -= background.getImage()
+
+            ccdInputs = stitched_coadd.ccds
+            apCorrMap = stitched_coadd.ap_corr_map
+            band = inputRefs.exposure_cells.dataId["band"]
```
TallJimbo, line 654: This block is common enough to merit being put into a function somewhere, even if the names of the connections need to be parameters.

TallJimbo (approved):

python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
arunkannawadi: This was the approach we had in an earlier iteration (during a live session) before switching to having the current setup of having two connections. It does make switching slightly less trivial considering the proliferating number of pipelines for LSST*Cam*

arunkannawadi: After some further thoughts, I'm afraid I have to push back on this suggestion. Doing this would defeat the idea of having the transition seamless with a simple toggle and more over will leave the door wide open for misconfiguration when one pipeline imports from another. Excluding `assembleCellCoadd` by hand introduces some friction to the switching, but that's something I am willing to tolerate.

PR #1122, opened 2025-05-05
DM-50698: Raise UpstreamFailureNoWorkFound when there are no rows to transform.

erykoff:


python/lsst/pipe/tasks/postprocess.py
```diff
@@ -850,6 +850,10 @@ def run(self, handle, funcs=None, dataId=None, band=None):
 
         df = self.transform(band, handle, funcs, dataId).df
         self.log.info("Made a table of %d columns and %d rows", len(df.columns), len(df))
+
+        if len(df) == 0:
+            raise UpstreamFailureNoWorkFound("No rows were created in upstream task to transform.")
```
isullivan, line 855: The error message is not very clear. Perhaps something like:
`"Input catalog is empty, so there is nothing to transform/standardize"`

isullivan (approved):
Looks good. A minor suggestion on wording for the error message.

PR #1123, opened 2025-05-06
DM-50695: Raise MatcherError if no reference sources have finite flux

isullivan:
The resulting error looks like:
```
lsst.meas.astrom.exceptions.MatcherFailure: No matched reference stars with valid fluxes: {'goodSrc': 369, 'goodRef': 0}
```


erykoff (approved):
Good point that MatcherFailure looks appropriate and consistent with usage elsewhere in this file.

PR #1124, opened 2025-05-06
DM-50729: Catch AlgorithmError in DetectCoaddSources

fred3m:


python/lsst/pipe/tasks/multiBand.py
```diff
@@ -180,11 +187,23 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+            )
+        except TooManyMaskedPixelsError as e:
+            error = AnnotatedPartialOutputsError(
+                e,
+                self,
+                exposure,
+                log=self.log,
+            )
+            result = Struct(outputSources=None, outputBackgrounds=None, outputExposure=None)
+            butlerQC.put(result, outputRefs)
```
TallJimbo, line 204: Since you're not actually writing anything, you can drop this line and the above it.

TallJimbo (approved):

PR #1125, opened 2025-05-07
DM-50755: Refactor SkyCorrectionTask to optimize memory usage

leeskelvin:


python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -339,67 +345,194 @@ def run(self, calExps, calBkgs, skyFrames, camera, backgroundToPhotometricRatioH
...
+        -------
+        calExp : `lsst.afw.image.ExposureF`
+            The calexp with the mask and skyCorr applied.
+        """
+        if isinstance(calExpHandle, DeferredDatasetHandle):
+            calExp: ExposureF = calExpHandle.get()
+        else:
+            # Here we clone the imaging data to avoid modifying data which is
+            # used in downstream processing.
+            calExp: ExposureF = calExpHandle.clone()
```
arunkannawadi, line 485: I don't see `from __future__ import annotations`. Does this type hint work without that?

leeskelvin: Thank you! Added.

PR #1126, opened 2025-05-09
DM-50804: Allow skyCorr to process incomplete visits

leeskelvin:


python/lsst/pipe/tasks/skyCorrection.py
```diff
@@ -366,6 +367,7 @@ def run(self, calExps, calBkgs, skyFrames, camera, backgroundToPhotometricRatioH
             _ = self._restoreOriginalBackgroundRefineMask(calExp, calBkg)
             masks.append(calExp.mask)
             skyCorrs.append(calBkg)  # Contains only the inverted original background elements at this stage
+            bgModel1Indices.append(len(calBkg._backgrounds))  # Index of the original background element
```
arunkannawadi, line 370: Anyway you can do this without using the attribute with a leading underscore?

arunkannawadi (approved):
So do I understand that you just want the `zip` to end earlier if there are fewer than expected components?

python/lsst/pipe/tasks/skyCorrection.py
arunkannawadi: It looks like `len(calBkg)` should just work.

leeskelvin: Yes, I think you're right, happy to switch to just `len(calBkg)`

leeskelvin:
We need to know the index of the `bgModel1` component for a couple of purposes. 1) to remove it if `undoBgModel1` is `True`, and 2) to construct a mosaic of only the new background list elements that come after the original elements index.

PR #1127, opened 2025-05-13
DM-50824: Add option to exclude detectors

cmsaunders:


fred3m (approved):

yalsayyad (changes requested):
Sorry about the fly-by. But we don't want to exclude detectors from the calculations on which visits to include in the coadds. We wanted to exclude detectors from being included in the "direct_warps" via the SelectImages family: `BaseSelectImagesTask`-`WcsSelectImagesTask` - `PsfWcsSelectImagesTask`

cmsaunders:
Ok, I see what you are saying. I'll get that implemented. 

Is it still worthwhile keeping the configs in the SelectVisits tasks so that the numbers are consistent with what is in the direct_warps, or should I just take that out?

yalsayyad (approved):
Looks good. Thank you!  As for the `selectDeepCoaddVisits.excludeDetectors`, I'm worried the name is confusing. I just ran it by @natelust who was sitting next to me at standup and he couldn't guess what that would do. Also that task uses a median statistic so that I doubt it would move the needle that much on the median PSF FWHM which is why I'm so ambivalent towards the idea...

cmsaunders:
Ok, I'll remove the first commit.

PR #1130, opened 2025-05-20
DM-51026: Add config to CalibrateImageTask to downsample footprints in dense regions.

erykoff:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -1101,6 +1111,33 @@ def _find_stars(self, exposure, background, id_generator, background_to_photomet
...
+            indices = np.arange(len(sources))[~sources["sky_source"]]
+            indices = gen.choice(
+                indices,
+                size=self.config.downsample_max_footprints,
+                replace=False,
+            )
+            skyIndices, = np.where(sources["sky_source"])
+            indices = np.concatenate((indices, skyIndices))
+
+            self.log.info("Downsampling from %d to %d footprints.", len(sources), len(indices))
```
mfisherlevine, line 1135: I think in the log message, the code and the config docstring it'd be nice to just say in words what the intention is wrt the sky sources so that future maintainers know what _should_ be being done, and therefore just add the sky source count to the log message so it's clear.

mfisherlevine (approved):
Apart from the one comment it looks great 👍

PR #1131, opened 2025-05-21
DM-51028: Add option to allow excluding astrometry errors for speed

mfisherlevine:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -368,6 +368,12 @@ class CalibrateImageConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Cali
         )
     )
 
+    do_include_astrometry_errors = pexConfig.Field(
```
erykoff, line 371: Probably `do_include_astrometric_errors`.  Just look at your docstring!

erykoff (commented):
Can you add a test?  😄

PR #1135, opened 2025-06-02
DM-48896: Create pipeline yaml files for DRP with ComCam 

BrunoSanchez:
We migrated legacy matching to new matching tasks for diffim fake source injection.
Now the matching is compatible with `source_injection` catalogs.


python/lsst/pipe/tasks/matchDiffimSourceInjected.py
```diff
@@ -0,0 +1,432 @@
...
+
+        Parameters
+        ----------
+        injectedCat : `astropy.table.table.Table`
+            Table of catalog of synthetic sources to match to detected diaSources.
+        diffIm : `lsst.afw.image.Exposure`
+            Difference image where ``diaSources`` were detected.
+        diaSources : `afw.table.SourceCatalog`
+            Catalog of difference image sources detected in ``diffIm``.
+        assocDiaSources : `pandas.DataFrame`
```
leeskelvin, line 125: I think this is vestigial.

leeskelvin (approved):
LGTM, but I haven't tested this at all. I assume you have, and all works well (please verify on Jira).

tests/test_matchSourceInjected.py
```diff
@@ -0,0 +1,188 @@
+#
+# This file is part of ap_pipe.
```
BrunoSanchez, line 2: change to `pipe_tasks`

PR #1137, opened 2025-06-09
DM-46523: Remove pandas support in tract matching

taranu:


python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -35,10 +35,8 @@
 import astropy.table
 import astropy.units as u
 import numpy as np
-import pandas as pd
 from smatch.matcher import sphdist
 from typing import Sequence
-import warnings
```
nevencaplar, line 41: We can just remove warnings? Because you dont expect any warning anymore?

taranu: Yes, it was only being used to issue the deprecation warnings which are all gone now. And in fact the import has to be removed or else it'll give unused import errors.

nevencaplar (approved):

PR #1138, opened 2025-06-10
DM-51156: Add max pixel value when psf candidate is done.

PFLeget:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -629,14 +634,18 @@ def __init__(self, initial_stars_schema=None, **kwargs):
                            "calib_astrometry_used",
                            # TODO DM-39203: these can be removed once apcorr is gone.
                            "apcorr_slot_CalibFlux_used", "apcorr_base_GaussianFlux_used",
-                           "apcorr_base_PsfFlux_used")
+                           "apcorr_base_PsfFlux_used",)
         for field in self.psf_fields:
             item = self.psf_schema.find(field)
             initial_stars_schema.addField(item.getField())
         id_type = self.psf_schema["id"].asField().getTypeString()
+        psfmaxvaluetype = self.psf_schema['psf_max_value'].asField().getTypeString()
```
arunkannawadi, line 642:
```suggestion
        psf_max_value_type = self.psf_schema['psf_max_value'].asField().getTypeString()
```

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -477,6 +483,12 @@ def _make_selection_schema_mapper(self, input_schema):
             size=10,
             doc="Color used in PSF fit."
         )
+        selection_schema.addField(
+            'psf_max_value',
+            type=np.float32,
+            doc="PSF max value.",
```
arunkannawadi, line 489: Could you please update these docs to be not just the fieldname? Something like "Peak value of the PSF model" or "Maximum value in the PSF/star image" for example, as appropriate?

python/lsst/pipe/tasks/calibrateImage.py
arunkannawadi: Also, why is this the only place to infer the value type when in other cases they are always set to `float32`?

PR #1139, opened 2025-06-17
DM-51430: Limit skyCorr to processing common detectors only

leeskelvin:


czwa (approved):
A comment explaining this might be useful for the future.

leeskelvin:
Thanks Chris. Modified the comment for this code block to:
```
        # Sort input/output connections by detector ID, padding where
        # necessary, to ensure that all detectors are processed consistently.
        # Detector IDs are defined from the intersection of calExps, calBkgs,
        # and optionally skyFrames and backgroundToPhotometricRatioHandles.
        # This resolves potential missing data issues when processing a visit
        # that contains only partial inputs.
```

PR #1141, opened 2025-07-08
DM-50987: Add sattle to pipelines

bsmartradio:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -384,6 +387,35 @@ class CalibrateImageConfig(pipeBase.PipelineTaskConfig, pipelineConnections=Cali
...
     )
 
+    run_sattle = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="If True, sattle service will populate catalog for use in "
+            "ip_diffim.detectAndMeasure alert verification."
+    )
+
+    sattle_historical = sattle_host = pexConfig.Field(
```
kfindeisen, line 397: I assume this is a typo:
```suggestion
    sattle_historical = pexConfig.Field(
```

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -901,6 +933,34 @@ def run(
             result.applied_photo_calib = photo_calib
         else:
             result.applied_photo_calib = None
+
+        if self.config.run_sattle:
+            if not self.config.sattle_host or not self.config.sattle_port:
+                raise RuntimeError("Sattle filtering is on but service endpoints not set.")
```
kfindeisen, line 939: Please check this in `CalibrateImageConfig.validate`.

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -901,6 +933,34 @@ def run(
...
+                                                    "exposure_start_mjd": exposure_start_jd,
+                                                    "exposure_end_mjd": exposure_end_jd,
+                                                    "boresight_ra": boresight_ra,
+                                                    "boresight_dec": boresight_dec,
+                                                    "historical": self.config.sattle_historical})
+
+            if r.status_code != 200:
+                self.log.warning(f'Sattle cache returned {r.status_code}: {r.text}')
+            else:
+                self.log.info('Successfully loaded sattle visit cache')
```
kfindeisen, line 962: This duplicates code with lsst/ip_diffim#418. Unfortunately, I'm not sure where to put it, especially since IIRC `pipe_tasks` depends on `ip_diffim`. Maybe @ktlim has an idea?

tests/test_calibrateImage.py
```diff
@@ -584,6 +586,65 @@ def test_calibrate_image_illumcorr(self):
         self.assertIn(key, result.exposure.metadata)
         self.assertEqual(result.exposure.metadata[key], True)
 
+    @staticmethod
+    def _mocked_requests_put(url, json=None):
```
kfindeisen, line 590: Same comments as in `ip_diffim`. If the Sattle access code is factored out, that should also eliminate most of the test redundancy.

kfindeisen (changes requested):
The most glaring issue is the code duplication with lsst/ip_diffim#418, but I also have one big-picture question: what, exactly, is this code for? It doesn't seem to use the Sattle results for anything. Can it all just be removed?

kfindeisen:
> The big picture is that sattle needs the exact exposure time and the boresight ra & dec to populate a cache that will later be used in ip_diffim.detectAndMeasure. That cache takes a small amount of time to populate, so we want to call it early in the pipeline execution process so we're not waiting on it later during detectAndMeasure.

I think a comment to that effect would be helpful (though, again, factoring the code might make that unnecessary).

kfindeisen (approved):
Looks good. As before, many of my comments from lsst/ip_diffim#418 apply here as well.

tests/test_calibrateImage.py
```diff
@@ -584,6 +588,39 @@ def test_calibrate_image_illumcorr(self):
...
+        calibrate = CalibrateImageTask(config=self.config)
+        calibrate.astrometry.setRefObjLoader(self.ref_loader)
+        calibrate.photometry.match.setRefObjLoader(self.ref_loader)
+        calibrate.run(exposures=self.exposure)
+
+    @mock.patch.dict(os.environ, {"SATTLE_URI_BASE": "fake_host:1234"})
+    @mock.patch('requests.put')
+    def test_sattle(self, mock_put):
+        """Test that run() returns reasonable values to be butler put.
+        """
```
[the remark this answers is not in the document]
ebellm: leftover docstring, updated.

PR #1145, opened 2025-07-16
DM-48936: Implement mock-based unit tests for mid-level measurement drivers

enourbakhsh:


tests/test_measurementDriver.py
```diff
@@ -0,0 +1,418 @@
...
+        # Create a minimal schema to house the sources.
+        schema = TestDataset.makeMinimalSchema()
+
+        # Create a bounding box for the images.
+        bbox = lsst.geom.Box2I(lsst.geom.Point2I(0, 0), lsst.geom.Extent2I(self.nx, self.ny))
+
+        # Store datasets per band.
+        self.datasets = {}
+
+        rng = np.random.RandomState(565)
```
taranu, line 342: This is a legacy function and shouldn't be used. Use e.g. `numpy.random.default_rng(seed=565)` instead.

I wonder if this is the cause of your semi-random result above? It shouldn't be, but two different results suggests a test order dependency.

taranu (commented):
Looks good. See comments/suggestions. If you don't figure out what's causing the one alternating result, please file another ticket for it.

PR #1149, opened 2025-07-22
DM-51903: Update Object.yaml to support Merian DECam bands

leeskelvin:


taranu (approved):
Hopefully this won't remain here long...

PR #1162, opened 2025-09-24
DM-52593: Evaluate coadd PSF at source centroid

taranu:


python/lsst/pipe/tasks/fit_coadd_psf.py
```diff
@@ -41,8 +42,7 @@ def get_catalog(self):
 
     def get_psf_image(self, source):
         """Return the PSF image for this object."""
-        bbox = source.getFootprint().getBBox()
-        center = bbox.getCenter()
+        center = Point2D(round(source.getX()), round(source.getY()))
```
leeskelvin, line 45: I'm sure there's a good reason, but why do you need to round x and y here? Point2D will accept floats I think?

taranu: I don't know if it's actually necessary but it keeps the position as the center of a pixel, which the bbox center was half the time (the other half being an edge). Might as well have it consistent, I think.

PR #1226, opened 2026-01-15
DM-53032: Add BrightStarStackTask to generate extended PSF models

bazkiaei:


PFLeget (approved):
I looked at the code, did a double check with Claude, and it looks good to me !

PR #1263, opened 2026-02-20
DM-54194: Add second moment in sky /AltAz and fgcm phot for PSF single visit table

PFLeget:
Adding optional information to write in PSF table for analysis.

### DIFF
--- a/python/lsst/pipe/tasks/finalizeCharacterization.py
+++ b/python/lsst/pipe/tasks/finalizeCharacterization.py
@@ -40,11 +40,13 @@ import numpy as np
 import esutil
 from smatch.matcher import Matcher
 
-
+import lsst.geom as geom
+from lsst.geom import LinearTransform
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 import lsst.daf.base as dafBase
 import lsst.afw.table as afwTable
+from lsst.afw.geom import Quadrupole
 import lsst.meas.algorithms as measAlg
 import lsst.meas.extensions.piff.piffPsfDeterminer  # noqa: F401
 from lsst.meas.algorithms import MeasureApCorrTask
@@ -97,7 +99,10 @@ class FinalizeCharacterizationConnectionsBase(
         super().__init__(config=config)
         if config is None:
             return None
-        if not self.config.psf_determiner['piff'].useColor:
+        # Keep fgcm_standard_star if using chromatic PSF OR if adding FGCM photometry
+        needs_fgcm = (config.psf_determiner['piff'].useColor
+                      or config.do_add_fgcm_photometry)
+        if not needs_fgcm:
             self.inputs.remove("fgcm_standard_star")
 
 
@@ -208,6 +213,21 @@ class FinalizeCharacterizationConfigBase(
     apply_ap_corr = pexConfig.ConfigurableField(
         target=ApplyApCorrTask,
         doc="Subtask to apply aperture corrections"
+    )
+    do_add_sky_moments = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Add second moments in sky (RA/Dec) and alt/az coordinates to the output table.",
+    )
+    do_add_fgcm_photometry = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Add FGCM photometry for all bands to the output table.",
+    )
+    fgcmPhotometryBands = pexConfig.ListField(
+        dtype=str,
+        default=['g', 'r', 'i', 'z', 'y'],
+        doc="List of bands to include for FGCM photometry (telescope-dependent).",
     )
 
     def setDefaults(self):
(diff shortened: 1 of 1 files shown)


kklaliotis (commented):
I double-checked the conversions here and they all agreed with what I was doing before. 

I think the only change I'd maybe make is making the new column names align a little better with the current versions of the catalog columns. For instance, the "I" has mostly been removed so source moments are called `slot_Shape_xx` now. So then I would maybe include the sky coords moments as `slot_Shape_uu`. The alt-az labels are a bit clunky but I think for the sake of clarity I'd leave them essentially as they are, just again omitting the I to be eg. `slot_Shape_altalt` . Similarly I'd shift the naming convention for psf shape columns to match the others with `PsfShape` now instead of `psfShape` .

 But I think overall the actual implementation looks good!

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -492,6 +583,129 @@ def _make_selection_schema_mapper(self, input_schema):
...
+
+            cd11 = cd_matrix[0, 0]
+            cd12 = cd_matrix[0, 1]
+            cd21 = cd_matrix[1, 0]
+            cd22 = cd_matrix[1, 1]
+
+            # Transform moments: M_sky = CD * M_pixel * CD^T
+            # Iuu = CD11*(Ixx*CD11 + Ixy*CD21) + CD12*(Ixy*CD11 + Iyy*CD21)
+            iuu[i] = (cd11 * (ixx[i] * cd11 + ixy[i] * cd21)
+                      + cd12 * (ixy[i] * cd11 + iyy[i] * cd21))
```
rmjarvis: This doesn't look right to me.  I think the correct formula is:
```
iuu[i] = (cd11 * (ixx[i] * cd11 + ixy[i] * cd12)
                      + cd12 * (ixy[i] * cd11 + iyy[i] * cd12))
```
I think maybe you forgot to transpose the final matrix.

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -492,6 +583,129 @@ def _make_selection_schema_mapper(self, input_schema):
...
+            # Ivv = CD21*(Ixx*CD12 + Ixy*CD22) + CD22*(Ixy*CD12 + Iyy*CD22)
+            ivv[i] = (cd21 * (ixx[i] * cd12 + ixy[i] * cd22)
+                      + cd22 * (ixy[i] * cd12 + iyy[i] * cd22))
```
rmjarvis: Similarly, this should be
```
ivv[i] = (cd21 * (ixx[i] * cd21 + ixy[i] * cd22)
                      + cd22 * (ixy[i] * cd21 + iyy[i] * cd22))
```

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -492,6 +583,129 @@ def _make_selection_schema_mapper(self, input_schema):
...
+            # Iuv = (CD11*Ixx + CD12*Ixy)*CD21 + (CD11*Ixy + CD12*Iyy)*CD22
+            iuv[i] = ((cd11 * ixx[i] + cd12 * ixy[i]) * cd21
+                      + (cd11 * ixy[i] + cd12 * iyy[i]) * cd22)
```
rmjarvis: This one I believe is correct.

rmjarvis (changes requested):
I don't thnk the formulae are quite right.

python/lsst/pipe/tasks/finalizeCharacterization.py
taranu: I don't know about the correctness but the functors have a unit test comparing against afw. Either way, this code should either use the functors directly or refactor to use one common method, not copy-paste.

Someone should also check that the afw and Python code use the same matrix indexing convention.

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -492,6 +583,129 @@ def _make_selection_schema_mapper(self, input_schema):
...
+        # (CD matrix is in degrees/pixel per FITS standard)
+        deg2_to_arcsec2 = 3600.0 ** 2
+
+        for i in range(n_sources):
+            # Skip sources with invalid moments
+            if not np.isfinite(ixx[i]) or not np.isfinite(iyy[i]) or not np.isfinite(ixy[i]):
+                continue
+
+            center = geom.Point2D(x[i], y[i])
+            cd_matrix = wcs.getCdMatrix(center)
```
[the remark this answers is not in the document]
rmjarvis: I'd suggest in this function calling it jac or jacobian.  And add a comment that despite the name, this function does return the local jacobian at the specified location.

TallJimbo:
> It would be convenient if this routine converted the moments into the normal WL convention, rather than requiring an addition sign flip to g2 downstream of this function.

Would it be just as good from your perspective if we applied that flip between the second moments and (g1, g2)?  I'm a little wary of changing our sky axis definition for moments to be something other than the WCS definition, since those have non-lensing consumers.

rmjarvis:
> Would it be just as good from your perspective if we applied that flip between the second moments and (g1, g2)? 

Up to you I guess.  Personally, I find the WCS convention weird.  The u,v coordinates are from the perspective of God looking through the universe down at Earth.  But whatever you find appropriate I suppose.

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -669,6 +886,43 @@ def add_src_colors(self, srcCat, fgcmCat, band):
...
+
+        Parameters
+        ----------
+        srcCat : `lsst.afw.table.SourceCatalog`
+            Source catalog to add photometry to.
+        fgcmCat : `numpy.ndarray`
+            FGCM standard star catalog with ra, dec, and mag_* columns.
+        """
+        # Initialize all FGCM magnitude columns to NaN
+        for band_name in ('u', 'g', 'r', 'i', 'z', 'y'):
```
arunkannawadi: And here and below, instead of assuming the bands, you can extract the relevant columns from the schema as
```python
fgcm_mag_cols = srcCat.schema.extract("fgcm_mag_*")
```
and then loop over them.

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -97,7 +97,10 @@ def __init__(self, *, config=None):
         super().__init__(config=config)
         if config is None:
             return None
-        if not self.config.psf_determiner['piff'].useColor:
+        # Keep fgcm_standard_star if using chromatic PSF OR if adding FGCM photometry
+        needs_fgcm = (self.config.psf_determiner['piff'].useColor
+                      or self.config.doAddFgcmPhotometry)
```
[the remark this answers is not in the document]
arunkannawadi: This is working only because `config` is `None`. It will break if `config` is passed, so `self.config` should be changed to `config`.

PR #1286, opened 2026-03-30
DM-51163: Enable multi-tract matching

taranu:
This also updates various configs to DRP v2 pipelines standards.

### DIFF
--- a/python/lsst/pipe/tasks/diff_matched_tract_catalog.py
+++ b/python/lsst/pipe/tasks/diff_matched_tract_catalog.py
@@ -262,12 +262,12 @@ class DiffMatchedTractCatalogConfig(
     )
     columns_target_select_true = pexConfig.ListField[str](
         doc='Target table columns to require to be True for selecting sources',
-        default=('detect_isPrimary',),
+        default=[],
         listCheck=is_sequence_set,
     )
     columns_target_select_false = pexConfig.ListField[str](
         doc='Target table columns to require to be False for selecting sources',
-        default=('merge_peak_sky',),
+        default=[],
         listCheck=is_sequence_set,
     )
     coord_format = pexConfig.ConfigField[ConvertCatalogCoordinatesConfig](
--- a/python/lsst/pipe/tasks/diff_matched_tract_catalog.py
+++ b/python/lsst/pipe/tasks/diff_matched_tract_catalog.py
@@ -158,10 +158,7 @@ class MatchedCatalogFluxesConfig(pexConfig.Config):
         return columns
 
 
-class DiffMatchedTractCatalogConfig(
-    pipeBase.PipelineTaskConfig,
-    pipelineConnections=DiffMatchedTractCatalogConnections,
-):
+class DiffMatchedTractCatalogBaseConfig(pexConfig.Config):
     column_match_candidate_ref = pexConfig.Field[str](
         default='match_candidate',
         doc='The column name for the boolean field identifying reference objects'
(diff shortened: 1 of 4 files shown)


python/lsst/pipe/tasks/diff_matched_tract_catalog.py
```diff
@@ -309,34 +307,27 @@ def validate(self):
...
+    refcat_sharding_type = pexConfig.ChoiceField[str](
+        doc="The type of sharding (spatial splitting) for the reference catalog",
+        allowed={"tract": "Tract-based shards", "none": "No sharding at all"},
+        default="tract",
+    )
+    target_sharding_type = pexConfig.ChoiceField[str](
+        doc="The type of sharding (spatial splitting) for the target catalog",
+        allowed={"tract": "Tract-based shards", "none": "No sharding at all"},
+        default="tract",
+    )
```
sr525, line 324: What is no sharding? Can this work with healpix?

python/lsst/pipe/tasks/match_tract_catalog.py
```diff
@@ -257,3 +439,44 @@ def run(
...
+        if compute_tract_target or compute_tract_ref:
+            if not self.config.diff_matched_catalog.coord_format.coords_spherical:
+                raise RuntimeError(
+                    f"Can't compute tract columns with unless"
+                    f" {self.config.diff_matched_catalog.coord_format.coords_spherical} == True"
+                )
+            ref_c, target_c = self.config.diff_matched_catalog.coord_format.format_catalogs(
+                catalog_ref=catalog_ref, catalog_target=catalog_target,
+            )
+            SpherePoint = lsst.geom.SpherePoint
```
sr525: Why isn't this just imported like this?

python/lsst/pipe/tasks/diff_matched_tract_catalog.py
taranu: No sharding literally means no sharding at all - no spatial dimension, not even a skymap.

The reason I added this option is some catalogs are small enough that they are entirely within one tract (9813 for COSMOS, 5063 for ECDFS), so it seemed wasteful to have to reingest the same catalog multiple times for skymaps that often have identical tract definitions - especially so if you end up matching two such small external catalogs with each other. Having said that, in retrospect, the benefits aren't that great and I'm not sure if I want to recommend doing it this way.

In principle yes, one could add healpix support, but it would not be completely trivial.

python/lsst/pipe/tasks/match_tract_catalog.py
taranu: It seemed odd to do a `from` import for an attribute of a module that was already imported, but I suppose it's not wrong.

python/lsst/pipe/tasks/match_tract_catalog.py
```diff
@@ -171,10 +234,29 @@ class MatchTractCatalogConfig(
...
+    )
     match_tract_catalog = pexConfig.ConfigurableField(
         target=MatchTractCatalogSubTask,
         doc="Task to match sources in a reference tract catalog with a target catalog",
     )
+    output_matched_catalog = pexConfig.Field[bool](
+        doc="Whether to run the diff_matched_catalog task and write a matched catalog,"
+            " not just the catalogs of match indices",
+        default=False,
+    )
```
sr525, line 259: Is there an option anywhere to output the catalogue of objects that did not match? I've found that useful a few times.

taranu: [To include the unmatched objects](https://github.com/lsst/pipe_tasks/blob/c6c61205a15cc55f516fa77ac32a879a074f7836/python/lsst/pipe/tasks/diff_matched_tract_catalog.py#L182), yes. If you mean to output that catalog separately, no. That would be helpful for performance but more inconvenient so I haven't bothered yet.

PR #1290, opened 2026-04-06
tickets/DM-54190 Refactor of the RGB pretty picture code.

natelust:


python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -270,6 +229,43 @@ def setDefaults(self):
...
+        Notes
+        -----
+        The following deprecated fields are migrated:
+        - ``gamutMethod`` -> ``gamutMapperConfig.gamutMethod``
+        - ``exposureBrackets`` -> ``exposureBracketerConfig.exposureBrackets``
+        - ``doLocalContrast`` -> ``localContrastConfig.doLocalContrast``
+        - ``doPSFDeconcovlve`` -> ``doPSFDeconvolve``
+        """
+        # check if gamutMethod is set
+        if len(self._history["gamutMethod"]) > 1:
```
TallJimbo: @andy-slac recently shared a story at a middleware meeting about Python garbage-collection troubles in which it looked like a large fraction of the effort was being spent trying to collect pex_config history objects, and I suggested that we could probably turn that stuff off by default and only turn it on when somebody asked for it.

And I'm not thrilled that this is accessing a private attribute.  Are we sure remapping the config options is worth it?

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -792,7 +887,13 @@ def run(self, inputCoadd: Exposure):
             This `Struct` will have an attribute named ``outputCoadd``.
 
         """
-        background = self.fixBackground(inputCoadd.image.array)
+        # background = self.fixBackground(inputCoadd.image.array)
```
TallJimbo: relic

python/lsst/pipe/tasks/prettyPictureMaker/_colorMapper.py
```diff
@@ -1,548 +1,215 @@
...
-        Key word arguments passed to the scaleColor function.
-    remapBounds : `Callable` or `None`
-        This is a callable that should remaps the input arrays such that each of
+    remap_bounds : `RemapBoundsFunction` or `None`
+        This is a callable that remaps the input arrays such that each of
         them fall within a zero to one range. This callable is given the
-        initial image as well as any parameters defined in the remapBoundsKwargs
-        parameter. Set to None for no remapping.
-    remapBoundsKwargs : `Mapping` or None
+        initial image. Set to None for no remapping.
```
TallJimbo: When this is `None`, does that mean the input arrays must already be 0-1?

python/lsst/pipe/tasks/prettyPictureMaker/_localContrast.py
```diff
@@ -192,13 +197,13 @@ def makeLapPyramid(
         List of padding sizes for rows (vertical padding).
     padX : `list` of `int`
         List of padding sizes for columns (horizontal padding).
-    gaussOut : `numba.typed.typedlist.List` of `NDArray` or None
+    gaussOut : `List` of `NDArray` or None
```
TallJimbo:
```suggestion
    gaussOut : `list` of `NDArray` or None
```
please search for and fix other occurrences in this file

TallJimbo (commented):
Not done yet; just committing my comments.

python/lsst/pipe/tasks/prettyPictureMaker/_localContrast.py
```diff
@@ -279,15 +283,15 @@ def _calculateOutput(
...
+                if not (val >= gamma[0] and val < gamma[1]):
                     continue
                 a = (plevelY[x] - gamma[0]) / (gamma[1] - gamma[0])
                 outLevelY[x] = (1 - a) * basisBottomY[x] + a * basisTopY[x]
 
 
 def levelPadder(numb: int, levels: int) -> list[int]:
-    """Determine if each level of transform will need to be padded by
-    one to make the level divisible by two.
+    """   Determine if each level of transform will need to be padded by
```
TallJimbo:
```suggestion
    """Determine if each level of transform will need to be padded by
```

python/lsst/pipe/tasks/prettyPictureMaker/_plugins.py
```diff
@@ -112,11 +152,29 @@ def register(self, order: float, kind: PluginType) -> Callable:
...
+            Parameters
+            ----------
+            func : `Callable`
+                Plugin function being registered.
+
+            Returns
+            -------
+            func : `Callable`
+                The same plugin function, now registered in the registry.
+
```
TallJimbo:
```suggestion
```

TallJimbo (approved):
Lots of minor comments (especially on docstrings), but no blockers.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: uhhhhh, so this was a long conversation with Yusra, and her strong opinion is we don't break configs and we have a process for a reason. I could go either way on this as we probably don't have many users for this yet.

On the other point though, I don't really know any other way to do this sort of thing with pex_config, and it is needed. Perhaps that means there needs to be a method to ask an object if a field has been changed.

Also on that, I do know that I have even just recently used the history feature to track down where a config was being set and I know Robert uses it often, so I would be a little sad for it to be gone, though maybe a re-work of it is in order, perhaps on the ticket mentioned above.

TallJimbo: I imagine you could just check the old option and the new option every time you inspect one of those values, or just remap the pairs of config fields to single Task attributes at Task construction.   The latter seems pretty similar to what you're doing here but without most of the drawbacks.

python/lsst/pipe/tasks/prettyPictureMaker/_functors/_lum_scale.py
```diff
@@ -0,0 +1,173 @@
...
+        # Scale the values with linear manipulation for contrast
+        intensities = abs(intensities)
+        # intensities = rgb_diffusion.diffuse_gray_image(intensities)
+        nj_to_lum = rgb.RGB_to_Oklab(
+            np.array([[[self.floor, self.floor, self.floor]]], dtype=float), (0.28, 0.28)
+        )[0, 0, 0]
+        top = np.arcsinh(self.stretch)
+        bottom = (np.arcsinh(nj_to_lum * self.stretch) - 0.2 * top) / 0.8
+        intensities = (np.arcsinh(intensities * self.stretch) - bottom) / (top - bottom)
+        logging.debug("arcsinh max %.4f", intensities.max())
```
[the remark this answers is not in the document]
natelust: Original I'm not quite sure what you mean here. Is that preferred over: logger = logging.getLogger(__name__)? I agree I need to do at least one

Edit: I looked at that logger and see, it is the same, just convenience.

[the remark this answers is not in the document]
timj: A lot of the time the Task has a logger (task.log) and you would pass that task logger into lower APIs so that the log messages are associated with the task.

Does `logging.debug` work out the logger name itself or use the root logger? At very least the convention is to name the logger `__name__`.

TallJimbo (approved):

PR #1293, opened 2026-04-09
DM-54555: Switch brightStarCutout to lsst.images output

leeskelvin:


python/lsst/pipe/tasks/brightStarSubtraction/brightStarCutout.py
```diff
@@ -458,36 +457,42 @@ def _get_bright_star_stamps(
...
             stamp_wcs = makeModifiedWcs(pixels_to_stamp_frame, input_exposure.wcs, False)
+            projection = Projection.from_legacy(
+                stamp_wcs,
+                DetectorFrame(
+                    instrument=input_exposure.visitInfo.instrumentLabel,
+                    visit=input_exposure.visitInfo.getId(),
+                    detector=input_exposure.detector.getId(),
+                    bbox=Box.from_legacy(stamp_bbox),
+                ),
+            )
```
TallJimbo: Because this `Projection` is for the pixels in your special star-centered, focal-plane-radial frame, I don't think `DetectorFrame` is appropriate here.

Unfortunately, I don't think we have a `Frame` type in `lsst.images` that would be appropriate, and that's not something the library is designed to let you define outside of it.

I've been sort of anticipating that we'd hit this limitation sooner or later (I know the current set of `Frame` types isn't complete), but I wanted to have at least one concrete use case before trying to solve the problem.  Since this is such a niche use case, I don't think we want to add a `Frame` type that exactly describes it.  A `Frame` type for coordinate systems centered on objects or sources might be appropriate - but I'm not sure what utility it would have, or what it might actually have in common with other coordinate systems that could be described that way.  So I think it's probably just time to add a vague `GenericFrame` (naming ideas welcome) to `lsst/images/transforms/_frames.py` (sorry; we'll have to revive that PR) to use here.

python/lsst/pipe/tasks/extended_psf/extended_psf_candidates.py
```diff
@@ -0,0 +1,431 @@
...
+            metadata=metadata,
+        )
+
+        self._psf = psf
+        self._stamp_info = stamp_info or BrightStarStampInfo()
+
+    def __getitem__(self, bbox: Box | EllipsisType) -> BrightStarStamp:
+        super().__getitem__(bbox)
+        if bbox is ...:
+            return self
```
TallJimbo, line 175: This `if` block needs to move before the `super()` call, I think.  This was a common problem in the `lsst.images` types that we recently fixed.

python/lsst/pipe/tasks/brightStarSubtraction/brightStarStamps.py
```diff
@@ -0,0 +1,431 @@
...
+        return f"BrightStarStamp({self.image!s}, {list(self.mask.schema.names)}, {self.stamp_info})"
+
+    def __repr__(self) -> str:
+        return (
+            f"BrightStarStamp({self.image!r}, mask_schema={self.mask.schema!r}, "
+            f"stamp_info={self.stamp_info!r})"
+        )
+
+    @property
+    def psf(self) -> Image:
```
TallJimbo: Might be best to rename this to `psf_kernel_image` so it's not as easily confused with attributes of other types that return an actual `PointSpreadFunction` instance.

python/lsst/pipe/tasks/extended_psf/extended_psf_candidates.py
```diff
@@ -0,0 +1,431 @@
...
+    ):
+        super().__init__(
+            image,
+            mask=mask,
+            variance=variance,
+            mask_schema=mask_schema,
+            projection=projection,
+            obs_info=obs_info,
+            metadata=metadata,
+        )
```
TallJimbo, line 168: It turns out `Image` and `Mask` can hold an `ObservationInfo` directly, too, so it's possible that this model is going to end up duplicating that information on serialization.

But looking deeper, this is mostly an upstream problem; we should be treating the `ObservationInfo` in `MaskedImage` more like we treat `Projection` - making sure we set it on all component images at construction, and then telling the components not to save it in `serialize`.  And we're very inconsistent about that in both `MaskedImage` and `VisitImage`.  I think if that got fixed, you could remove `obs_info` from your serialization model here, because you'd get one on the `MaskedImage` serialization model.

I'd like to get this all worked out before you start saving things in the new format, but I don't want to keep blocking you.  Since I think you'll need an `lsst.images` branch for this ticket after all - for the frame issue - how about I add a commit fixing all of these upstream issues to that branch, and then adjusting here downstream?

python/lsst/pipe/tasks/brightStarSubtraction/brightStarStamps.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def writeFits(self, filename: str):
+        """Write this `BrightStarStamps` object to a FITS file.
+
+        Parameters
+        ----------
+        filename : `str`
+            Name of the FITS file to write.
+        """
+        fits.write(self, filename)
```
TallJimbo: Per comment on `daf_butler`, I'm hoping you can switch the formatter to `lsst.images.fits.GenericFormatter` and then drop `readFits[WithOptions]` and `writeFits`.

tests/test_brightStarStamps.py
```diff
@@ -0,0 +1,147 @@
...
+        del self.bright_star_stamps
+        del self.stamp_infos
+        del self.global_metadata
+
+    def testBrightStarStamps(self):
+        """Test that BrightStarStamps can be serialized and deserialized."""
+
+        with NamedTemporaryFile() as file:
+            self.bright_star_stamps.writeFits(file.name)
+            bright_star_stamps = BrightStarStamps.readFits(file.name)
```
TallJimbo: If you want to test the butler support here as well, rewrite this as:
```
with lsst.images.tests.RoundtripFits(self.bright_star_stamps, storageClass="BrightStarStamps") as roundtrip:
    pass
bright_star_stamps = roundtrip.result
```
You can also use `roundtrip.inspect()` inside the context manager to poke at what the FITS file actually looks like with `astropy.io.fits` if you want to test that as well.

python/lsst/pipe/tasks/extended_psf/extended_psf_candidates.py
leeskelvin: That sounds good, please do make use of the prior `lsst.images` branch if helpful. More than happy to wait a little longer to get this right from the outset.

python/lsst/pipe/tasks/brightStarSubtraction/brightStarCutout.py
leeskelvin: Switched to:
```
projection = Projection.from_legacy(stamp_wcs, GeneralFrame(unit=u.pixel))
```

python/lsst/pipe/tasks/extended_psf/extended_psf_candidates.py
leeskelvin: This led to a fair amount of simplification inside `brightStarStamps.py` beyond just `obs_info`, thank you.

python/lsst/pipe/tasks/brightStarSubtraction/brightStarStamps.py
leeskelvin: Switched to `lsst.images.fits.formatters.GenericFormatter`, which I think was the intent? (Rather than modifying the `__init__` to raise `GenericFormatter` into `fits`).

python/lsst/pipe/tasks/extended_psf/extended_psf_candidates.py
```diff
@@ -0,0 +1,339 @@
...
+    position_x: float | None = None
+    position_y: float | None = None
+    focal_plane_radius: Quantity | None = None
+    focal_plane_angle: Quantity | None = None
+
+    def __str__(self) -> str:
+        attrs = ", ".join(f"{k}={v!r}" for k, v in self.__dict__.items())
+        return f"BrightStarStampInfo({attrs})"
+
+    __repr__ = __str__
```
TallJimbo, line 92: If you define `__repr__` instead, `str` will automatically delegate to it.

TallJimbo (approved):

python/lsst/pipe/tasks/extended_psf/extended_psf_candidates.py
leeskelvin: In this case, it seems not though. For my testing on an `epc.star_info` object, I have:
```
epc = epcs[0]
print(epc.star_info)
epc.star_info
```
returning:
```
visit=2025102600226 detector=178 ref_id=4980307157180392448 ref_mag=16.378769863064512 position_x=3485.5599400703168 position_y=2400.2426411600445 focal_plane_radius=<Quantity 300.59689709 mm> focal_plane_angle=<Quantity 1.52257155 rad>

ExtendedPsfCandidateInfo(visit=2025102600226, detector=178, ref_id=4980307157180392448, ref_mag=16.378769863064512, position_x=3485.5599400703168, position_y=2400.2426411600445, focal_plane_radius=<Quantity 300.59689709 mm>, focal_plane_angle=<Quantity 1.52257155 rad>)
```

It looks like the delegation rule is only true when the inherited implementation is object.str. I forgot why I did it this way originally, but revisiting it now, I realized that I don't like the default pydantic `__str__`, without any commas between the items, nor identifying the storage class name at the beginning.

TallJimbo: Ah, that makes sense; I wasn't think about whether the `__str__` Pydantic provided.

leeskelvin:
Thanks for the review comments above @TallJimbo. A fair amount of reshuffling has occurred at the end of this PR. I initially switched from `brightStar*` to `extendedPsf*` everywhere, but then it occurred to me that perhaps namespaces and file names should also be in `snake_case` instead of `camelCase` (i.e., `extended_psf`). I changed the namespaces and filenames to `snake_case` on a subsequent commit, making it easy to revert if you prefer the original layout.

Other than that, as discussed, I've removed the legacy extended PSF code here as well. And, as a last-minute addition, I added read_/write_fits methods for the `ExtendedPsfCandidates` class as well. I've tested that these outputs can be saved and read back in successfully in a local notebook.

TallJimbo:
:+1:, all sounds good.  Subpackage boundaries seem like a perfectly good place to switch to snake_case, and I think matching the style of the code within the subpackage is a good idea.

PR #1295, opened 2026-04-15
DM-54227: Switch to using Images.ColorImage as container

natelust:


python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -104,10 +105,17 @@ class PrettyPictureConnections(
...
+        name=BaseSkyMap.SKYMAP_DATASET_TYPE_NAME,
+        dimensions=("skymap",),
+    )
+
     outputRGB = Output(
         doc="A RGB image created from the input data stored as a 3d array",
-        name="rgb_picture_array",
-        storageClass="NumpyArray",
+        name="rgb_picture",
+        storageClass="ColorImage",
```
leeskelvin, line 118: Do you also need to modify `inputRGB` for `PrettyMosaicConnections` to match? I'm unsure if you want to keep the _output_ as a numpy array in that task too.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -506,7 +523,10 @@ def run(self, images: Mapping[str, Exposure]) -> Struct:
         # pack the joint mask back into a mask object
         lsstMask = Mask(width=jointMask.shape[1], height=jointMask.shape[0], planeDefs=maskDict)
         lsstMask.array = jointMask  # type: ignore
-        return Struct(outputRGB=colorImage.astype(dtype), outputRGBMask=lsstMask)  # type: ignore
+        return Struct(
+            outputRGB=ColorImage(colorImage.astype(dtype), bbox=image_box, projection=image_wcs),
```
leeskelvin, line 527: AFAICT, `image_box` and `image_wcs` are not used by downstream tasks like `HighOrderHipsTask`, right? That task seems to use the skymap to derive what it needs. Do you intend to replace the skymap call in that task to use these values as well?

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -369,20 +377,29 @@ def _match_sigmas_and_recenter(self, *arrays, factor=1):
...
+            A box that defines this image as part of a larger region.
 
         Returns
         -------
         result : `Struct`
             A struct with the corresponding RGB image, and mask used in
             RGB image construction. The struct will have the attributes
-            outputRGBImage and outputRGBMask. Each of the outputs will
+            outputRGB and outputRGBMask. Each of the outputs will
             be a `NDarray` object.
```
leeskelvin: Old reference to `NDarray`.

leeskelvin (approved):
Looking healthy, with a few comments which I hope are of use.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: You are right I did miss updating that

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: At the moment no, image_box and image_wcs are new `lsst.images` objects that are being written out for end user sake. I would rather keep the internals consistent as is for now instead of doing a lot of `to_legacy` calls. Later on down the line this work will get put in as well start to transition more of the new type objects.

natelust: The additional context for this is smaller changes are better right now for this being ported back for dp2

PR #1296, opened 2026-04-19
DM-54675: Add higher-order moments to ExposureSummaryStats

laurenam:


python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -506,6 +518,42 @@ def update_psf_stats(
...
+            starE41Median = float(np.nanmedian(starM40 - starM04))
+            starE42Median = float(np.nanmedian(2.0*(starM31 + starM13)))
+
+            summary.starComa1Median = float(starComa1Median)
+            summary.starComa2Median = float(starComa2Median)
+            summary.starTrefoil1Median = float(starTrefoil1Median)
+            summary.starTrefoil2Median = float(starTrefoil2Median)
+            summary.starKurtosisMedian = float(starKurtosisMedian)
+            summary.starE41Median = float(starE41Median)
+            summary.starE42Median = float(starE42Median)
```
arunkannawadi, line 551: Do we need to cast it to float again given that we already did it in the previous block of code? Even better, why not just make the earlier ones attributes of `summary` instead of a 2-step process?

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -506,6 +518,42 @@ def update_psf_stats(
...
+            starM12 = psf_cat[self.config.starHigherOrderMomentBase + "_12"]
+            starM21 = psf_cat[self.config.starHigherOrderMomentBase + "_21"]
+            starM30 = psf_cat[self.config.starHigherOrderMomentBase + "_30"]
+            starM04 = psf_cat[self.config.starHigherOrderMomentBase + "_04"]
+            starM13 = psf_cat[self.config.starHigherOrderMomentBase + "_13"]
+            starM22 = psf_cat[self.config.starHigherOrderMomentBase + "_22"]
+            starM31 = psf_cat[self.config.starHigherOrderMomentBase + "_31"]
+            starM40 = psf_cat[self.config.starHigherOrderMomentBase + "_40"]
+
+            starComa1Median = float(np.nanmedian(starM30 + starM12))
```
arunkannawadi:
```suggestion
            summary.starComa1Median = float(np.nanmedian(starM30 + starM12))
```

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -506,6 +518,42 @@ def update_psf_stats(
...
+            starComa2Median = float(np.nanmedian(starM21 + starM03))
+            starTrefoil1Median = float(np.nanmedian(starM30 - 3.0*starM12))
+            starTrefoil2Median = float(np.nanmedian(3.0*starM21 - starM03))
+            starKurtosisMedian = float(np.nanmedian(starM40 + 2.0*starM22 + starM04))
+            starE41Median = float(np.nanmedian(starM40 - starM04))
+            starE42Median = float(np.nanmedian(2.0*(starM31 + starM13)))
```
arunkannawadi: These equations look correct to me!

arunkannawadi (approved):

python/lsst/pipe/tasks/computeExposureSummaryStats.py
laurenam: Good point about the double casting!  As for the 2-step process, I was following the established pattern in the file (I _think_ we generally prefer consistency in a given file?)

arunkannawadi: Things started getting assigned to `summary`, i.e.., `summary.psfStarDeltaE1Median` before this, so it should be okay. Up to you.

PR #1299, opened 2026-04-30
DM-54843: Add fixed G12 option and error floor to HG12 photometric fitting

mjuric:
Summary

  This branch adds the ability to fix the G12 slope parameter in HG12 photometric fitting, along with a systematic error floor,  and exposes both as pipeline configuration options.

  Changes

  python/lsst/pipe/tasks/ssp/photfit.py
  - fitHG12 now accepts fixedG12 — when set, G12 is held fixed and only H (absolute magnitude) is fit. The 7-element return  signature is preserved, with G_err=NaN and HG_cov=NaN when G12 is fixed, and reduced chi-squared uses nobsv - 1 instead of nobsv - 2.
  - fitHG12 now accepts magSigmaFloor — a systematic magnitude error (default 0.0) added in quadrature to measurement errors  before fitting. This accounts for unmodeled systematics (calibration, rotation curves) that dominate over photon noise for  bright objects.
  - Added an early return for empty input arrays, which previously caused an IndexError.

  python/lsst/pipe/tasks/consolidateSsTables.py
  - Added fixedG12 config field (pexConfig.Field, optional float, default None) — when set, fixes G12 to the given value across  all bands.
  - Added magSigmaFloor config field (float, default 0.01 mag).
  - Logs both config values before SSObject computation.

  python/lsst/pipe/tasks/ssp/ssobject.py
  - Threaded fixedG12 and magSigmaFloor from compute_ssobject through compute_ssobject_entry to fitHG12 using functools.partial.
  - Updated nDof calculation to account for the reduced number of free parameters when G12 is fixed.

  tests/test_photfit.py (new)
  - Unit tests verifying HG12 model evaluation against independently calculated test vectors (H=17.30, G12=0.42, 15 observations  spanning 0.8–60° phase angle).
  - Tests for free G12 recovery, fixed G12 recovery, wrong fixed G12 producing worse chi2, magSigmaFloor reducing chi2, graceful  handling of degenerate inputs, and distance correction consistency.

  analysis/hg12_comparison.ipynb (new; will be deleted via rebase before merging to 'main')
  - Analysis notebook comparing a DP2 run with free G12 vs fixed G12=0.5. Key findings: fixing G12 nearly doubles successful fits  in sparse bands, eliminates unphysical G12 values (30–58% were outside [-0.5, 1.5]), and produces sensible H error bars  (previously ~10^7 due to ill-conditioned covariance).


analysis/hg12_comparison.ipynb
isullivan, line 1: This notebook is useful, but does not belong in `pipe_tasks` (I see your comment now that you intended to do this).

python/lsst/pipe/tasks/ssp/photfit.py
```diff
@@ -107,32 +107,55 @@ def fit(mag, phase, sigma, model=HG12_model, params=[0.1]):
     return sol
 
 
-def fitHG12(mag, magSigma, phaseAngle, tdist, rdist):
+def fitHG12(mag, magSigma, phaseAngle, tdist, rdist, fixedG12=None, magSigmaFloor=0.0):
```
isullivan, line 116: Please add a docstring briefly describing these parameters and what this function returns.

python/lsst/pipe/tasks/ssp/photfit.py
```diff
@@ -107,32 +107,55 @@ def fit(mag, phase, sigma, model=HG12_model, params=[0.1]):
...
     # This uses the HG12 function from Muinonen et al (2010)
     nobsv = len(mag)
 
+    if nobsv == 0:
+        return (np.nan,) * 6 + (0,)
```
isullivan: This return seems fragile to me, it will break if `HG12_model` ever changes what it returns. That may be unavoidable, but all the more reason to define the return value in the docstring.

isullivan (approved):
The code changes look fine to me, but a small amount of documentation added to define the input and returned parameters would be very helpful.

python/lsst/pipe/tasks/ssp/photfit.py
mjuric: Good point -- done!

python/lsst/pipe/tasks/ssp/photfit.py
mjuric: Good point -- I'm now returning a `namedtuple` (and I changed most of the callers not to use indexing but explicitly access the members).

mjuric:
Closing in favor of https://github.com/lsst/pipe_tasks/pull/1300 (which has the correctly named branch).

PR #1300, opened 2026-05-02
DM-54843: Add fixed G12 option and error floor to HG12 photometric fitting

mjuric:
Summary

  This branch adds the ability to fix the G12 slope parameter in HG12 photometric fitting, along with a systematic error floor,  and exposes both as pipeline configuration options.

  Changes

  python/lsst/pipe/tasks/ssp/photfit.py
  - fitHG12 now accepts fixedG12 — when set, G12 is held fixed and only H (absolute magnitude) is fit. The 7-element return  signature is preserved, with G_err=NaN and HG_cov=NaN when G12 is fixed, and reduced chi-squared uses nobsv - 1 instead of nobsv - 2.
  - fitHG12 now accepts magSigmaFloor — a systematic magnitude error (default 0.0) added in quadrature to measurement errors  before fitting. This accounts for unmodeled systematics (calibration, rotation curves) that dominate over photon noise for  bright objects.
  - Added an early return for empty input arrays, which previously caused an IndexError.

  python/lsst/pipe/tasks/consolidateSsTables.py
  - Added fixedG12 config field (pexConfig.Field, optional float, default None) — when set, fixes G12 to the given value across  all bands.
  - Added magSigmaFloor config field (float, default 0.01 mag).
  - Logs both config values before SSObject computation.

  python/lsst/pipe/tasks/ssp/ssobject.py
  - Threaded fixedG12 and magSigmaFloor from compute_ssobject through compute_ssobject_entry to fitHG12 using functools.partial.
  - Updated nDof calculation to account for the reduced number of free parameters when G12 is fixed.

  tests/test_photfit.py (new)
  - Unit tests verifying HG12 model evaluation against independently calculated test vectors (H=17.30, G12=0.42, 15 observations  spanning 0.8–60° phase angle).
  - Tests for free G12 recovery, fixed G12 recovery, wrong fixed G12 producing worse chi2, magSigmaFloor reducing chi2, graceful  handling of degenerate inputs, and distance correction consistency.

Note: this was developed and reviewed on https://github.com/lsst/pipe_tasks/pull/1299

### DIFF
--- a/python/lsst/pipe/tasks/consolidateSsTables.py
+++ b/python/lsst/pipe/tasks/consolidateSsTables.py
@@ -80,6 +80,13 @@ class ConsolidateSsTablesConfig(
             "If None (default), both H and G12 are fit.",
     )
 
+    magSigmaFloor = pexConfig.Field(
+        dtype=float,
+        default=0.01,
+        doc="Systematic magnitude error floor (mag) added in quadrature "
+            "to measurement errors before HG12 fitting.",
+    )
+
 
 class ConsolidateSsTablesTask(pipeBase.PipelineTask):
     """Consolidate per-patch ssSource tables into a single table.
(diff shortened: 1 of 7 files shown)


isullivan (approved):
Looks good

PR #1302, opened 2026-05-05
DM-54482: Integrate PSF-shapelet subtask into calibrateImage as a donut-detector

laurenam:


python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -748,6 +761,11 @@ def __init__(self, initial_stars_schema=None, **kwargs):
             doc="PSF max value.",
         )
         afwTable.CoordKey.addErrorFields(self.psf_schema)
+        # This is to ensure the shapelets metrics make it into the exposure
+        # summary stats even if all calibrations fail.
+        self.initial_summary = None
```
TallJimbo: Please pass this around as a local variable and method argument/return value instead of a task instance attribute.  That might require changing some signatures of (hopefully) private methods, but it makes it a lot easier to read how information and logic flows through the task.

python/lsst/pipe/tasks/computeExposureSummaryStats.py
```diff
@@ -307,14 +307,18 @@ def run(self, exposure, sources, background):
         exposure : `lsst.afw.image.ExposureF`
         sources : `lsst.afw.table.SourceCatalog`
         background : `lsst.afw.math.BackgroundList`
+        summary : `lsst.afw.image.ExposureSummary`, optional
+            Summary object to be appended to if not `None`.  If `None` a new
+            summary opject will be created.
```
TallJimbo:
```suggestion
            summary object will be created.
```

python/lsst/pipe/tasks/calibrateImage.py
```diff
@@ -1338,6 +1372,18 @@ def log_psf(msg, addToMetadata=False):
         self.metadata["initial_psf_positive_peak_count"] = detections.numPosPeaks
         self.metadata["initial_psf_negative_peak_count"] = detections.numNegPeaks
         self.psf_source_measurement.run(detections.sources, result.exposure)
+
```
TallJimbo, line 1369: This is closely related to a thread I started on the `meas_algorithm` PR, but what do you think about taking [some of?] the score calculation out of `ComputeRoughPsfShapeletsTask.run` (it can stay on another method in that task), so we can run the shapelet stuff _before_ `psf_source_measurement.run`, and then add the centroid-offset term into the score afterwards?  (I suppose we'd need some way to know whether it was included or not.)

python/lsst/pipe/tasks/calibrateImage.py
laurenam: Yeah, that’s what I tried to do initially, but because of the exception handling in `_compute_psf`, I couldn’t get it returned when the PSF modelling fails.  I’ll try factoring out the shapelets measurement into its own method (but will have to think through passing catalogs around).

python/lsst/pipe/tasks/calibrateImage.py
laurenam: I've updated both PRs to do the shapelet decomposition, which also includes the computation of power only from the non-atmospheric decompositions coefficients, `shapelets_only_iq_score`, before `psf_source_measurement.run`.  After the call to `psf_source_measurement.run`, a call to a separate method on the shapelets task computes `shapelets_iq_score` which includes power from the median centroid offset (where I require at least 3 finite `slot_Centroid_x` &  `slot_Centroid_y` entries, otherwise the metric will be `nan`).

TallJimbo (approved):

PR #1304, opened 2026-05-11
Tickets/DM-54868 hips and rgb fixes for background issues.

natelust:


python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -650,6 +654,64 @@ class PrettyPictureBackgroundFixerConnections(
...
+                    for offset_x, offset_y in itertools.product((-1, 0, 1), (-1, 0, 1)):
+                        if not (offset_x or offset_y):
+                            # Skip the input that matches the quantum data ID;
+                            # it's already got an edge to this quantum.
+                            continue
+                        neighbor_patch_id = tract_info.getSequentialPatchIndexFromPair(
+                            Index2D(x=patch_index.x + offset_x, y=patch_index.y + offset_y)
+                        )
+                        neighbor_data_id = DataCoordinate.standardize(base_data_id, patch=neighbor_patch_id)
+                        if neighbor_data_id not in flat_data_ids:
```
fred3m, line 731: Is this sufficient? Maybe I'm wrong but it looks like you'll miss when x wraps around.

For example, for patch `(0, y)`, for any `y`, the x and y indices for offset (-1, 0) will be (-1, y)`, which will be translated to `num_x *y-1`, which will exist. A similar affect would be obtained for the right edge, so the code as it is might be incorrectly wrapping around the edges.

python/lsst/pipe/tasks/rgb2hips/_high_order_hips.py
```diff
@@ -224,7 +224,15 @@ def run(self, input_images: Iterable[tuple[NDArray, SkyWcs, Box2I]], healpix_id)
...
-                existing[warpped_box_slices][are_warpped] = warpped.array[are_warpped]
+                # where this existing array is not nan, there is already data
+                pre_populated = np.isfinite(existing[warpped_box_slices])
+                new_info = np.logical_not(pre_populated)
+                new_and_warp = np.logical_and(new_info, are_warpped)
+                pre_and_warp = np.logical_and(pre_populated, are_warpped)
+                existing[warpped_box_slices][new_and_warp] = warpped.array[new_and_warp]
+                existing[warpped_box_slices][pre_and_warp] = (
+                    warpped.array[pre_and_warp] + existing[warpped_box_slices][pre_and_warp]
+                ) * 0.5
```
fred3m, line 235: I think that this fails for corners where 4 tracts meet, right? Maybe that doesn't happen in HIPS space but at least in skymap space there are corners where 4 tracts meet, in which case this will not be the correct mean.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
-        numpy.ndarray
-            An array representing the estimated background level across the image.
+        result : `numpy.ndarray`
+            Boolean mask array where True indicates background pixels.
+
+        Notes
+        -----
+        This method works best for images with relatively uniform background. It may
+        not perform well in fields with high density or diffuse flux, as noted in
+        the implementation comments.
```
fred3m, line 969: This docstring needs to be updated. It has extra parameters and is missing a definition for `threshold_pair`.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
+            Input image array for which to find background pixels.
+        pos_sigma_mult : `float`
+            How many sigma to consider as background in the positive direction
+        max_flux_imbalance : `float`
+            Limit on the ratio of the area below the determined average to that
+            above. If the ratio is in excess of this value, then there is
+            assumed to be diffuse background flux in the image, and zero is
+            assumed to be the background level. Some excess flux is expected as
+            there is a real distribution of faint flux.
+        max_serach_flux : `float`
```
fred3m: `max_search_flux`

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
-        This method works best for images with relatively uniform background. It may
-        not perform well in fields with high density or diffuse flux, as noted in
-        the implementation comments.
+        max_flux_imbalance : `float`
+            Limit on the ratio of the area below the determined average to that
+            above. If the ratio is in excess of this value, then there is
+            assumed to be diffuse background flux in the image, and zero is
+            assumed to be the background level. Some excess flux is expected as
+            there is a real distribution of faint flux.
+        max_serach_flux : `float`
```
fred3m: `max_search_flux`

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -169,7 +174,14 @@ class PrettyPictureConfig(PipelineTaskConfig, pipelineConnections=PrettyPictureC
...
             "Flux threshold below which most flux will be considered noise, used to estimate noise properties"
         ),
-        default=10,
+        default=2,
+    )
+    maxNoiseImbalance = Field[float](
+        doc=(
+            "When recentering noise, if the ratio of counts of positive pixels, to negative pixesl passes "
+            "this threhsold, consider there to be extened low flux and only estimate noise below zero."
+        ),
```
fred3m: `pixels`, `threshold`, `extended`

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -893,17 +1073,94 @@ def run(self, inputCoadd: Exposure):
...
+            joint_thresh = self._findImageStatistics(
+                background_pixels,
+                self.config.pos_sigma_multiplier,
+                self.config.max_flux_imbalance,
+                self.config.max_search_flux,
+            )
+            # There is no background to be found, return early
+            if np.isinf(joint_thresh[0]):
+                # output = ExposureF(inputCoadd, deep=True)
+                # return Struct(outputCoadd=output)
```
fred3m: Remove commented code

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -893,17 +1073,94 @@ def run(self, inputCoadd: Exposure):
...
+
+                for value, y_pos, x_pos in zip(tmp_values, tmp_yloc, tmp_xloc):
+                    if np.sqrt((y_pos - center_y) ** 2 + (x_pos - center_x) ** 2) < inside_rad:
+                        values.append(value)
+                        yloc.append(y_pos)
+                        xloc.append(x_pos)
+
+        # At least 15 points are requred for TPS with 4th order polynomial
+        if len(yloc) < 15:
+            return Struct(outputCoadd=inputCoadd)
```
fred3m, line 1150: Should this be a `WorkNotFound` error? It seems like there should at least be a warning here instead of silently failing.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -893,17 +1073,94 @@ def run(self, inputCoadd: Exposure):
...
+            return Struct(outputCoadd=output)
+
+        if neighbors:
+            for n_exp in neighbors:
+                tmp_values, tmp_yloc, tmp_xloc = self._findControlPoints(
+                    n_exp, origin, self.config.use_detection_mask, joint_thresh
+                )
+                # if len(tmp_values) == 0:
+                #     output = ExposureF(inputCoadd, deep=True)
+                #     return Struct(outputCoadd=output)
```
fred3m: Remove commented code

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -893,17 +1073,94 @@ def run(self, inputCoadd: Exposure):
...
+        primary_ref = None
+        neighbor_refs = []
+        for ref in inputRefs.inputCoadd:
+            if quantum_patch == ref.dataId["patch"]:
+                primary_ref = ref
+            else:
+                neighbor_refs.append(ref)
+        if primary_ref is None:
+            # This should be unreachable
+            raise RuntimeError("There is a major problem input ref associated with quantum cant be found")
```
fred3m: Please fix grammar and typos in the raise message.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
+            # This means there is an excess flux, possibly diffuse source,
+            # assume all flux is diffuse
+            return np.inf, -np.inf
+
+        # create a new masking threshold that's the determined
         # mean plus std from the fit
-        threshhold = mu_hat + pos_sigma_mult * sigma_hat
-        image_mask = (image < threshhold) * (image > (mu_hat - 5 * sigma_hat))
-        return image_mask
+        threshhold_pos = min(mu_hat + pos_sigma_mult * sigma_hat, max_search_flux)
```
fred3m: `threshold_pos`

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -323,12 +335,22 @@ def _find_normal_stats(self, array):
 
         try:
             # Fit half-normal distribution to absolute negative values
-            mu, sigma = halfnorm.fit(np.abs(values_neg))
+            mu, sigma = halfnorm.fit(np.abs(values_neg - center))
```
fred3m: I think that you need to pass `floc=0` here, since your ABS call is essentially folding the positive and negative sides of the data together. Without setting `floc=0` the minimum of the fit will float, causing you to underestimate `mu`. It might not make a huge difference numerically but it's a real bias that can easily be avoided.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -323,12 +335,22 @@ def _find_normal_stats(self, array):
...
 
-        return center, sigma
+        # examine for excess positive flux, this means there is contaminating signal
+        new_cut = array[array < (mu + 3 * sigma)]
+        positivity_ratio = np.sum(new_cut > mu) / np.sum(new_cut < mu)
+
+        if positivity_ratio > self.config.maxNoiseImbalance:
+            # This means there is an excess flux, possibly diffuse source,
+            # only estimate around zero.
+            mu, sigma = halfnorm.fit(np.abs(values_noise[values_noise < 0]))
```
fred3m: This bit of code concerns me. You're making an entirely new assumption about where the noise is centered. Even as is, it still needs the `floc=0` added since you are taking an absolute value, which then assumes that `mu` must be the origin of the halfnorm. Have you tested to see what happens when different bands take different paths (in other words r-band uses this path when g- and i-band images use the mu != 0 path?

I wonder if this is necessary because of the above incorrect fit that isn't about 0, or because `_match_sigmas_and_recenter` is doing the wrong thing. Namely, there is a discontinuity at sigma that might explain why sometimes you have extra flux and why the zoomed out tiles don't show enough flux when zoomed out with a Lanczos (like you discussed in the meeting).

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -668,6 +744,22 @@ class PrettyPictureBackgroundFixerConfig(
...
+    extra_pixel_rad = Field[int](
+        doc=(
+            "If there are neighboring input images consider control points that are radius of input image"
+            ", (x^2 + x^2)^(1/2), plus this many pixels away from the center of the input image as control"
+            "points to use."
+        ),
+        default=2000,
+    )
+    max_flux_imbalance = Field[float](
+        doc="When determining background, if the ratio of counts of positive pixels, to negative pixesl"
```
fred3m: pixels

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
+        sub_image = image[image < max(initial_std + 3 * initial_std, max_search_flux)]
+
+        center = mode(np.round(sub_image, 2)).mode
+
         # Don't do anything if there are no pixels to check
         if np.any(mask):
             # use a minimizer to determine best mu and sigma for a Gaussian
             # given only samples below the mean of the Gaussian.
-            mu_hat, sigma_hat = halfnorm.fit(np.abs(image[mask] - maxLikely))
+            mu_hat, sigma_hat = halfnorm.fit(np.abs(image[image < center] - center))
```
fred3m: Needs `floc=0` here

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
+        maxLikely = np.median(image[image < max_search_flux], axis=None)
 
         # find all the pixels that are fainter than this
         # and find the std. This is just used as an initialization
         # parameter and doesn't need to be accurate.
         mask = image < maxLikely
         initial_std = (image[mask] - maxLikely).std()
 
+        # new estimate
+        sub_image = image[image < max(initial_std + 3 * initial_std, max_search_flux)]
```
fred3m: Did you intend this to be `sub_image = image[image < max(maxLikely + 3 * initial_std, max_search_flux)]`? `initial_std + 3 * initial_std` is the same thing as `4 * initial_std`.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
         # Don't do anything if there are no pixels to check
         if np.any(mask):
             # use a minimizer to determine best mu and sigma for a Gaussian
             # given only samples below the mean of the Gaussian.
-            mu_hat, sigma_hat = halfnorm.fit(np.abs(image[mask] - maxLikely))
+            mu_hat, sigma_hat = halfnorm.fit(np.abs(image[image < center] - center))
+            mu_hat += center
             # mu_hat = maxLikely
         else:
             mu_hat, sigma_hat = (maxLikely, 2 * initial_std)
```
fred3m: Should you fix this to be `3 * initial_std` so that it is consistent with the rest of this function?

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
+        center = mode(np.round(sub_image, 2)).mode
+
         # Don't do anything if there are no pixels to check
         if np.any(mask):
             # use a minimizer to determine best mu and sigma for a Gaussian
             # given only samples below the mean of the Gaussian.
-            mu_hat, sigma_hat = halfnorm.fit(np.abs(image[mask] - maxLikely))
+            mu_hat, sigma_hat = halfnorm.fit(np.abs(image[image < center] - center))
+            mu_hat += center
             # mu_hat = maxLikely
```
fred3m: Not in this PR but there commented lines should be removed.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
```diff
@@ -776,93 +868,178 @@ def findBackgroundPixels(image, pos_sigma_mult=1):
...
+            # assume all flux is diffuse
+            return np.inf, -np.inf
+
+        # create a new masking threshold that's the determined
         # mean plus std from the fit
-        threshhold = mu_hat + pos_sigma_mult * sigma_hat
-        image_mask = (image < threshhold) * (image > (mu_hat - 5 * sigma_hat))
-        return image_mask
+        threshhold_pos = min(mu_hat + pos_sigma_mult * sigma_hat, max_search_flux)
+        threshhold_neg = mu_hat - pos_sigma_mult * sigma_hat
```
fred3m: I'm a bit confused as to why you need a negative threshold at all. Isn't everything negative assumed to be oversubtraction? If you do need it, is this unbalanced `threshold_neg` appropriate? What I mean is that `threshold_pos` is clamped at `max_search_flux`, so shouldn't the negative threshold also be clamped? At least please leave a comment that explains why this is done this way.

fred3m (changes requested):
There are a couple of small bugs where you calculate the halfnorm from and some other questions I have about the algorithm that could explain some of the features that we were discussing yesterday at the A&P meeting.

python/lsst/pipe/tasks/rgb2hips/_high_order_hips.py
natelust: So yes technically this is not an exact average of all inputs at once. What it is is incrementally adding warped data to an accumulation array. If there was already data at the location new values are to be placed into, it averages with existing data. For a corner this ends up being (((tract_1 + tract_2)/2 + tract_3)/2 + tract_4) / 2 instead of (tract_1 + tract_2 + tract_3 + tract_4) / 4. As this exists to feather patches together, that should be mostly similar from our processing the difference is not that meaningful. I would need to profile memory usage here, but it is is reasonable I could probably allocate another array of the same dimensions, and make an accumulator N image, and then divide at the end.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: I added logging.warning messages to the two locations. We specifically do not want it to be any kind of failure, as we do want all down steam code to continue to run in these conditions. These conditions basically mean, all of the image is considered to have extended flux I want to leave it unmodified.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: I'm not sure I agree here. What is going on is that we are selecting points below the value that is the first guess at what the central tendency is. We specifically want the low side as it is less likely to be contaminated by extended, low flux. However, the halfnorm method wants to model the positive half of distributions, so this is mirrored with abs. Also, determining the central tendency of tends to be biased a bit high, again because of real objects, and so the real central tendency tends to be just off zero, and so we let the fitter find the offset.  However, in thinking about this response, I realize I have a sign issues and the final mu should be `center - mu_hat` in this case, because of the abs mirroring.

natelust: I added some of this explanation as a code comment

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: In this case I do agree that floc=0 is the right thing since I am pinning at zero intentionally. This whole noise rescaling thing exists because anything below zero in an rgb image gets clipped to black (because what color is negative flux?) the idea for this is not that every pixel must be over black, but assuming a flat background, shifting the noise from being centered about zero, to being centered a little bit above zero, creates an image where statistically more than half of the background pixels are not black. This creates a "grayish" image when averaged over a small area, which is what the eye tends to do. 

When each band takes. a different path you can get slightly different colored noise, but this tends to not happen because this condition kicks in in areas where there is extended flux, which is highly correlated in each band _and_ because each bands sift is related to the minimum across all bands.

This is not related to failures in fits or anything of the sort, but simply there being excess extended flux. If you look at the histogram, it does not look gaussian at all, but a (mostly or completely) increasing function to the right, aka something that looks more like a cdf than a pdf. This test is specifically designed to test this case, if there is an imbalance in area about a determined mean. It the condition is met, it falls back to assuming there is excess flux, and the true background is centered around zero, meaning the noise is described by the tail below zero.

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: nope, good catch, I meant `maxLikely + ...`

python/lsst/pipe/tasks/prettyPictureMaker/_task.py
natelust: The reason we need a lower threshold is that occasionally for some reason we have images with a few random pixels that are REALLY negative. We generate control points for background fixing from bins of the pixels we consider background. When there are a few pixels that are so negative, it brings the estimate way way down, and the control point then drags the background unrealistically high. By excluding them we get a more realistic estimate, and the only consequence is that these pixels that were excluded will still end up negative, and be clipped to black. In principle we could add a lower clipped boundary too, but it would not change the results much at all as the estimation of the sigma at the low side is robust and we do not expect contaminating flux from the low side (by construction of the algorithm)

PR #1316, opened 2026-06-04
DM-55036: add tasks for rewriting visit_image and difference_image via lsst.images

TallJimbo:


TallJimbo:
@leeskelvin , could you please take a look at the last commit here involving the `extended_psf` subpackage?  I've made it essentially required to provide a `plane_map` when converting from a legacy mask, which broke the code here as it was.

python/lsst/pipe/tasks/extended_psf/extended_psf_cutout.py
```diff
@@ -481,10 +481,15 @@ def _get_extended_psf_candidates(
                 focal_plane_angle=candidate["angle_radians"] * u.rad,
             )
 
+            plane_map = get_legacy_visit_image_mask_planes()
+            plane_map["NEIGHBOR"] = MaskPlane(
+                "NEIGHBOR", "Flux in pixel was contaminated by a neighboring source."
```
leeskelvin: Perhaps this description reads better:
```
Flux in pixel is attributed to a neighboring source detection footprint.
```

leeskelvin:
Look good, thanks. I'm happy the ePSF unit tests picked up on this and flagged it for you.

timj (approved):

PR #1320, opened 2026-06-09
DM-55129: Add task for rewriting coadds via lsst.images.

TallJimbo:


python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
```diff
@@ -0,0 +1,234 @@
...
+    )
+    object_background = cT.Input(
+        "deep_coadd_background",
+        storageClass="Background",
+        dimensions={"patch", "band"},
+        doc="The background model used to generate the object catalog, to be subtracted from the coadd.",
+        minimum=0,
+    )
+    object_mask = cT.Input(
+        "{legacy_prefix}deep_coadd.mask",
```
arunkannawadi, line 69: I didn't think we could get just `mask` component alone out like that (I should know)

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
```diff
@@ -0,0 +1,234 @@
...
+        default=(
+            "An alternate background optimized for visually attractive RGB images. "
+            "'Subtracting' this background will generally *add* flux, correcting "
+            "for most oversubtraction problems, but leaving in scattered light and "
+            "some instrumental backgrounds in some cases. "
+            "Because this background is fit to a difference of two wholly different "
+            "coadds that may have different input images, it can also be pulled "
+            "up or down by bright variable or transient objects."
+        ),
+    )
```
arunkannawadi, line 151: Style comment: These appear to have the older `dtype=str` instead of the newer `Field[str]`

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def setDefaults(self) -> None:
```
arunkannawadi, line 153: `super().setDefaults()` missing.

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
TallJimbo: Yup, this definitely works.

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
TallJimbo: I can't say that there was a ton of intent here, but I sort of backed off the `Field[str]` syntax after I realized that it mistypes fields with `optional=True`, and I consider wrong annotations a lot worse than missing annotations (and moreover, the implementation of `Field[str]` syntax was complicated enough I didn't see an easy way to fix it).

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
```diff
@@ -0,0 +1,235 @@
...
+                    alt_bg = field_from_legacy_background(alt_bg_result.background, unit=astropy.units.nJy)
+                    future_cell_coadd.backgrounds.add(
+                        self.config.alternate_background_name,
+                        alt_bg,
+                        self.config.alternate_background_description,
+                    )
+            else:
+                self.log.warning("No alternate background coadd found; rewriting without it.")
+        elif alternate_background_coadd is not None:
+            raise TypeError("alternate_background_coadd provided but config.do_alternate_background=False")
```
timj:
```suggestion
            raise TypeError("alternate_background_coadd provided but config.do_alternate_background_fit=False")
```
?

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
```diff
@@ -0,0 +1,235 @@
...
+        if object_background is not None:
+            # We subtract the deep_coadd_background using afw to avoid tiny
+            # differences between the afw and lsst.images spline
+            # implementations, in the definition of deep_coadd itself (though
+            # that will still ultimately differ from the original due to lossy
+            # compression).
+            future_cell_coadd.image.array -= object_background.getImage().array
+            future_cell_coadd.backgrounds.add(
+                "object",
+                field_from_legacy_background(object_background, unit=astropy.units.nJy),
```
timj, line 229: In RewriteVisitImageTask you pass bounds into calls like this. Should that be done for these calls?

timj (approved):

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
TallJimbo: I can't claim this was super intentional, because it doesn't matter much either way, but I think that difference is broadly sensible, because of the kinds of bounds involved: a visit image's vignetted-polygon really does inform where the background is good, but that's not so true of a coadd background model, where at least interior missing cells are interpolated okay.

PR #1323, opened 2026-06-17
DM-55220: add more components when rewriting difference images

TallJimbo:


python/lsst/pipe/tasks/rewrite_images/_rewrite_difference_image.py
```diff
@@ -59,12 +66,48 @@ class RewriteDifferenceImageConnections(
         deferLoad=True,  # So we can pass preserve_quantization=True as a parameter.
         doc="The input image to convert.",
     )
+    template_psf = cT.Input(
```
isullivan, line 69: A pet peeve, which is really beyond the scope of this ticket: I really don't like the abbreviation of `cT` or `ct` for `connectionTypes`, since it leads to confusing errors if the code is imported in `analysis_tools` that have metrics with units `ct`. I know `cT` shouldn't actually conflict, so I'm not requesting you change the style here.

python/lsst/pipe/tasks/rewrite_images/_rewrite_difference_image.py
```diff
@@ -100,13 +155,42 @@ def runQuantum(
         inputs = butlerQC.get(inputRefs)
         difference_image = inputs.pop("legacy_exposure").get(parameters={"preserve_quantization": True})
         visit_summary = inputs.pop("visit_summary")
+        coadd_data_ids_by_uuid = {h.ref.id: h.ref.dataId for h in inputs.pop("template_coadds")}
```
isullivan, line 158: Minor point, but perhaps the type `template_coadds` could be configurable.

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
```diff
@@ -182,6 +183,15 @@ def run(
             tract_info=tract_info,
             bbox=Box.from_legacy(patch_bbox),
         )
+        if getattr(future_cell_coadd, "_opaque_metadata", None) is None:
+            future_cell_coadd._opaque_metadata = FitsOpaqueMetadata()
+        primary_header = future_cell_coadd._opaque_metadata.headers.setdefault(
+            ExtensionKey(), astropy.io.fits.Header()
```
isullivan, line 191: I don't see `astropy.io.fits` imported

python/lsst/pipe/tasks/rewrite_images/_rewrite_cell_coadd.py
```diff
@@ -182,6 +183,15 @@ def run(
...
+        )
+        primary_header.set("INSTRUME", "LSSTCam")
+        primary_header.set("ORIGIN", "NSF-DOE Vera C. Rubin Observatory")
+        primary_header.set("TELESCOP", "Rubin:Simonyi")
+        primary_header.set("DATE", astropy.time.Time.now().fits, "UTC date this HDU was written.")
```
isullivan, line 196: I don't see `astropy.time` imported.

isullivan (approved):
A couple imports appear to be missing.

python/lsst/pipe/tasks/rewrite_images/_rewrite_difference_image.py
TallJimbo: I suspect anything in `analysis_tools` that uses counts should actually use `electron` (except maybe for HSC/DECam), because that's what we prefer for the pixel units coming out of the LSST ISR task.

python/lsst/pipe/tasks/rewrite_images/_rewrite_difference_image.py
TallJimbo: The dataset type name is, as is always the case for connecftions; this string is the internal name of the connection attribute.

PR #1325, opened 2026-06-22
DM-55282: Add additional info about max number of cosmic ray spans in TooManyCosmicRays exception.

erykoff:


python/lsst/pipe/tasks/repair.py
```diff
@@ -40,13 +40,16 @@ class TooManyCosmicRays(pipeBase.AlgorithmError):
     ----------
     maxCosmicRays : `int`
         Maximum number of cosmic rays allowed.
+    maxCosmicRaySpans : `int`
+        Maximum number of cosmic ray spans allowed.
     """
-    def __init__(self, maxCosmicRays, **kwargs):
+    def __init__(self, maxCosmicRays, maxCosmicRaySpans=-1, **kwargs):
```
laurenam, line 46: Heh, so it's not an API change?

python/lsst/pipe/tasks/repair.py
```diff
@@ -40,13 +40,16 @@ class TooManyCosmicRays(pipeBase.AlgorithmError):
...
         msg = f"Cosmic ray task found more than {maxCosmicRays} cosmics."
```
laurenam, line 47: The message should be updated (to either identify which condition wasn't met or at least indicate it was one of two...)

laurenam (approved):

PR #1334, opened 2026-07-20
DM-55521: update coadd processing tasks to work with lsst.images

TallJimbo:


python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
```diff
@@ -131,6 +139,25 @@ class DeblendCoaddSourcesMultiConfig(PipelineTaskConfig,
         doc="Task to deblend an images in multiple bands"
     )
     idGenerator = SkyMapIdGeneratorConfig.make_field()
+    imageType = ChoiceField(
```
taranu, line 142: Hmm, are you anticipating keeping this around and adding to allowed at some point, even after `legacy` is deprecated? Or did you mainly want better documentation via the allowed dict than you'd get from a boolean Field?

python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
```diff
@@ -194,12 +229,47 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
     def run(self, coadds, bands, mergedDetections, deconvolvedCoadds, idFactory):
+        """Deblend coadds from multiple bands together.
+
+        Parameters
+        ----------
+        coadds : `list` [`lsst.afw.image.Exposure` | \
+                `lsst.images.cells.CellCoadd`]
+            Coadds to deblend.
+        bands : `list` [`str`]
+            Names or the bands for ``coadds`` (zip-iteration compatible).
```
taranu, line 240: Are there lists that _aren't_ zip-iteration compatible, or do you just mean that the bands have to be in the same order as the coadds?

python/lsst/pipe/tasks/fit_coadd_psf.py
```diff
@@ -24,13 +24,15 @@
     "CoaddPsfFitSubConfig", "CoaddPsfFitSubTask", "CoaddPsfFitTask",
 ]
 
+
```
taranu: Accidental, or did you think a new import was going to be needed?

python/lsst/pipe/tasks/fit_coadd_psf.py
```diff
@@ -168,6 +174,23 @@ class CoaddPsfFitConfig(
         doc="Task to fit PSF models for a single coadd",
     )
     idGenerator = SkyMapIdGeneratorConfig.make_field()
+    image_type = pexConfig.ChoiceField(
```
taranu, line 176: Would it make sense to define this once in a module somewhere, or is it safer to duplicate?

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -177,6 +190,45 @@ class DetectCoaddSourcesConfig(PipelineTaskConfig, pipelineConnections=DetectCoa
...
+        "output images are lsst.images.cells.CellCoadd.",
+        dtype=str,
+        default="object",
+    )
+    backgroundDescription = Field(
+        "Description of the subtracted background, to be stored with the image when the input and "
+        "output images are lsst.images.cells.CellCoadd.",
+        dtype=str,
+        default=(
+            "Background subtracted from the image when generating the Object catalog. "
```
taranu: Would it make sense to put this string in a separate module, if not the whole Field?

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -333,7 +422,19 @@ def run(self, exposure, idFactory, expId, patchInfo=None):
             # inability to persist empty BackgroundList.
             emptyBg = self._makeEmptyBackground(exposure, patchInfo)
             backgrounds.append(emptyBg)
-
+        if cell_coadd is not None:
+            cell_coadd.image.array[...] = exposure.image.array
```
taranu, line 432: Huh... is there any functional difference between `[...]` and `[:]` and `.flat`?

python/lsst/pipe/tasks/multiBand.py
```diff
@@ -853,3 +988,13 @@ def run(self, exposure, sources, skyInfo, exposureId, ccdInputs=None,
         results = Struct()
         results.outputSources = sources
         return results
+
+    def _ensureMaskPlanes(self):
```
taranu, line 998: Is this for (current or future) backwards compatibility if running on coadds without newer masks like SPIKE? It's worth a comment regardless.

taranu (approved):
Looks good; see comments and questions.

I assume most of this is going to be tested through CI pipelines but can anything be covered by basic unit tests here, e.g. `_ensureMaskPlanes`?

python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
TallJimbo: I'm not planning to keep it around forever, but on this ticket I really wished that `useCellCoadd` had been a `ChoiceField` instead a boolean, because then I would have been able to use that instead of inventing another config field and leaving `useCellCoadd` sort of dangling in a lot of cases.  So this is probably fighting the last war, but I figured I'd leave room for a plot twist in the migration just in case.

python/lsst/pipe/tasks/deblendCoaddSourcesPipeline.py
TallJimbo: Yes, I mean it should have corresponding elements in the same order.

python/lsst/pipe/tasks/fit_coadd_psf.py
TallJimbo: Accidental; fixed.

python/lsst/pipe/tasks/fit_coadd_psf.py
TallJimbo: The docstrings pretty much always differ, even if only slightly sometimes.

python/lsst/pipe/tasks/multiBand.py
TallJimbo: Yeah, I'm not super comfortable having all of this full-pipeline context here, either.  But I didn't identify any alternatives that I liked a lot better.

Some options:

- Just move it to a module-level constant in this same module, so it's at least importable on its own and a little less buried.
- Make the default empty or much more generic, and override it in drp_pipe.

If you like one of those better, I'm happy to go with it.

python/lsst/pipe/tasks/multiBand.py
TallJimbo: In this limit, if you do `.flat` on both of them, I think it'll amount to the same.  But `[...]` on a 2-d array like this actually resolves to `[:, :]` (i.e. still 2-d), so it would definitely behave differently if broadcasting was in play, and there may be other differences as well.

python/lsst/pipe/tasks/multiBand.py
TallJimbo: Added this docstring:
```
        """Ensure the global mask dictionary has all of the mask planes
        needed for PixelFlags algorithms.

        When mask planes are added, this essentially guarantees that the
        corresponding PixelFlags columns will be wholly False, and usually
        we'd prefer to remove them from the configuration.  But those config
        changes imply a schema changes, and that's not always viable (e.g. on
        a release branch).
        """
```

python/lsst/pipe/tasks/multiBand.py
taranu: I slightly prefer the first option.

TallJimbo:
> I assume most of this is going to be tested through CI pipelines but can anything be covered by basic unit tests here, e.g. _ensureMaskPlanes?

There's just so little test scaffolding for these tasks available that it doesn't seem worthwhile to add all of that just for a little change like this.  Of course that's self-perpetuating, but I can at least confirm that `ci_lsstcam` definitely fails without that `_ensureMaskPlanes` fix.

PR #1335, opened 2026-07-22
DM-55564: Refactor FinalizeCharacterization tasks to reduce i/o

erykoff:


python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -348,20 +343,19 @@ class FinalizeCharacterizationTaskBase(pipeBase.PipelineTask):
...
 
-    def __init__(self, initInputs=None, **kwargs):
-        super().__init__(initInputs=initInputs, **kwargs)
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
 
-        self.schema_mapper, self.schema = self._make_output_schema_mapper(
-            initInputs['src_schema'].schema
-        )
+        # DM-55564: We defer setting the schema and subtasks until
```
isullivan: I wouldn't include the DM ticket number

python/lsst/pipe/tasks/finalizeCharacterization.py
```diff
@@ -750,17 +744,11 @@ def concat_isolated_star_cats(
         merge_cat_counter = 0
         merge_source_counter = 0
 
-        handle = isolated_star_cat_dict[list(isolated_star_cat_dict.keys())[0]]
-        all_source_columns = handle.get(component='columns')
         source_columns = [self.config.id_column, 'obj_index']
-        # visit can be used if it is in the input catalog.
-        if visit is not None and visit in all_source_columns:
```
isullivan, line 757: Not really an issue since you're removing it, but how did this ever work before? Was it supposed to be `'visit' in all_source_columns:`?

isullivan (approved):
Looks good, only a minor comment on a docstring.

python/lsst/pipe/tasks/finalizeCharacterization.py
erykoff: All of that was to work around the way I wrote the tests originally and was too lazy to fix properly.  Which I regret because it wasn't so hard to fix the tests after all, and this turned out to be unnecessarily hammering i/o.

PR #1340, opened 2026-08-24
DM-55829: Warn instead of failing on minor DifferenceImage conversion problems

TallJimbo:


timj (approved):
If adding `log` is the difference between failing and warning that's a bit of a subtle interface 😄

TallJimbo:
Yeah, I regret that, too, but these tasks have now basically fulfilled their only purpose and I'd rather just wait to remove them than make small edits that won't get used.

